PR #7, opened 2017-02-22
Adds authorizations to orders.

cpu:
Refactors the `acme` package to clearly separate out `core` objects that 
are used by Pebble internally vs those that are read to/from protocol messages.

Few misc fixes:
 * fixed returning the parsedCSR for the order endpoint
 * fixed the pebble-client shell to not bail on the "meta" directory
 * updated the in-memory DB's "add" functions to return a count to avoid
   needing funcs like `countRegistration`

Added the authorization types, challenge types, identifier types, and a 
"pending" status type.

New orders now have a pending authorization created for each of the
identifiers present in the CSR's SAN fields. Each authorization has a
http-01 challenge created for it. Authorizations, orders and challenges
are persisted in memory and have GET methods.

TODO: Create TLS-SNI-02 and DNS-01 challenges


acme/common.go
```diff
@@ -19,48 +13,59 @@ const (
 	ResourceNewOrder = Resource("new-order")
 )
 
+type Status string
```
jsha: I think this defining of special types for status, identifier, etc. is overkill in the Boulder codebase, and I'd like to simplify it away here. Can we instead just define string-valued constants, e.g. `StatusPending = "pending"`?

acme/common.go
```diff
@@ -19,48 +13,59 @@ const (
...
 	ToSAgreed bool             `json:"terms-of-service-agreed"`
 	Orders    string           `json:"orders"`
-	Status    string
 }
 
-// OrderRequest is used for new-order requests
-type OrderRequest struct {
-	ID             string   `json:"id"`
-	Status         string   `json:"status"`
+// Order is used for new-order requests
```
jsha: I think this comment needs updating.

cmd/pebble-client/main.go
```diff
@@ -98,7 +98,8 @@ func (c *client) updateDirectory() error {
 		return err
 	}
 
-	var directory map[string]string
+	fmt.Printf("Directory: %s\n", string(respBody))
```
jsha: Debugging printf.

cmd/pebble-client/main.go
```diff
@@ -33,7 +33,7 @@ func userAgent() string {
 
 type client struct {
 	server    *url.URL
-	directory map[string]string
+	directory map[string]interface{}
```
jsha: Is this change necessary? What is it for?

core/common.go
```diff
@@ -0,0 +1,24 @@
+package core
+
+import (
+	"crypto/rand"
+	"encoding/base64"
+	"fmt"
+	"io"
+)
+
+// RandomString and NewToken come from Boulder core/util.go
```
jsha: AFAICT RandomString and NewToken are only used in wfe.go? So they should probably stay there. Creating a common.go is risky new ground. :-) I would also support creating a token.go in the wfe directory, if that's preferable.

wfe/wfe.go
```diff
@@ -390,23 +399,116 @@ func (wfe *WebFrontEndImpl) NewRegistration(
 		return
 	}
 
-	wfe.db.addRegistration(&newReg)
-	wfe.log.Printf("There are now %d registrations in memory\n", wfe.db.countRegistrations())
+	count, err := wfe.db.addRegistration(&createdReg)
+	if err != nil {
+		wfe.sendError(acme.InternalErrorProblem("Error persisting registration"), response)
```
jsha: Let's say "saving" rather than persisting since these aren't persistent.

wfe/wfe.go
```diff
@@ -390,23 +399,116 @@ func (wfe *WebFrontEndImpl) NewRegistration(
...
+				Identifier: ident,
+			},
+		}
+		authz.URL = wfe.relativeEndpoint(request, fmt.Sprintf("%s%s", authzPath, authz.ID))
+		// Create the challenges for this authz
+		err := wfe.makeChallenges(authz, request)
+		if err != nil {
+			return err
+		}
+		// Persist the authorization in memory
```
jsha: Same here.

wfe/wfe.go
```diff
@@ -390,23 +399,116 @@ func (wfe *WebFrontEndImpl) NewRegistration(
...
+		count, err := wfe.db.addAuthorization(authz)
+		if err != nil {
+			return err
+		}
+		fmt.Printf("There are now %d authorizations in the db\n", count)
```
jsha: Debugging printf.

wfe/wfe.go
```diff
@@ -390,23 +399,116 @@ func (wfe *WebFrontEndImpl) NewRegistration(
...
+			Type:  acme.ChallengeHTTP01,
+			Token: core.NewToken(),
+			URL:   authz.URL,
+		},
+	}
+	count, err := wfe.db.addChallenge(chal)
+	if err != nil {
+		return err
+	}
+	fmt.Printf("There are now %d challenges in the db\n", count)
```
jsha: Debugging printf.

wfe/wfe.go
```diff
@@ -427,61 +530,75 @@ func (wfe *WebFrontEndImpl) NewOrder(
...
-	_, err = wfe.db.addOrder(order)
+	// Add the order to the in-memory DB
+	count, err := wfe.db.addOrder(order)
 	if err != nil {
 		wfe.sendError(
 			acme.InternalErrorProblem("Error saving order"), response)
 		return
 	}
+	fmt.Printf("Added order %q to the db\n", order.ID)
+	fmt.Printf("There are now %d orders in the db\n", count)
```
jsha: Debugging printf.

acme/common.go
cpu: That's fair - I'll revert to string constants :+1:

cmd/pebble-client/main.go
cpu: It was necessary after the addition of the `meta` entry. At that point the directory isn't strictly strings. `pebble-client` was panic'ing after the recent ToS changes without it.

wfe/wfe.go
cpu: I left these `printf`'s in on purpose to make it easier to discern the state of the Pebble server. I'll update them to use the logging instead of a raw printf. Fair?

cmd/pebble-client/main.go
jsha: Oh whoops, I didn't notice this was in pebble-client. :+1:

wfe/wfe.go
jsha: Sounds good. We could also do something with expvar if we wanted to get fancy.

core/common.go
cpu: Hah. Yes... I did feel guilty creating yet another `common.go`. I think the idea of using `token.go` in the WFE is the best bet. It felt like wfe.go was getting really large, splitting it up within the package achieves what I wanted. I switched both functions to be unexported as well.

cpu:
@jsha ready for another :mag: - thanks!

PR #8, opened 2017-03-10
Support HTTP-01 challenge validation.

cpu:
This commit adds initial support for HTTP-01 challenge validation. No
other challenge types are implemented at this time.

The implementation is largely a pared down version of what Boulder does
presently. For simplicities sake there are a number of important
differences that make Pebble *absolutely* unappropriate for production
uses:
  1) The validation is performed in the same goroutine as the WFE's HTTP
     handler instead of in a separate go routine. It might be worth
     revamping this in the near future.
  2) There is no read limit on the HTTP request body and timeouts are
     not as strict
  3) The validity of both pending and valid authzs is hardcoded at 1h.

As compared to Boulder the Pebble challenge design chooses to use
a separate endpoint (e.g. `/chalZ/9999` for chal 9999) instead of
referring to challenges subindexed from authorizations (e.g.
`/authZ/9999/1` for the first challenge of authz 9999). I believe this
is still within compliance with the overall specification and will
further exercise clients to not expect Boulder's choices to be
universal. The pebble client is updated to allow specifying a URL
to POST in addition to a directory endpoint to allow POSTing these
separate non-directory defined challenge URLs.

### DIFF
--- a/acme/problems.go
+++ b/acme/problems.go
@@ -11,6 +11,8 @@ const (
 	malformedErr      = errNS + "malformedRequest"
 	badNonceErr       = errNS + "badNonce"
 	agreementReqErr   = errNS + "agreementRequired"
+	connectionErr     = errNS + "connection"
+	unauthorizedErr   = errNS + "unauthorized"
 )
 
 type ProblemDetails struct {
@@ -68,5 +70,21 @@ func AgreementRequiredProblem(detail string) *ProblemDetails {
 		Type:       agreementReqErr,
 		Detail:     detail,
 		HTTPStatus: http.StatusForbidden,
+	}
+}
+
+func ConnectionProblem(detail string) *ProblemDetails {
+	return &ProblemDetails{
+		Type:       connectionErr,
+		Detail:     detail,
+		HTTPStatus: http.StatusBadGateway, // TODO(@cpu): Change this later?
+	}
+}
+
+func UnauthorizedProblem(detail string) *ProblemDetails {
+	return &ProblemDetails{
+		Type:       unauthorizedErr,
+		Detail:     detail,
+		HTTPStatus: http.StatusUnauthorized,
 	}
 }
--- a/acme/problems.go
+++ b/acme/problems.go
@@ -75,16 +75,14 @@ func AgreementRequiredProblem(detail string) *ProblemDetails {
 
 func ConnectionProblem(detail string) *ProblemDetails {
 	return &ProblemDetails{
-		Type:       connectionErr,
-		Detail:     detail,
-		HTTPStatus: http.StatusBadGateway, // TODO(@cpu): Change this later?
+		Type:   connectionErr,
+		Detail: detail,
 	}
 }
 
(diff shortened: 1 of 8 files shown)


acme/problems.go
```diff
@@ -70,3 +72,19 @@ func AgreementRequiredProblem(detail string) *ProblemDetails {
 		HTTPStatus: http.StatusForbidden,
 	}
 }
+
+func ConnectionProblem(detail string) *ProblemDetails {
+	return &ProblemDetails{
+		Type:       connectionErr,
+		Detail:     detail,
+		HTTPStatus: http.StatusBadGateway, // TODO(@cpu): Change this later?
```
jsha: I think it makes the most sense to leave out HTTPStatus for stuff that is returned in the error field of challenges.

va/va.go
```diff
@@ -0,0 +1,162 @@
...
+	httpRequest.Header.Set("User-Agent", userAgent())
+	httpRequest.Header.Set("Accept", "*/*")
+
+	transport := &http.Transport{
+		// We don't expect to make multiple requests to a client, so close
+		// connection immediately.
+		DisableKeepAlives: true,
+	}
+
+	// TODO(@cpu) - should this be made to match Boulder more closely?
```
jsha: I think this is a good example of keeping it simple in Pebble. There's no spec guidance, and I think "too many redirects" is an unlikely failure mode for clients. Let's just leave this out.

va/va.go
```diff
@@ -0,0 +1,162 @@
...
+		if len(via) > maxRedirect {
+			return fmt.Errorf("Maximum redirects (%d) exceeded", maxRedirect)
+		}
+		return nil
+	}
+
+	client := &http.Client{
+		Transport:     transport,
+		CheckRedirect: logRedirect,
+		// TODO(@cpu) - configurable http timeout?
```
jsha: Nah

va/va.go
cpu: Sounds good to me.

cpu:
> The main reason to do this separately would be to break any assumptions like "once I've gotten a request from the ACME server, and responded to it, I can consider the authorization done."
> Probably it makes the most sense to do this in a followup robustness-checking PR that would do something like check, sleep 1 sec, check again, and update the authz, all on a separate goroutine.

Ok, that makes sense. I agree that breaking that assumption is important. When you describe this robustness-checking PR, what is the "check again" after the initial check & sleep? Is that doing the HTTP validation again if the first "check" didn't produce a correct value?

jsha:
>  what is the "check again" after the initial check & sleep

This is to simulate the eventual multi-viewpoint validation we want to do.

cpu:
> This is to simulate the eventual multi-viewpoint validation we want to do.

Aha! Ok, I'll fold in the immediate feedback and then do that as a follow-on PR. Thanks!

jsha:
To be clear, I don't think that's particularly urgent. It should probably come after issuance, for instance.

cpu:
@jsha ready for another pass when you have a chance.

PR #9, opened 2017-03-11
Switch to kid-based signing and verifying.

jsha:
As specified in the latest spec.

In the process, switch to go-jose v2.


jsha:
I thought I did update the pebble-client shell to do this. What am I missing?

cpu:
Oops. You're right. I thought there was a change to be made with the `resource` argument in POSTs but you got that too.

PR #11, opened 2017-03-14
Various fixes noticed during chisel integration.

jsha:
Move Key from ACME type to internal type, since this was made internal-only in
the most recent draft.

Make NotBefore/NotAfter omitempty.

Make Challenges a list of objects rather than a list of URLs, per the spec (yep,
this is somewhat inconsistent with how authorizations are list of URLs).

Change Challenge.URL to Challenge.URI.

Format Order and Authorization Expires properly (RFC3339), and in UTC.

Write out the ACME object for Orders rather than the internal object.

(Note: Builds on top of #10; merge that first).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,4 @@
+language: go
+
+go:
+  - 1.8
--- a/acme/common.go
+++ b/acme/common.go
@@ -1,7 +1,5 @@
 package acme
 
-import "gopkg.in/square/go-jose.v2"
-
 // acme.Resource values identify different types of ACME resources
 type Resource string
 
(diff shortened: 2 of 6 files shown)


cpu:
> Make Challenges a list of objects rather than a list of URLs, per the spec (yep,
this is somewhat inconsistent with how authorizations are list of URLs).

I remember mentally thinking that these were supposed to be in-lined when I wrote the code. I recall having a conversation about it with Richard and wishing it was URLs instead of objects but there was a line of reasoning that made the embedding the preferred approach. Thanks for fixing!

PR #13, opened 2017-03-17
Validation-related fixes.

jsha:
Format all times in UTC and RFC3339.
When POSTing a challenge, return the body and provide an "up" link.
Fix the direction of the expiration check on authzs.

### DIFF
--- a/wfe/wfe.go
+++ b/wfe/wfe.go
@@ -481,7 +481,7 @@ func (wfe *WebFrontEndImpl) makeAuthorizations(order *core.Order, request *http.
 			Type:  acme.IdentifierDNS,
 			Value: name,
 		}
-		now := wfe.clk.Now()
+		now := wfe.clk.Now().UTC()
 		expires := now.Add(pendingAuthzExpire)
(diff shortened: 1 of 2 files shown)


wfe/wfe.go
```diff
@@ -836,6 +836,12 @@ func (wfe *WebFrontEndImpl) updateChallenge(
 				fmt.Sprintf("Failed to validate challenge: %q", err)), response)
 		return
 	}
+	response.Header().Add("Link", link(existingChal.Authz.URL, "up"))
```
cpu: Aha! I forgot about these relations. I think the Authz's need an "up" link to their order as well. Do you want to add that into this PR or should I carry it into my WIP?

jsha: Let's leave it out for now. I'd like to discuss with the ACME list whether all these link relations are really necessary. This one is for now, because the Python `acme` module expects it (thought it could be modified to not).

cpu: Sounds good!

PR #14, opened 2017-03-17
More validation fixes.

jsha:
Add an HTTPPort config and plumb it through.
Calculate the KeyAuthorization based on regID, rather than storing it on
challenge (wasn't getting stored previously).
Show URL in validation records for ease of debugging.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,3 +2,11 @@ language: go
 
 go:
   - 1.8
+
+before_install:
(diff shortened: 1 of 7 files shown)


jsha:
They get serialized when fetching the authz or challenge.

cpu:
> They get serialized when fetching the authz or challenge.
Ahhh! I always forget that the JSON tags on the structure aren't _mandatory_ for the field to be included.

PR #15, opened 2017-03-17
Check out Certbot and run chisel2 against Pebble.

jsha:


.travis.yml
```diff
@@ -1,4 +1,25 @@
...
+  - . venv/bin/activate
+  - cd -
+
+before_script:
+  - pebble &
+  # Wait for pebble to come up
+  - until </dev/tcp/localhost/14000 ; do sleep 0.1 ; done
+
+script:
+  - python ./certbot/tools/chisel2.py  $(openssl rand -hex 3).127.0.0.1.ipna.me $(openssl rand -hex 3).127.0.0.1.ipna.me
```
cpu: Why use `ipna.me` domains here? Any particular reason?

jsha: I didn't want to implement a fake DNS thing, so I needed something that would resolve to 127.0.0.1. This was a handy one. We could alternately install dnsmasq as part of the Travis setup.

cpu: What about [adding a Travis host entry](https://docs.travis-ci.com/user/hosts)?

PR #16, opened 2017-03-21
Certificate issuance

cpu:
This PR adds the initial CA skeleton for doing order issuance.

Pebble generates a root & intermediate keypair/certificate at startup. The intermediate is used for certificate signing purposes and the root issues the intermediate and is otherwise just there to mimick production. In the future we should introduce additional intermediates & chain options to use the full specification. 

Couple other changes:
* Fixed the embedded challenges in authorizations to use a pointer so that the embedded contents are updated when the challenge is completed.
* Replaced a few WFE `Printf`'s with log statements.
* Updated the VA to log whether an HTTP validation was a success or a failure.
* Added a pointer from Authorizations to the Order they belong to
* Added enforcement of Order expiry for authz updates.
* Moved the MemoryStore out of the `wfe` package. This was primarily to let the CA store the root & intermediate certificates in the DB. This allows using the `/certZ/` endpoint to retrieve the root & intermediate.

### DIFF
--- a/core/types.go
+++ b/core/types.go
@@ -86,8 +86,8 @@ func (c Certificate) Chain() []byte {
 	// Add zero or more issuers
 	issuer := c.Issuer
 	for {
-		// if the issuer is nil, or the issuer's issuer is nil then it isn't a leaf
-		// or an intermediate. Don't put it in the chain
+		// if the issuer is nil, or the issuer's issuer is nil then we've reached
+		// the root of the chain and can break
 		if issuer == nil || issuer.Issuer == nil {
 			break
 		}
@@ -96,5 +96,5 @@ func (c Certificate) Chain() []byte {
 	}
 
 	// Return the chain, leaf cert first
-	return bytes.Join(chain, []byte{})
+	return bytes.Join(chain, nil)
 }
--- a/core/types.go
+++ b/core/types.go
@@ -1,9 +1,12 @@
 package core
 
 import (
+	"bytes"
 	"crypto"
 	"crypto/x509"
 	"encoding/base64"
+	"encoding/pem"
+	"fmt"
 	"time"
 
 	"github.com/letsencrypt/pebble/acme"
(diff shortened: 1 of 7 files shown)


core/types.go
```diff
@@ -27,6 +32,7 @@ type Authorization struct {
 	ID          string
 	URL         string
 	ExpiresDate time.Time
+	Order       *Order
```
jsha: I think you meant this to be a list of orders? But AFAICT this isn't used in this PR so should wait.

db/memorystore.go
```diff
@@ -132,11 +135,37 @@ func (m *memoryStore) addChallenge(chal *core.Challenge) (int, error) {
...
+	}
+
+	m.certificatesByID[certID] = cert
+	return len(m.certificatesByID), nil
+}
+
+func (m *MemoryStore) GetCertificateByID(id string) *core.Certificate {
+	m.RLock()
+	defer m.RUnlock()
+	if cert, present := m.certificatesByID[id]; present {
```
cpu: Good catch - this applies equally to the other memory store `GetX` functions. I'll simplify all of them with this round of review feedback.

core/types.go
cpu: It's used by the WFE's `updateChallenge` function where a POST to an Authz checks the Order the authz is associated with to determine if it is expired. If it is then POSTing the Authz isn't allowed.

Should it be a `[]*Order` field? Right now pebble creates authorizations uniquely per order. You can't have an authz shared across orders (at least right now).

core/types.go
```diff
@@ -48,3 +54,47 @@ func (ch Challenge) ExpectedKeyAuthorization(key *jose.JSONWebKey) string {
...
+
+	// Add the leaf certificate
+	chain = append(chain, c.PEM())
+
+	// Add zero or more issuers
+	issuer := c.Issuer
+	for {
+		// if the issuer is nil, or the issuer's issuer is nil then it isn't a leaf
+		// or an intermediate. Don't put it in the chain
+		if issuer == nil || issuer.Issuer == nil {
```
cpu: You're right - thanks!

cpu: Oops! Spoke too soon. I remember why I did this: without checking `issuer.Issuer == nil` then the root issuer is included in the chain as well as the leaf & intermediate. Removing the check results in a chain of three certs when it should be two.

core/types.go
jsha: I was confused by some diff collapsing and thought this was a field on Registration. This is fine for now. Thanks!

PR #17, opened 2017-04-05
Multi-validation & true async issuance

cpu:
**Updates the README**
5491a24 updates the README to include Chisel usage instructions 
& reference the current-most draft.

**Adds multi-validation and true async issuance.**
7d615b5 splits the operation of the CA and VA across multiple
go-routines instead of doing all work on the WFE's handler routine.

The VA now performs multiple validations, sleeping a random amount of
time between each. This forces the ACME client to poll the
authorizations to learn when they have switched from pending to invalid
or valid. All validation attempts must succeed for the authorization to
be valid.

The WFE no longer communicates directly with the CA, instead when the VA
updates the last authorization on an order to valid status it invokes
the CA's CompleteOrder function in a separate goroutine. This will
complete the order & update it with a certificate pointer. The WFE uses
this pointer to construct the certificate URL as required.

### DIFF
--- a/ca/ca.go
+++ b/ca/ca.go
@@ -240,19 +240,17 @@ func (ca *CAImpl) CompleteOrder(order *core.Order) {
 	order.Status = acme.StatusProcessing
 
 	csr := order.ParsedCSR
-	domains := make([]string, len(csr.DNSNames))
-	copy(domains, csr.DNSNames)
-
 	// issue a certificate for the csr
-	cert, err := ca.newCertificate(domains, csr.PublicKey)
+	cert, err := ca.newCertificate(csr.DNSNames, csr.PublicKey)
 	if err != nil {
 		ca.log.Printf("Error: unable to issue order: %s", err.Error())
 		return
 	}
 	ca.log.Printf("Issued certificate serial %s\n", cert.ID)
 
-	// Update the order to valid status with a certificate URL
+	// Update the order to valid status and store a cert ID for the wfe to use to
+	// render the certificate URL for the order
 	order.Status = acme.StatusValid
-	order.Certificate = order.CertPathPrefix + cert.ID
+	order.CertID = cert.ID
 	ca.log.Printf("Order %s has Certificate %s", order.ID, order.Certificate)
 }
--- a/ca/ca.go
+++ b/ca/ca.go
@@ -258,5 +258,5 @@ func (ca *CAImpl) CompleteOrder(order *core.Order) {
 	// Update the order to valid status and store a cert ID for the wfe to use to
 	// render the certificate URL for the order
 	order.Status = acme.StatusValid
-	order.CertID = cert.ID
+	order.CertificateObject = cert
 }
--- a/ca/ca.go
+++ b/ca/ca.go
@@ -221,6 +221,10 @@ func New(log *log.Logger, db *db.MemoryStore) *CAImpl {
 }
 
 func (ca *CAImpl) CompleteOrder(order *core.Order) {
+	// Update the order to reflect that we're now processing it
+	order.Lock()
+	defer order.Unlock()
+
 	// Check the authorizations - this is done by the VA before calling
 	// CompleteOrder but we do it again for robustness sake.
 	for _, authz := range order.AuthorizationObjects {
@@ -232,17 +236,12 @@ func (ca *CAImpl) CompleteOrder(order *core.Order) {
 		authz.RUnlock()
 	}
 
-	order.RLock()
 	if order.Status != acme.StatusPending {
 		ca.log.Printf("Error: Asked to complete order %s is not status pending, was status %s",
 			order.ID, order.Status)
 		return
 	}
-	order.RUnlock()
 
-	// Update the order to reflect that we're now processing it
-	order.Lock()
-	defer order.Unlock()
 	ca.log.Printf("Order %s is fully authorized. Ready to issue", order.ID)
 	order.Status = acme.StatusProcessing
 
--- a/ca/ca.go
+++ b/ca/ca.go
@@ -13,6 +13,7 @@ import (
 	"math/big"
 	"time"
 
+	"github.com/letsencrypt/pebble/acme"
 	"github.com/letsencrypt/pebble/core"
 	"github.com/letsencrypt/pebble/db"
 )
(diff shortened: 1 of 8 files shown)


cpu:
Note: CI is going to fail until Chisel is updated to handle multiple validations & async issuance.

va/va.go
```diff
@@ -29,96 +39,177 @@ func userAgent() string {
...
+	for {
+		// Pull a task off of the tasks channel
+		task, ok := <-va.tasks
+		if !ok {
+			break
+		}
+		va.log.Printf("Pulled a task from the Tasks queue: %#v", task)
+		va.log.Printf("Starting %d validations.", concurrentValidations)
+
+		var wg sync.WaitGroup
```
jsha: This effectively serializes the tasks; If one authorization is validating, another authorization can't be validating at the same time. I think that might potentially mask some client bugs. Also, this loop seems a little long, and I think the goto may not be necessary.

What if instead, this was a very short loop:

```
for task := range va.tasks {
  go process(task)
}
```

And the bulk of this function could be moved into `process(..)`.

va/va.go
```diff
@@ -29,96 +39,177 @@ func userAgent() string {
...
+		// Update the validated date for the challenge regardless of good/bad outcome
+		now := va.clk.Now().UTC()
+		chal.ValidatedDate = now
+		chal.Validated = chal.ValidatedDate.Format(time.RFC3339)
+
+		// Read a validation result from the task results channel for each of the
+		// concurrent validations into a slice for processing.
+		for i := 0; i < concurrentValidations; i++ {
+			va.log.Printf("Reading a result from task.Results")
+			result := <-task.Results
```
jsha: Since we wait on the WaitGroup above, `task.Results` is guaranteed to be exactly filled, and there is guaranteed to be no concurrent access. So there's no reason to make it a channel. It can just be a slice, and then you can `range` over it. Alternately you could get rid of the WaitGroup and use just the channel, which would allow you to surface errors without waiting for all validations to come in. Basically, we need just one of these concurrency primitives here.

Also, I'd recommend factoring this out into a `firstError` function or similar, that returns either an error or nil. That way you can call that function and have an if/else that very clearly handles both the error and non-error cases in one place.

va/va.go
```diff
@@ -29,96 +39,177 @@ func userAgent() string {
...
+	va.log.Printf("Validating HTTP01")
+	body, url, err := va.fetchHTTP(task.Identifier, task.Challenge.Token)
+
+	result := core.ValidationRecord{
+		URL:         url,
+		ValidatedAt: va.clk.Now(),
+	}
 	if err != nil {
-		return err
+		result.Error = err
```
jsha: I think switching this function from one that returns an error, to one that mutates its input to set an error, is a step backwards in clarity. We had similar code in Boulder, but moved towards the "return an error" approach.

Stepping back a bit, this is related to the choice to have a `vaTask` contain both input and output. What if we defined `vaTask` to be only the input to a task, and the output from a task to be a ValidationRecord? Or in other words, treat vaTask as immutable. This might be an argument to use a `chan *ValidationRecord` in place of a `WaitGroup`. The `wg` argument to `performValidation` could be replaced with a `chan`, and the switch statement could do `ch <- va.validateHTTP01(task)`.

cpu:
@jsha I believe all of your feedback is addressed, can you give this PR another pass when you have a chance? Thanks!

ca/ca.go
```diff
@@ -218,3 +219,44 @@ func New(log *log.Logger, db *db.MemoryStore) *CAImpl {
...
+	for _, authz := range order.AuthorizationObjects {
+		// Lock the authorization for reading
+		authz.RLock()
+		if authz.Status != acme.StatusValid {
+			return
+		}
+		authz.RUnlock()
+	}
+
+	order.RLock()
```
jsha: You need to lock earlier, before you fetch order.AuthorixationObjects, Also I think a full lock would be fine; we're not worried about performance here, and there's no deadlock possibility, right?

ca/ca.go
```diff
@@ -218,3 +219,44 @@ func New(log *log.Logger, db *db.MemoryStore) *CAImpl {
...
+	if err != nil {
+		ca.log.Printf("Error: unable to issue order: %s", err.Error())
+		return
+	}
+	ca.log.Printf("Issued certificate serial %s for order %s\n", cert.ID, order.ID)
+
+	// Update the order to valid status and store a cert ID for the wfe to use to
+	// render the certificate URL for the order
+	order.Status = acme.StatusValid
+	order.CertID = cert.ID
```
jsha: Why do you need to store a copy of `cert.ID`? Why not use it directly?

ca/ca.go
```diff
@@ -224,35 +224,39 @@ func (ca *CAImpl) CompleteOrder(order *core.Order) {
...
-	ca.log.Printf("Issued certificate serial %s\n", cert.ID)
+	ca.log.Printf("Issued certificate serial %s for order %s\n", cert.ID, order.ID)
 
-	// Update the order to valid status with a certificate URL
+	// Update the order to valid status and store a cert ID for the wfe to use to
+	// render the certificate URL for the order
 	order.Status = acme.StatusValid
-	order.Certificate = order.CertPathPrefix + cert.ID
-	ca.log.Printf("Order %s has Certificate %s", order.ID, order.Certificate)
+	order.CertID = cert.ID
```
jsha: Since this field is only used to check whether a certificate is done, maybe it would be better to create a field that contains the cert DER, since we'll actually need that later. Otherwise, it's a bit confusing why there are two "ID" fields.

ca/ca.go
cpu: "Why not use it directly" -> You mean a pointer to the certificate?

ca/ca.go
cpu: Good catch on locking too late here. Thanks.

cpu:
@jsha Another round of feedback applied - thanks!

ca/ca.go
jsha: Yeah; I wrote this comment before the earlier one. Is the certificate stored in memory separately from the order? If so, then yes a pointer to the certificate would be good. Otherwise, DER should work.

jsha: Aha, should have read the latest diffs before replying. This looks great.

PR #18, opened 2017-04-13
Add TLS-SNI-02 Challenge Support

cpu:
134a091 adds TLS-SNI-02 challenges as a supported challenge type in
Pebble. The WFE is modified to construct a TLS-SNI-02 challenge along
with an HTTP-01 challenge for every authz. A new configuration parameter
is introduced to control what port the VA uses for TLS, allowing Chisel
to eventually stand up a TLS-SNI-02 server that won't need root to bind
443.

The validation implementation is largely identical to Boulder (thanks
Calavera!) with modifications for Pebble's VA & some slight
simplifications where possible.

### DIFF
--- a/va/va.go
+++ b/va/va.go
@@ -4,14 +4,17 @@ import (
 	"fmt"
 	"io/ioutil"
 	"log"
+	"math/rand"
 	"net/http"
 	"net/url"
 	"runtime"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/jmhodges/clock"
 	"github.com/letsencrypt/pebble/acme"
+	"github.com/letsencrypt/pebble/ca"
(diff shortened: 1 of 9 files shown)


va/va.go
```diff
@@ -38,6 +46,36 @@ func userAgent() string {
 		userAgentBase, runtime.GOOS, runtime.GOARCH)
 }
 
+// certNames collects up all of a certificate's subject names (Subject CN and
+// Subject Alternate Names) and reduces them to a unique, sorted set, typically for an
+// error message
```
jsha: If this is just error messages, I think it can be simpler; it doesn't need to lowercase or unique the names, just join them. And in fact, you can make this method return a string instead for more convenience.

cpu:
@jsha Feedback addressed - thanks!

PR #19, opened 2017-04-19
Adds DNS-01 challenge type support.

cpu:
This commit adds DNS-01 challenges as a supported challenge type for
Pebble. The WFE is modified to construct a DNS-01 challenge along with
the HTTP-01 and TLS-SNI-02 challenges for every new pending authz.
A new configuration parameter is added to the config file for the
upstream DNS resolvers to use.

The validation implementation is very close to Boulder's VA & `bdns`
implementation, but simplified for just TXT records & a simple DNS-01
challenge mechanism. Note: this commit adds a dependency on miekg/dns
and you may need to `go get ./...` since we don't do any sort of godeps
magic/vendoring at present.

### DIFF
--- a/va/va.go
+++ b/va/va.go
@@ -244,7 +244,6 @@ func (va VAImpl) validateDNS01(task *vaTask) *core.ValidationRecord {
 	authorizedKeysDigest := base64.RawURLEncoding.EncodeToString(h.Sum(nil))
 
 	for _, element := range txts {
-		va.log.Printf("TXT: %q", element)
 		if subtle.ConstantTimeCompare([]byte(element), []byte(authorizedKeysDigest)) == 1 {
 			return result
 		}
--- a/va/va.go
+++ b/va/va.go
@@ -1,10 +1,12 @@
 package va
 
 import (
+	"context"
 	"crypto/sha256"
 	"crypto/subtle"
 	"crypto/tls"
 	"crypto/x509"
+	"encoding/base64"
 	"encoding/hex"
 	"fmt"
 	"io/ioutil"
@@ -22,6 +24,7 @@ import (
 	"github.com/letsencrypt/pebble/acme"
 	"github.com/letsencrypt/pebble/ca"
 	"github.com/letsencrypt/pebble/core"
+	"github.com/miekg/dns"
 )
 
 const (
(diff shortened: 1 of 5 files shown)


va/va.go
```diff
@@ -199,17 +215,117 @@ func (va VAImpl) performValidation(task *vaTask, results chan<- *core.Validation
...
+func (va VAImpl) lookupTXT(hostname string) ([]string, error) {
+	const qtype = dns.TypeTXT
+	var txt []string
+
+	dnsTimeout := time.Second * 5
+	ctx, _ := context.WithDeadline(context.Background(), va.clk.Now().Add(dnsTimeout))
+
+	m := new(dns.Msg)
+	m.SetQuestion(dns.Fqdn(hostname), qtype)
+	// Set DNSSEC OK bit for resolver
```
jsha: This is not accurate. The docs for this:

> func (dns *Msg) SetEdns0(udpsize uint16, do bool) *Msg
>    SetEdns0 appends a EDNS0 OPT RR to the message. TSIG should always the
>    last RR in a message.

So this should be something like "accept large UDP replies." Also note that (a) we don't expect large replies because we don't evaluate DNS ourselves, and (b) if you're talking to 8.8.8.8, you're going over the public internet, which I think has a lower MTU than in-datacenter. So you'd expect IP fragmentation with a 4096-bit reply. I think let's delete this for now, and we can experiment with different settings if it causes real-world problems.

va/va.go
```diff
@@ -22,6 +24,7 @@ import (
 	"github.com/letsencrypt/pebble/acme"
 	"github.com/letsencrypt/pebble/ca"
 	"github.com/letsencrypt/pebble/core"
+	"github.com/miekg/dns"
```
jsha: Is there a reason to use miekg/dns here instead of https://golang.org/pkg/net/#LookupTXT ?

cpu: I didn't know it was available :-) I was following the Boulder VA which used this dep. I'll try to rework on top of `net.LookupTXT`

va/va.go
cpu: Interesting! that will teach me to lift a comment without verifying. That came [unedited from Boulder](https://github.com/letsencrypt/boulder/blob/409f1623e6369ed1902fd704d577df1657878abd/bdns/dns.go#L228:L229).

I will remove it. Thanks!

jsha: Yeah, I remember being confused about this bit and this comment several times. I think probably part of the confusion is that the main reason to set this higher is because DNSSEC replies are large.

cpu:
@jsha feedback addressed - thanks!

PR #25, opened 2017-05-03
Rename challenge field from uri to url

shred:
I'm sorry, but I found another one... 😉 In the Challenge resource, the `uri` field has been renamed to `url`.


jsha:
Nice work, thanks! Note: build is failing because the in-progress acme-v2 branch of certbot doesn't yet support the `url` version. I'll go ahead and merge for pebble, then update the acme-v2 branch.

PR #27, opened 2017-05-17
Switch to lookup-based validation.

jsha:
The latest version of the ACME spec requires all requests other than
registration to be authenticated by looking up an account key via the
"kid" JWS header parameter. This change implements that by defining two methods
to get keys from a JWS: one that extracts the key from the "jwk" header field
(for use during registration) and one that extracts it from the "kid" header
field (for use with all other requests). It also enforces that the two header
fields are mutually exclusive.

### DIFF
--- a/wfe/wfe.go
+++ b/wfe/wfe.go
@@ -294,7 +294,8 @@ func (wfe *WebFrontEndImpl) parseJWS(body string) (*jose.JSONWebSignature, error
 	return parsedJWS, nil
 }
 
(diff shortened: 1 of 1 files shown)


wfe/wfe.go
```diff
@@ -388,7 +403,7 @@ func (wfe *WebFrontEndImpl) NewRegistration(
 	response http.ResponseWriter,
 	request *http.Request) {
 
-	body, key, prob := wfe.verifyPOST(ctx, logEvent, request)
+	body, key, prob := wfe.verifyPOST(ctx, logEvent, request, wfe.extractJWK)
```
cpu: Can you add a brief comment here describing why this is the only place to use `extractJWK` vs `lookupJWK`?

PR #30, opened 2017-07-05
Verify contact scheme & email addresses

cpu:
This commit adds a `verifyContacts` method to the WFE that checks a new account's contact array to validate the contacts use the `mailto:` protocol and are generally well-formed. This adds support for testing
the "invalidContact" and "unsupportedContact" error types from the latest ACME drafts.

PR #31, opened 2017-07-05
Add support for "accountDoesNotExist" error type.

cpu:
Recent drafts define an explicit "accountDoesNotExist" error type for cases where the account indicated by a request can't be located. This commit updates Pebble to return this error type where appropriate.

### DIFF
--- a/wfe/wfe.go
+++ b/wfe/wfe.go
@@ -13,10 +13,12 @@ import (
 	"io/ioutil"
 	"log"
 	"net/http"
+	"net/mail"
 	"net/url"
 	"path"
 	"strings"
 	"time"
+	"unicode"
 
 	"gopkg.in/square/go-jose.v2"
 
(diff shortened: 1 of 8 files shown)


cpu:
@jsha This should be clear of conflicts again as well. I merged #30 into this so if they are merged in order of PR number I _think_ it should be conflict free.

wfe/wfe.go
```diff
@@ -873,7 +874,8 @@ func (wfe *WebFrontEndImpl) getAcctByKey(key crypto.PublicKey) (*core.Account, *
 	// Find the existing account object for that key ID
 	var existingAcct *core.Account
 	if existingAcct = wfe.db.GetAccountByID(regID); existingAcct == nil {
-		return nil, acme.MalformedProblem("No existing account for signer's public key")
+		return nil, acme.AccountDoesNotExistProblem(
+			"No existing account for signer's public key")
```
jsha: Instead let's say "URL in JWS 'kid' field does not correspond to an account."

PR #32, opened 2017-07-05
Add support for "only-return-existing" on new-account.

cpu:
The later ACME draft specifies that one can retrieve account information for a given public key by signing a JWS request to the new-account endpoint that contains "only-return-existing: true". In such a case if the server can't find an existing account it should *not* create one but instead return an error. This commit adds this behaviour to Pebble.


cpu:
Ditto on this PR being the child of 31 and 30 and merging in order of PR #.

PR #34, opened 2017-07-08
Refactored config, tuneable VA sleeps.

cpu:
#### Make pebble config accessible to components.

This PR refactors the Pebble config from being accessible only to the toplevel cmds and moves it such that it can be passed directly to the components (e.g. the VA) to reduce the number of parameters being added to constructors.

#### Allow configuring validation sleep in config.
This PR updates the VA to use a ConfigDuration read from the config file to determine the maximum random sleep between validation attempts. Setting this value to "0s" in the config will result in now sleeps between validation and can be useful for testing without incurring extra delays.

### DIFF
--- a/core/config.go
+++ b/core/config.go
@@ -2,15 +2,50 @@ package core
 
 import (
 	"encoding/json"
+	"errors"
 	"io/ioutil"
+	"time"
 )
 
 type Config struct {
 	Pebble struct {
-		ListenAddress string
-		HTTPPort      int
-		TLSPort       int
+		ListenAddress   string
+		HTTPPort        int
+		TLSPort         int
+		ValidationSleep ConfigDuration
 	}
+}
+
+// ConfigDuration is just an alias for time.Duration that allows
+// serialization to JSON. It is borrowed from boulder/cmd/config.go
+type ConfigDuration struct {
+	time.Duration
+}
+
+// ErrDurationMustBeString is returned when a non-string value is
+// presented to be deserialized as a ConfigDuration
+var ErrDurationMustBeString = errors.New("cannot JSON unmarshal something other than a string into a ConfigDuration")
+
+// UnmarshalJSON parses a string into a ConfigDuration using
+// time.ParseDuration.  If the input does not unmarshal as a
+// string, then UnmarshalJSON returns ErrDurationMustBeString.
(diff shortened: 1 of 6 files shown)


jsha:
> This PR refactors the Pebble config from being accessible only to the toplevel cmds and moves it such that it can be passed directly to the components (e.g. the VA)

We originally were going somewhat in this direction in Boulder, but Jeff convinced me it was a bad idea, because it created tight coupling between the main code and the components, and made it hard to test things. We came up with the [Config plan](https://github.com/letsencrypt/boulder/wiki/Config-plan), which is mostly true today, with some exceptions (in particular we still have a bunch of exported fields on components that don't need to be). Pebble is different, and we don't *have* to follow that plan, but I think it would be confusing to have such a significantly different architectural pattern here than in Boulder. What do you think?

core/config.go
```diff
@@ -0,0 +1,63 @@
...
+	"io/ioutil"
+	"time"
+)
+
+type Config struct {
+	Pebble struct {
+		ListenAddress   string
+		HTTPPort        int
+		TLSPort         int
+		ValidationSleep ConfigDuration
```
jsha: I was actually thinking more along the lines of an environment variable, so it would be easy to override without creating diffs to the config that we'd have to remember not to commit. Sorry I forgot to mention this yesterday; I think we had talked about the possibility, but it was a long time ago.

cpu:
> We originally were going somewhat in this direction in Boulder, but Jeff convinced me it was a bad idea, because it created tight coupling between the main code and the components, and made it hard to test things. We came up with the Config plan, which is mostly true today, with some exceptions (in particular we still have a bunch of exported fields on components that don't need to be). Pebble is different, and we don't have to follow that plan, but I think it would be confusing to have such a significantly different architectural pattern here than in Boulder. What do you think?

Ahhh. That makes sense. I wondered why I didn't wire the config through initially and it was probably modelled after Boulder.

I'm fine with continuing in line with how we manage configuration in Boulder. I don't think the rationale is as applicable to Pebble (there are no unit tests to be hard to write, and I don't think we care for much abstraction between components), but I also don't have a strong preference to change it. I'll close this PR and submit something simpler with an environmental variable. Thanks!

PR #37, opened 2017-08-04
Reject JWS with unprotected headers in `parseJWS`.

cpu:
This PR explicitly rejects any HTTP POSTS that arrive with
unprotected headers to `parseJWS`. Since go-jose strips these
unprotected headers in the result from `jose.ParseSigned` we
do a quick JSON unmarshal to test for the unprotected headers
ourselves prior to calling `ParseSigned`.

An instance where the wrong `err` was being checked in 
`verifyPOST` is also fixed.

Resolves https://github.com/letsencrypt/boulder/issues/2940

### DIFF
--- a/wfe/wfe.go
+++ b/wfe/wfe.go
@@ -284,6 +284,22 @@ func keyToID(key crypto.PublicKey) (string, error) {
 }
 
 func (wfe *WebFrontEndImpl) parseJWS(body string) (*jose.JSONWebSignature, error) {
+	// Parse the raw JWS JSON to check that the unprotected Header field is not
+	// being used for a key ID or a JWK. This must be done prior to
+	// `jose.parseSigned` since it will strip away these headers.
+	var unprotected struct {
+		Header map[string]string
+	}
+	if err := json.Unmarshal([]byte(body), &unprotected); err != nil {
+		return nil, errors.New("Parse error reading JWS")
+	}
+
+	// ACME v2 never uses values from the unprotected JWS header. Reject JWS that
+	// include unprotected headers.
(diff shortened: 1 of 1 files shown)


shred:
With this commit, Pebble does not accept JWS Compact Serialization (RFC 7515 section 3.1) input any more. It fails on the attempt to parse the binary input as JSON ("Parse error reading JWS").

I don't know if this is actually a regression, as the ACME specification does not explicitly mention that the use of the compact format is also acceptable. However, Boulder and also Pebble have accepted the format up to now.

The compact format neither supports unprotected headers nor multiple signatures, so it should be safe to accept it as an alternative format.

jsha:
@shred does acme4j use JWS Compact Serialization? Does it rely on a library that only supports JWS Compact Serialization? Ideally I'd like to consolidate on a single serialization for our v2 implementations if the support is out there.

shred:
@jsha Yes, it broke acme4j, but I can easily fix that... I was just wondering if you were aware of the loss of compact serialization support.

If you like to consolidate to a single serialization type, the compact format should be explicitly excluded in the specs IMHO. It might lead to incompatible clients if some servers support it, and some don't.

cpu:
@shred Apologies about the sudden breakage!

I was under the impression the spec already forbade the compact serialization format but I must be mistaken. I agree that we should try to make it more obvious if we intend to only support the one.

shred:
No need to apologize. It's an easy fix, and it also showed that my integration test efforts payed off. :smile:

I have checked the specs again, but the word "compact" is not used at all. So the compact format is not forbidden expressis verbis.

PR #43, opened 2017-08-27
Do not send `orders` field of account when empty.

cpu:
Prior to this commit the `orders` field of an account was always
serialized to JSON even when it was empty (e.g. `""`). This commit
updates the JSON tag to include `omitempty`. Now the field does not
appear in response to new-account creation requests.

This will help the [acme4j](https://github.com/shred/acme4j/b) project [remove a workaround](https://github.com/shred/acme4j/blob/b708b2f199fef8a561794ea321f3b1c9b926e552/acme4j-client/src/main/java/org/shredzone/acme4j/Account.java#L324:L327) required 
only for Pebble.

### DIFF
--- a/acme/common.go
+++ b/acme/common.go
@@ -35,7 +35,7 @@ type Account struct {
 	Status             string   `json:"status"`
 	Contact            []string `json:"contact"`
 	ToSAgreed          bool     `json:"terms-of-service-agreed"`
-	Orders             string   `json:"orders"`
+	Orders             string   `json:"orders,omitempty"`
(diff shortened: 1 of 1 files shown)


cpu:
cc @shred - I think this will let you remove the `public static boolean workaround()` function and its references from the Pebble acme4j provider class. If you find anything else that makes you require carrying a workaround please let me know so we can fix it :-)

shred:
The account part seems to be work in progress, so I didn't want to bother you with this problem. The change looks good to me, but I will test it as soon as it is merged. Thank you!

PR #47, opened 2017-11-06
Remove proactive issuance and add order finalization.

cpu:
This commit implements https://github.com/ietf-wg-acme/acme/pull/342 - replacing proactive issuance and CSR as part of new-order with an explicit order finalization step that delivers the CSR.

This is largely a port of the work done to add order finalization to the WIP ACMEv2 support in Boulder:
https://github.com/letsencrypt/boulder/pull/3169 

I haven't tested this end-to-end yet - There are likely bugs lurking :-)

### DIFF
--- a/wfe/wfe.go
+++ b/wfe/wfe.go
@@ -775,7 +775,7 @@ func (wfe *WebFrontEndImpl) NewOrder(
 		orderNames = append(orderNames, ident.Value)
 	}
 
-	// Store the unique lower version of the names on the order ob
+	// Store the unique lower version of the names on the order object
 	order.Names = uniqueLowerNames(orderNames)
 
 	// Create the authorizations for the order
--- a/wfe/wfe.go
+++ b/wfe/wfe.go
@@ -970,10 +970,9 @@ func (wfe *WebFrontEndImpl) FinalizeOrder(
 		}
 	}
 
-	// Lock and update the order to be in processing status with a parsed CSR available.
+	// Lock and update the order with the parsed CSR.
 	existingOrder.Lock()
 	existingOrder.ParsedCSR = parsedCSR
-	existingOrder.Status = acme.StatusProcessing
 	existingOrder.Unlock()
 
 	// Check whether the order is ready to issue, if it isn't, return a problem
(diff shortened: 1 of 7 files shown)


cpu:
@shred Reports that ACME4J was able to provision a new-order but upon finalizing it the order remained in pending state. This is likely a bug in this PR - I'm going to try and investigate further in the next few days when I complete the Boulder work I have assigned.

jsha:
@cpu - you mention above that the order remained pending after completion; have you had a chance to look at that?

wfe/wfe.go
```diff
@@ -595,19 +602,17 @@ func (wfe *WebFrontEndImpl) verifyOrder(order *core.Order, reg *core.Account) *a
 	if reg == nil {
 		return acme.InternalErrorProblem("Account is nil")
 	}
-	csr := order.ParsedCSR
-	if csr == nil {
-		return acme.InternalErrorProblem("Parsed CSR is nil")
+	idents := order.Identifiers
+	if idents == nil || len(idents) == 0 {
```
jsha: len(nil) == 0, so you can simplify this to `if len(idents) == 0`

wfe/wfe.go
```diff
@@ -777,6 +769,15 @@ func (wfe *WebFrontEndImpl) NewOrder(
 		return
 	}
 
+	// Collect all of the DNS identifier values up into a []string
+	var orderNames []string
+	for _, ident := range order.Identifiers {
+		orderNames = append(orderNames, ident.Value)
+	}
+
+	// Store the unique lower version of the names on the order ob
```
jsha: *order object

cpu:
> @cpu - you mention above that the order remained pending after completion; have you had a chance to look at that?

Apologies, I lost this thread over the past couple of weeks. I applied your review feedback and then tested an order finalization against 60d0ba8 successfully. After finalizing a fully authorized order a certificate was issued asynchronously and a subsequent poll of the order returned it in valid status with a certificate URL.

A side-effect of moving the state change from pending -> processing from the wfe to the CA component to help simplify the locking semantics is that when the finalize API request returns the order to the user its still in pending status. The CA processes the order & status change in a separate go routine. I don't think the spec says that the order is to be reflected back in a new-order response, just a 200 OK - we might be able to just drop that response entirely instead of sending it back with a pending status when it should ~probably be processing. Thoughts?

cpu:
> I don't think the spec says that the order is to be reflected back in a new-order response, just a 200 OK - we might be able to just drop that response entirely instead of sending it back with a pending status when it should ~probably be processing.

I'm wrong. When I specified this I wrote:

"A valid request to finalize an order will return the order to be finalized. "

I'm not sure how inappropriate it is for the order to remain "pending". In some sense it is, the CA isn't processing it yet.

jsha:
I think it's fine to return the order in a "pending" state.

PR #55, opened 2017-11-10
Change go get path to prevent error

robaman:
Change go get path to prevent error in docs
package github.com/letsencrypt/pebble: no Go files in $GOPATH/src/github.com/letsencrypt/pebble

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ clients are not hardcoding URLs.)
 ## Install
 
 1. [Set up Go](https://golang.org/doc/install) and your `$GOPATH`
-2. `go get -u github.com/letsencrypt/pebble`
+2. `go get -u github.com/letsencrypt/pebble/cmd/pebble`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -47,7 +47,7 @@ clients are not hardcoding URLs.)
 ## Install
 
 1. [Set up Go](https://golang.org/doc/install) and your `$GOPATH`
-2. `go get -u github.com/letsencrypt/pebble`
+2. `go get -u github.com/letsencrypt/pebble/cmd/pebble`
```
cpu: Thanks for catching this. Does `go get -u github.com/letsencrypt/pebble/...` work for you on a clean system? I think it would be a better replacement to bring in the pebble-shell and any other cmds we might add in-tree.

robaman: Definitely. Yep, it works on a clean system. Updating PR.

PR #59, opened 2017-11-22
Fix tests and include a copy of chisel2

jsha:
I've pushed some changes to the Certbot acme-v2 branch to make it work again. These included a change to the `acme` package API, specifically that `poll_order` became `poll_order_and_request_issuance`, so chisel2 needed to be updated to the new API.

While we are ironing out the Certbot acme v2 API, I think it makes sense for Pebble to have its own copy of chisel2, since it can be revised more easily than the copy in Boulder. I've added a copy of chisel2 that is up to date with the latest acme-v2 branch on Certbot.

For some reason this branch is showing a bunch of old commits, but if we squash merge I think it should be fine.

### DIFF
--- a/test/chisel2.py
+++ b/test/chisel2.py
@@ -1,6 +1,6 @@
 """
 A simple client that uses the Python ACME library to run a test issuance against
-a local Boulder server. Unlike chisel.py this version implements the most recent
+a local Chisel server. Unlike chisel.py this version implements the most recent
 version of the ACME specification. Usage:
 
 $ virtualenv venv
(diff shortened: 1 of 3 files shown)


test/chisel2.py
```diff
@@ -0,0 +1,216 @@
+"""
+A simple client that uses the Python ACME library to run a test issuance against
+a local Boulder server. Unlike chisel.py this version implements the most recent
```
cpu: s/Boulder/Pebble/g?

test/chisel2.py
```diff
@@ -0,0 +1,216 @@
...
+from acme import errors as acme_errors
+from acme import jose
+from acme import messages
+from acme import standalone
+
+logging.basicConfig()
+logger = logging.getLogger()
+logger.setLevel(int(os.getenv('LOGLEVEL', 0)))
+
+DIRECTORY = os.getenv('DIRECTORY', 'http://localhost:4001/directory')
```
cpu: If we're storing this locally we might as well default to the same port/directory as Pebble uses.

test/chisel2.py
```diff
@@ -1,6 +1,6 @@
 """
 A simple client that uses the Python ACME library to run a test issuance against
-a local Boulder server. Unlike chisel.py this version implements the most recent
+a local Chisel server. Unlike chisel.py this version implements the most recent
```
cpu: s/Chisel/Pebble/g :laughing:

Issue #63, opened 2017-12-05
Listen on HTTPS with minica issued leaf cert


cpu:
Pebble should include a [miniCA](https://github.com/jsha/minica) generated CA certificate in the repo. We should also include a leaf certificate issued from the CA for `127.0.0.1`. 

Users requiring trusted TLS for their **test code** (don't do this for anything else!) can import the CA certificate into their client trust store. Users with more complicated needs (e.g. a different subject SAN in the leaf certificate) should replace the CA & leaf cert in the repo with their own.

PR #65, opened 2017-12-06
Use HTTPS w/ miniCA PKI for Pebble ACME API.

cpu:
This commit adds a root certificate & an issued end-entity certificate
(plus accompanying private keys) for the Pebble API. By default the EE
cert has SANs for `127.0.0.1`, `pebble`, and `localhost`. READMEs are
added to explain the cert generation & **DANGER** around adding the CA
to a prod trust store. Pebble's default config is updated to use the EE
cert & the `pebble` command now invokes `http.ListenAndServeTLS`.

This will make Pebble a better test server since the ACME specification [says](https://tools.ietf.org/html/draft-ietf-acme-acme-08#section-6.1):
> Use of HTTPS is REQUIRED.

Resolves https://github.com/letsencrypt/pebble/issues/63


cpu:
> Generally looks good; in the documentation about the test root, instead of suggesting to add to a trust store, I would say that your test client should offer a configuration option with a list of roots to trust.

Sounds good. I'll make some edits.

PR #70, opened 2017-12-08
Do not expect/validate "Type" in challenge POST.

### LINKED ISSUE #69: Challenge update expect "type"
Pebble says

> Challenge update was type , existing challenge is type dns-01

The corresponding code is

https://github.com/letsencrypt/pebble/blob/878eac3d373923ff76f9f80423f57a91f1df32c8/wfe/wfe.go#L1107-L1114

But the ACME example just reads

```json
  "payload": base64url({
    "keyAuthorization": "IlirfxKKXA...vb29HhjjLPSggwiE"
  }),
```

– https://ietf-wg-acme.github.io/acme/draft-ietf-acme-acme.html#responding-to-challenges

I think pebble shouldn't expect or check a "type" field here.

cpu:
Previous to this commit the Pebble `validateChallengeUpdate` function
assumed that the POST JWS payload sent to the challenge update endpoint
contained both "type" and "keyAuthorization" fields. If the provided
"type" value didn't match the server's understanding of the challenge
type an error was returned.

The current ACME draft specification does not require this "type" field.
Section 7.5.1 only describes sending a "keyAuthorization". This
commit removes the expectation that a type field is sent, and the
constraint that it be equal to the expected challenge type.

This resolves #69 - Thanks to @sophie-h for flagging the divergence 
& reporting the bug!

### DIFF
--- a/wfe/wfe.go
+++ b/wfe/wfe.go
@@ -1104,15 +1104,6 @@ func (wfe *WebFrontEndImpl) validateChallengeUpdate(
 	chal.RLock()
 	defer chal.RUnlock()
 
-	// Check that the challenge update is the same type as the challenge
-	// NOTE: Boulder doesn't do this at the time of writing and instead increments
-	//       a "StartChallengeWrongType" stat
-	if update.Type != chal.Type {
-		return nil, acme.MalformedProblem(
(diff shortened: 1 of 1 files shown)


cpu:
@jsha Gentle bump on :mag:ing this PR. Thanks!

PR #86, opened 2018-01-08
Clarify missing functionality

cpu:


cpu:
Self-merging since this is only a small README change.

Issue #84, opened 2018-01-06
Required 'orders' key missing in account objects


sophie-h:
> orders (required, string):
> A URL from which a list of orders submitted by this account can be fetched via a GET request, as described in Section 7.1.2.1.

PR #87, opened 2018-01-08
Return "orders" field of account even when empty.

cpu:
This commit updates the `acme/common.go` type for `Account` to remove
the `omitempty` json tag. This ensures the field is returned even when
empty (e.g. because the account was just created).

Also clarify that there isn't a way to get or update account details
presently. (E.g. the fact that `orders` is never populated is OK for now
because there's not any way to get an Account object in a response
except from a successful newAccount req that has an empty orders field).

cc @sophie-h

Resolves https://github.com/letsencrypt/pebble/issues/84

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,9 @@ Pebble is **NOT INTENDED FOR PRODUCTION USE**. Pebble is for **testing only**.
 By design Pebble will drop all of its state between invocations and will
 randomize keys/certificates used for issuance.
 
+Pebble is not yet a **complete** ACME implementation. It does not presently
+support revocation, account key rollover or account update/retreival.
+
 ## Goals
 
 1. Produce a simplified testing front end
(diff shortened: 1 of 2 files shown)


sophie-h:
Hey @cpu,

I don't think this resolves #84 because this returns an empty string (correct?) for 'orders' which is not a URL with `https:` schema.

cpu:
Hi @sophie-h,

> I don't think this resolves #84 because this returns an empty string (correct?) for 'orders' which is not a URL with https: schema.

Oops, I overlooked that the field is `string` not `[]string`. My intention was for it to return `orders: []`, an empty array of strings. I think that's correct since there are no orders yet and so no order URLs.

I'll fix the field type so that the above happens correctly. Thanks for catching that!

sophie-h:
Noo :laughing:

```
   orders (required, string):  A URL from which a list of orders
      submitted by this account can be fetched via a GET request, as
      described in Section 7.1.2.1.
```

cpu:
@sophie-h Ugh! I swore there was a point when the spec returned `orders (required, []string)` of URLs.

I'm going to close this PR then since I don't have cycles to implement a new endpoint for retrieving orders and the spec doesn't allow for an empty response of any use. Clients should keep track of their own orders for now. Thanks for clearing the cobwebs out of my head on this one.

Issue #90, opened 2018-02-07
newAccount returning status as empty string


rmbolger:
Currently testing against commit eb01c4f using Windows PowerShell 5.1's `Invoke-WebRequest`.

When I submit a newAccount request to the `/sign-me-up` endpoint, everything goes as expected except the value of `status` in the returned account object. The HTTP response code is 201. The headers include a new nonce and Location Url of the new account. And the body is returned as:

```javascript
{
   "status": "",
   "contact": [
      "mailto:me@example.com"
   ],
   "termsOfServiceAgreed": true,
   "onlyReturnExisting": false
}
```

[Section 7.3 of the 09 spec](https://tools.ietf.org/html/draft-ietf-acme-acme-09#section-7.3) would have me believe `status` should be set to `"valid"`. Or am I misinterpreting?

Here are the associated Pebble logs:

```
Pebble 2018/02/07 14:48:07 Generated new root issuer with serial 7b3527e7d7cfcc40
Pebble 2018/02/07 14:48:07 Generated new intermediate issuer with serial 160a8981343113e1
Pebble 2018/02/07 14:48:07 Configured to reject 15% of good nonces
Pebble 2018/02/07 14:48:07 Pebble running, listening on: 0.0.0.0:14000
Pebble 2018/02/07 14:48:23 HEAD /nonce-plz -> calling handler()
Pebble 2018/02/07 14:52:23 POST /sign-me-up -> calling handler()
Pebble 2018/02/07 14:52:23 There are now 1 accounts in memory
```

cpu:
Hi @rmbolger 

> Section 7.3 of the 09 spec would have me believe status should be set to "valid". Or am I misinterpreting?

You're not misinterpreting. Sounds like a Pebble bug. Thanks for reporting!

cpu:
Hi again @rmbolger :wave:

This should be fixed for you in master with 48220a83e693ea77ccadbba76a44ca74ef4fa6f4. Thanks again

rmbolger:
Thanks everyone.

PR #91, opened 2018-02-09
Rework NewAccount to ensure a status is returned.

cpu:
Prior to this commit `wfe.NewAccount` returned an account object with an
empty status field when it should be "valid". This commit ensures that
the initial account object is created with the `Status` field set to
"valid".

This commit also refactors the `NewAccount` code such that the
`acme.Account` type can be used for already created accounts and doesn't
need to maintain a `ToSAgreed` or `OnlyReturnExisting` field. Instead
the WFE uses a throw away struct for the new-account request that holds
the `ToSAgreed` and `OnlyReturnExisting` fields. This helps prevent the
client from POSTing their own `Status` or `Orders` field and makes the
internal types cleaner.

Thanks to @rmbolger for flagging this spec divergence. Resolves https://github.com/letsencrypt/pebble/issues/90

### DIFF
--- a/acme/common.go
+++ b/acme/common.go
@@ -24,11 +24,9 @@ type Identifier struct {
 }
 
 type Account struct {
-	Status             string   `json:"status"`
-	Contact            []string `json:"contact"`
-	ToSAgreed          bool     `json:"termsOfServiceAgreed"`
-	Orders             string   `json:"orders,omitempty"`
-	OnlyReturnExisting bool     `json:"onlyReturnExisting"`
+	Status  string   `json:"status"`
+	Contact []string `json:"contact"`
+	Orders  string   `json:"orders,omitempty"`
 }
 
 // An Order is created to request issuance for a CSR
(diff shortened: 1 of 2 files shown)


shred:
I am not sure if this PR breaks `onlyReturnExisting`.

For my test, I create a new account with `{"termsOfServiceAgreed":true}`. After that, I try to create a new account for the same key pair and with `{"onlyReturnExisting":true}`.

I expect that the location

cpu:
@shred You're correct - I overlooked this side effect when I moved the ToS check as part of this PR. Apologies for the regression!

This should be fixed in master now thanks to https://github.com/letsencrypt/pebble/pull/93 and @munnerz :tada:

cpu:
> The only regression I have in my tests now is that when I fetch the account resource, the termsOfServiceAgreed field is now missing (was true previously). The field is optional though, so I don't think it's really a bug.

Agreed. I think I'd prefer to have the server not send that field in the response since for Pebble's there's no case where you could get back a value other than true.

PR #94, opened 2018-02-10
Add support for updating existing user accounts

munnerz:
This PR adds support for a user posting an account update to the account's URL.

Currently, this just reads a new Account object from the request body, copying across the specified Contact field as required.

Section 7.3.2 of the spec however seems to show another kind of structure, one that I cannot see support for currently in Pebble, nor in the (upcoming) Golang ACMEv2 client (https://go-review.googlesource.com/c/crypto/+/86635):

```
7.3.2.  Account Update

   If the client wishes to update this information in the future, it
   sends a POST request with updated information to the account URL.
   The server MUST ignore any updates to "order" fields or any other
   fields it does not recognize.  If the server accepts the update, it
   MUST return a response with a 200 (OK) status code and the resulting
   account object.

   For example, to update the contact information in the above account,
   the client could send the following request:

   POST /acme/acct/1 HTTP/1.1
   Host: example.com
   Content-Type: application/jose+json

   {
     "protected": base64url({
       "alg": "ES256",
       "kid": "https://example.com/acme/acct/1",
       "nonce": "ax5RnthDqp_Yf4_HZnFLmA",
       "url": "https://example.com/acme/acct/1"
     }),
     "payload": base64url({
       "contact": [
         "mailto:certificates@example.com",
         "mailto:admin@example.com"
       ]
     }),
     "signature": "hDXzvcj8T6fbFbmn...rDzXzzvzpRy64N0o"
   }

Servers SHOULD NOT respond to GET requests for account resources as
   these requests are not authenticated.  If a client wishes to query
   the server for information about its account (e.g., to examine the
   "contact" or "certificates" fields), then it SHOULD do so by sending
   a POST request with an empty update.  That is, it should send a JWS
   whose payload is an empty object ({}).
```

/cc @cpu

### DIFF
--- a/db/memorystore.go
+++ b/db/memorystore.go
@@ -42,6 +42,16 @@ func (m *MemoryStore) GetAccountByID(id string) *core.Account {
 	return m.accountsByID[id]
 }
 
+func (m *MemoryStore) UpdateAccountByID(id string, acct *core.Account) error {
+	m.RLock()
+	defer m.RUnlock()
+	if m.accountsByID[id] == nil {
+		return fmt.Errorf("account with ID %q does not exist", id)
+	}
+	m.accountsByID[id] = acct
+	return nil
+}
+
 func (m *MemoryStore) AddAccount(acct *core.Account) (int, error) {
 	m.Lock()
 	defer m.Unlock()
--- a/db/memorystore.go
+++ b/db/memorystore.go
@@ -43,8 +43,8 @@ func (m *MemoryStore) GetAccountByID(id string) *core.Account {
 }
 
 func (m *MemoryStore) UpdateAccountByID(id string, acct *core.Account) error {
-	m.RLock()
-	defer m.RUnlock()
+	m.Lock()
+	defer m.Unlock()
 	if m.accountsByID[id] == nil {
 		return fmt.Errorf("account with ID %q does not exist", id)
 	}
--- a/wfe/wfe.go
+++ b/wfe/wfe.go
@@ -570,7 +570,7 @@ func (wfe *WebFrontEndImpl) UpdateAccount(
 	err := json.Unmarshal(body, &updateAcctReq)
 	if err != nil {
 		wfe.sendError(
-			acme.MalformedProblem("Error unmarshaling body JSON"), response)
+			acme.MalformedProblem("Error unmarshaling account update JSON body"), response)
 		return
 	}
 
--- a/wfe/wfe.go
+++ b/wfe/wfe.go
@@ -224,8 +224,8 @@ func (wfe *WebFrontEndImpl) Handler() http.Handler {
 	wfe.HandleFunc(m, authzPath, wfe.Authz, "GET")
 	wfe.HandleFunc(m, challengePath, wfe.Challenge, "GET", "POST")
 	wfe.HandleFunc(m, certPath, wfe.Certificate, "GET")
+	wfe.HandleFunc(m, acctPath, wfe.UpdateAccount, "POST")
 
-	// TODO(@cpu): Handle POST to acctPath for existing account updates
 	return m
 }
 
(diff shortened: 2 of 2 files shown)


munnerz:
Ah, ignore me, that PR **does** support those extra fields, they are just hidden by a JWS wrapper 😄 

So this PR, combined with #93 will allow a user to call `NewAccount` (either with or without onlyReturnExisting), receive a Location header response, and then look that account up.

db/memorystore.go
```diff
@@ -42,6 +42,16 @@ func (m *MemoryStore) GetAccountByID(id string) *core.Account {
 	return m.accountsByID[id]
 }
 
+func (m *MemoryStore) UpdateAccountByID(id string, acct *core.Account) error {
+	m.RLock()
+	defer m.RUnlock()
+	if m.accountsByID == nil {
+		return fmt.Errorf("account with ID %q does not exist", id)
```
munnerz: D'oh. Just noticed my own mistake here!

wfe/wfe.go
```diff
@@ -552,6 +552,66 @@ func (wfe *WebFrontEndImpl) verifyContacts(acct acme.Account) *acme.ProblemDetai
...
+			Status:  existingAcct.Status,
+			Orders:  existingAcct.Orders,
+		},
+		Key: existingAcct.Key,
+		ID:  existingAcct.ID,
+	}
+
+	// Verify that the contact information provided is supported & valid
+	prob = wfe.verifyContacts(newAcct.Account)
+	if prob != nil {
```
munnerz: Just noticed I also don't ever set these new contact fields... Another d'oh

munnerz:
@cpu updated to address my own review comments 😄

wfe/wfe.go
```diff
@@ -552,6 +552,78 @@ func (wfe *WebFrontEndImpl) verifyContacts(acct acme.Account) *acme.ProblemDetai
...
+
+	existingAcct, prob := wfe.getAcctByKey(key)
+	if prob != nil {
+		wfe.sendError(prob, response)
+		return
+	}
+
+	// if this update contains no contacts, simply return the existing account
+	// and return early.
+	if len(updateAcctReq.Contact) == 0 {
```
munnerz: @cpu if a user were to submit a request that explicitly states `contacts: []`, how should the ACME server behave? Should it throw an error because the user has requested to set no contacts, or should it assume this is a no-op (i.e. assume the user has submitted `{}` as the request body).

I think we can achieve this behaviour through use of `omitempty`, although this change in behaviour could break existing clients (as the spec does not seem to be explicit here).

cpu: We handle this as a no-op in Boulder (equiv to submitting {}) and I think it's fair to keep that here:

https://github.com/letsencrypt/boulder/blob/2dc3b56fa9d68a935259bb199fc17523821cff04/ra/ra.go#L390:L392

munnerz: 👍 that is the current behaviour, but just checking to make sure it's correct! Great 😄

db/memorystore.go
```diff
@@ -42,6 +42,16 @@ func (m *MemoryStore) GetAccountByID(id string) *core.Account {
 	return m.accountsByID[id]
 }
 
+func (m *MemoryStore) UpdateAccountByID(id string, acct *core.Account) error {
+	m.RLock()
```
cpu: I think this needs to be a `m.Lock()` with a `defer r.Unlock()` because you are changing `accountsByID`.

wfe/wfe.go
```diff
@@ -552,6 +552,78 @@ func (wfe *WebFrontEndImpl) verifyContacts(acct acme.Account) *acme.ProblemDetai
...
+	}
+
+	// updateAcctReq is the ACME account information submitted by the client
+	var updateAcctReq struct {
+		Contact []string `json:"contact"`
+	}
+	err := json.Unmarshal(body, &updateAcctReq)
+	if err != nil {
+		wfe.sendError(
+			acme.MalformedProblem("Error unmarshaling body JSON"), response)
```
cpu: nit: "body JSON" would probably be more descriptive if called "account update JSON body".

db/memorystore.go
munnerz: Nice catch - I missed that this was a RWLock!

munnerz:
@cpu review addressed! 🔥

PR #96, opened 2018-03-06
Remove `KeyAuthorization` from challenge updates.

cpu:
Per ACME draft-10 challenge update POST bodies [should no longer include
the `KeyAuthorization` field](https://github.com/ietf-wg-acme/acme/commit/4bdfe043058874080aca94cdcc9a47980012715c), the server can calculate this on its own.
Similarly, it shouldn't be returned in challenge bodies sent to the
client because the client can calculate it themselves.

This commit updates Pebble to *reject* challenge updates with
a `KeyAuthorization` field. This is fairly aggressive: We could simply
ignore this field, but Pebble is meant to encourage good client
behaviour so we'll be more aggressive than Boulder will.

### DIFF
--- a/wfe/wfe.go
+++ b/wfe/wfe.go
@@ -1346,7 +1346,7 @@ func (wfe *WebFrontEndImpl) updateChallenge(
 	}
 
 	var chalResp struct {
-		KeyAuthorization string
+		KeyAuthorization *string
 	}
 	err := json.Unmarshal(body, &chalResp)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


cpu:
This breaks in CI because Chisel2 sends the legacy field:
```
{
   "type": "urn:ietf:params:acme:error:malformedRequest",
   "detail": "Challenge response body contained legacy KeyAuthorzation field, POST body should be `{}`",
   "status": 400
}
```
Fixing that is likely blocked on https://github.com/letsencrypt/boulder/issues/3514 in Boulder.

wfe/wfe.go
```diff
@@ -1356,22 +1345,37 @@ func (wfe *WebFrontEndImpl) updateChallenge(
...
 			acme.MalformedProblem("Error unmarshaling body JSON"), response)
 		return
 	}
 
+	// Historically challenges were updated by POSTing a KeyAuthorization. This is
+	// unnecessary, the server can calculate this itself. We could ignore this if
+	// sent (and that's what Boulder will do) but for Pebble we'd like to be more
+	// aggressive about pushing clients implementations in the right direction, so
+	// we treat this as a malformed request.
+	if chalResp.KeyAuthorization != "" {
```
jsha: Let's make this field a pointer so we can check if it's present at all. Otherwise we'll accept its presence with an empty string.

cpu: Good idea, thanks!

cpu: Resolved in ee20d06

milux:
https://github.com/letsencrypt/pebble/issues/100

Issue #100, opened 2018-03-14
PR #96 boulder-incompatible change enforcement


milux:
> We could ignore this if sent (and that's what Boulder will do) but for Pebble we'd like to be more aggressive about pushing clients implementations in the right direction, so we treat this as a malformed request.

What seems to be a good idea on the first glimpse, as discussed in https://github.com/shred/acme4j/issues/59, doesn't push clients "in the right direction", it enforces LE-/boulder-incompatible changes, which is rather horrible...
I suggest to silently ignore the `KeyAuthorization` field until live LE boulder instances properly support ACME v2 draft 10.

cpu:
I added a flag in https://github.com/letsencrypt/pebble/pull/102 that can be used to selectively enable/disable the strict enforcement of the `keyAuthorization` field. This will let existing code that worked with Pebble continue to work while also providing an easy way to check whether upcoming changes will break client code.

Generally speaking, Pebble is not intended to serve as a 1:1 model for checking compatibility with Boulder and Let's Encrypt's staging/production environments. If that's your goal I recommend you run a Boulder instance to test against. 

The README says this too:
> The goal is to emphasize client specification compatibility and to avoid "over-fitting" on Boulder and the Let's Encrypt production service.

In this case we made a choice to emphasize client specification compatibility and it broke clients that were over-fitting Boulder. I'm happy to add something like `-strict` to make this easier to navigate but the intention is always to make Pebble more aggressive than Boulder in rolling out protocol changes.

milux:
Understood and agreed. Thank you! 👍

cpu:
@milux Thanks for the feedback here and on ACME4J. I appreciate hearing from folks using Pebble.

https://github.com/letsencrypt/pebble/pull/102 is merged so I'm going to close this issue now. Thanks!

PR #101, opened 2018-03-14
Disable enforcement of boulder-incompatible protocol change

milux:
This is a PR for the measure suggested in #100 to bring back LE compatibility.


cpu:
Thanks for the PR, I'm going to close it in favour of https://github.com/letsencrypt/pebble/pull/102. Your PR leaves commented code in branch which is not ideal and also does not allow a way for client developers to check whether the upcoming changes will break their client.

PR #111, opened 2018-04-07
Implement account deactivation.

### LINKED ISSUE #108: Implement account deactivation
Per https://tools.ietf.org/html/draft-ietf-acme-acme-10#section-7.3.7

> A client can deactivate an account by posting a signed update to the
   server with a status field of "deactivated."  Clients may wish to do
   this when the account key is compromised or decommissioned.

Pebble currently seems to ignore an account update with `{"status":"deactivated"}`. The resulting account object returned still has `"status":"valid"`.

I realize this likely falls under the category of stuff like revocation and account key rollover that's not implemented yet. I just wanted to throw it out there for tracking.

calavera:
ACME v11 is not very clear about the problem returned after the account
has been deactivated:

> Once an account is deactivated, the server MUST NOT accept further
> requests authorized by that account's key.

This change assumes that any further request that requires to check the
account will return an unauthorized error. That might be something to
ammend in the ACME spec.

Fixes #108

Signed-off-by: David Calavera <david.calavera@gmail.com>

### DIFF
--- a/wfe/wfe.go
+++ b/wfe/wfe.go
@@ -587,6 +587,7 @@ func (wfe *WebFrontEndImpl) UpdateAccount(
 	// updateAcctReq is the ACME account information submitted by the client
 	var updateAcctReq struct {
 		Contact []string `json:"contact"`
+		Status  string   `json:"status,omitempty"`
 	}
 	err := json.Unmarshal(body, &updateAcctReq)
 	if err != nil {
@@ -601,9 +602,9 @@ func (wfe *WebFrontEndImpl) UpdateAccount(
 		return
 	}
 
-	// if this update contains no contacts, simply return the existing account
-	// and return early.
-	if len(updateAcctReq.Contact) == 0 {
+	// if this update contains no contacts or deactivated status,
+	// simply return the existing account and return early.
+	if len(updateAcctReq.Contact) == 0 && updateAcctReq.Status != acme.StatusDeactivated {
 		err = wfe.writeJsonResponse(response, http.StatusOK, existingAcct)
 		if err != nil {
 			wfe.sendError(acme.InternalErrorProblem("Error marshalling account"), response)
@@ -623,12 +624,22 @@ func (wfe *WebFrontEndImpl) UpdateAccount(
 		ID:  existingAcct.ID,
 	}
 
-	newAcct.Contact = updateAcctReq.Contact
-	// Verify that the contact information provided is supported & valid
-	prob = wfe.verifyContacts(newAcct.Account)
-	if prob != nil {
-		wfe.sendError(prob, response)
+	switch {
+	case updateAcctReq.Status == acme.StatusDeactivated:
+		newAcct.Status = updateAcctReq.Status
+	case updateAcctReq.Status != "" && updateAcctReq.Status != newAcct.Status:
+		wfe.sendError(
+			acme.MalformedProblem(fmt.Sprintf(
+				"Invalid account status: %q", updateAcctReq.Status)), response)
 		return
+	case len(updateAcctReq.Contact) > 0:
+		newAcct.Contact = updateAcctReq.Contact
+		// Verify that the contact information provided is supported & valid
+		prob = wfe.verifyContacts(newAcct.Account)
+		if prob != nil {
+			wfe.sendError(prob, response)
+			return
+		}
 	}
 
 	err = wfe.db.UpdateAccountByID(existingAcct.ID, newAcct)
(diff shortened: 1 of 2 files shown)


wfe/wfe.go
```diff
@@ -623,12 +624,16 @@ func (wfe *WebFrontEndImpl) UpdateAccount(
 		ID:  existingAcct.ID,
 	}
 
-	newAcct.Contact = updateAcctReq.Contact
-	// Verify that the contact information provided is supported & valid
-	prob = wfe.verifyContacts(newAcct.Account)
-	if prob != nil {
-		wfe.sendError(prob, response)
-		return
+	if updateAcctReq.Status == "deactivated" {
```
cpu: I think it would be worth returning a `probs.Malformed` instance if the user POST's an update with `Status != "deactivated"` to be a little bit more strict here.

The way you have it implemented now will ignore an invalid status in an update which isn't unreasonable, but I'd like to shake out as many client errors in Pebble as possible. 

It looks like Boulder is generous and does similar: it only checks that the update status != the existing status: https://github.com/letsencrypt/boulder/blob/699c7e4c4448bf70cd2410504684cc60737cd504/wfe/wfe.go#L1138-L1153 which is fine because ultimately the SA ignores the update status and explicitly uses the deactivated status in `sa.DeactivateRegistration`: https://github.com/letsencrypt/boulder/blob/699c7e4c4448bf70cd2410504684cc60737cd504/sa/sa.go#L1357

calavera:
@cpu I've updated this change with your suggestion and rebased the PR to have only one commit. :raised_hands:

wfe/wfe.go
```diff
@@ -69,6 +69,10 @@ const (
 
 	// POST requests with a JWS body must have the following Content-Type header
 	expectedJWSContentType = "application/jose+json"
+
+	// deactivatedStatus is the only valid status accepted by ACME to update Accounts
+	// and auhtorizations.
+	deactivatedStatus = "deactivated"
```
cpu: Could you add this to `acme/common.go` as a new top level exported status alongside `StatusPending`, `StatusValid`, etc? Only the `wfe` references it presently but I think its best to keep all of the ACME constants in one place where possible.

wfe/wfe.go
```diff
@@ -623,12 +628,22 @@ func (wfe *WebFrontEndImpl) UpdateAccount(
...
-	prob = wfe.verifyContacts(newAcct.Account)
-	if prob != nil {
-		wfe.sendError(prob, response)
+	switch {
+	case updateAcctReq.Status == deactivatedStatus:
+		newAcct.Status = updateAcctReq.Status
+	case updateAcctReq.Status != newAcct.Status:
+		wfe.sendError(
+			acme.MalformedProblem(fmt.Sprintf(
+				"Invalid account status: %s", updateAcctReq.Status)), response)
```
cpu: I think this `%s` should be `%q`, otherwise sending `{ "status": "" }` returns `"Invalid account status: "`

wfe/wfe.go
```diff
@@ -623,12 +628,22 @@ func (wfe *WebFrontEndImpl) UpdateAccount(
...
 
-	newAcct.Contact = updateAcctReq.Contact
-	// Verify that the contact information provided is supported & valid
-	prob = wfe.verifyContacts(newAcct.Account)
-	if prob != nil {
-		wfe.sendError(prob, response)
+	switch {
+	case updateAcctReq.Status == deactivatedStatus:
+		newAcct.Status = updateAcctReq.Status
+	case updateAcctReq.Status != newAcct.Status:
```
cpu: I think this switch case might need to be `updatedAcctReq.Status != "" && updatedAcctReq.Status != newAcct.Status`. 

It looks like as-written this breaks an account update that only changes the contact field, not sending any new status field:

```
$ pebble-client -email foo@bar.com
welcome to the pebble shell
Requesting directory from "https://localhost:14000/dir"
Requesting nonce from "https://localhost:14000/nonce-plz"
Registering new account with "https://localhost:14000/sign-me-up"
Requesting nonce from "https://localhost:14000/nonce-plz"
Your account ID is "https://localhost:14000/my-account/4e74148cfb32e323d86667990bc9b9ce5a28b47cfa6b86f76c09a762cec9f523"
Starting REPL environment...
$> Enter a directory endpoint or a URL to POST: https://localhost:14000/my-account/4e74148cfb32e323d86667990bc9b9ce5a28b47cfa6b86f76c09a762cec9f523
$> Enter JSON body, empty line to finish : 
{ "contact":["mailto:foo@foo.com"] }

Requesting nonce from "https://localhost:14000/nonce-plz"
REPL error: Response 400: {
   "type": "urn:ietf:params:acme:error:malformedRequest",
   "detail": "Invalid account status: ",
   "status": 400
}
```

calavera:
@cpu Updated! Thanks again for the reviews :sun_with_face:

PR #112, opened 2018-04-09
Replace TLS-SNI-02 with TLS-ALPN-01.

mdebski:
This pull request contains a simple TLS-ALPN-01 implementation, as described in draft [0]

[0] https://github.com/rolandshoemaker/acme-tls-alpn

### DIFF
--- a/va/va.go
+++ b/va/va.go
@@ -256,15 +256,30 @@ func (va VAImpl) validateTLSALPN01(task *vaTask) *core.ValidationRecord {
 		ValidatedAt: va.clk.Now(),
 	}
 
-	certs, problem := va.fetchCerts(hostPort, &tls.Config{
-		ServerName: task.Identifier,
-		NextProtos: []string{acme.ACMETLS1Protocol},
+	cs, problem := va.fetchConnectionState(hostPort, &tls.Config{
+		ServerName:         task.Identifier,
+		NextProtos:         []string{acme.ACMETLS1Protocol},
+		InsecureSkipVerify: true,
 	})
 	if problem != nil {
 		result.Error = problem
 		return result
 	}
 
+	if !cs.NegotiatedProtocolIsMutual || cs.NegotiatedProtocol != acme.ACMETLS1Protocol {
+		result.Error = acme.UnauthorizedProblem(fmt.Sprintf(
+			"Cannot connect using %s for %s challenge",
+			acme.ACMETLS1Protocol,
+			acme.ChallengeTLSALPN01,
+		))
+		return result
+	}
+
+	certs := cs.PeerCertificates
+	if len(certs) == 0 {
+		result.Error = acme.UnauthorizedProblem(fmt.Sprintf("No certs presented for %s challenge", acme.ChallengeTLSALPN01))
+		return result
+	}
 	leafCert := certs[0]
 
 	// Verify SNI
@@ -295,9 +310,7 @@ func (va VAImpl) validateTLSALPN01(task *vaTask) *core.ValidationRecord {
 	return result
 }
 
-func (va VAImpl) fetchCerts(hostPort string, config *tls.Config) ([]*x509.Certificate, *acme.ProblemDetails) {
-	config = config.Clone()
-	config.InsecureSkipVerify = true
+func (va VAImpl) fetchConnectionState(hostPort string, config *tls.Config) (*tls.ConnectionState, *acme.ProblemDetails) {
 	conn, err := tls.DialWithDialer(&net.Dialer{Timeout: time.Second * 5}, "tcp", hostPort, config)
 
 	if err != nil {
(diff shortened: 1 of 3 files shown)


va/va.go
```diff
@@ -254,83 +256,54 @@ func (va VAImpl) validateTLSSNI02(task *vaTask) *core.ValidationRecord {
...
-	conn, err := tls.DialWithDialer(
-		&net.Dialer{Timeout: time.Second * 5}, "tcp", hostPort,
-		&tls.Config{
-			ServerName:         sni,
-			InsecureSkipVerify: true,
-		})
+func (va VAImpl) fetchCerts(hostPort string, config *tls.Config) ([]*x509.Certificate, *acme.ProblemDetails) {
+	config = config.Clone()
+	config.InsecureSkipVerify = true
+	conn, err := tls.DialWithDialer(&net.Dialer{Timeout: time.Second * 5}, "tcp", hostPort, config)
```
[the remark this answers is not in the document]
mdebski: Good catch!

va/va.go
```diff
@@ -254,96 +256,76 @@ func (va VAImpl) validateTLSSNI02(task *vaTask) *core.ValidationRecord {
...
-	if len(leafCert.DNSNames) != 2 {
-		names := certNames(leafCert)
-		msg := fmt.Sprintf(
-			"%s challenge certificate doesn't include exactly 2 DNSName entries. "+
-				"Received %d certificate(s), first certificate had names %q",
-			acme.ChallengeTLSSNI02, len(certs), names)
-		return acme.MalformedProblem(msg)
+	if !cs.NegotiatedProtocolIsMutual || cs.NegotiatedProtocol != acme.ACMETLS1Protocol {
+		result.Error = acme.UnauthorizedProblem(fmt.Sprintf(
+			"Cannot connect using %s for %s challenge",
```
cpu: I think this error message should include the acronym "ALPN" somewhere, maybe: `"Cannot mutually negotiate TLS ALPN %q for %s challenge"` ?

va/va.go
```diff
@@ -254,96 +256,76 @@ func (va VAImpl) validateTLSSNI02(task *vaTask) *core.ValidationRecord {
...
-		}
+	certs := cs.PeerCertificates
+	if len(certs) == 0 {
+		result.Error = acme.UnauthorizedProblem(fmt.Sprintf("No certs presented for %s challenge", acme.ChallengeTLSALPN01))
+		return result
 	}
+	leafCert := certs[0]
 
-	if !validSanAName || !validSanBName {
+	// Verify SNI
```
cpu: Can you expand on this comment? I think it would be worth highlighting that you're using `task.Identifier` here and the importance of this check with respect to `acme-tls/1` and the required properties of the challenge type.

va/va.go
```diff
@@ -254,96 +256,76 @@ func (va VAImpl) validateTLSSNI02(task *vaTask) *core.ValidationRecord {
...
+	if len(leafCert.DNSNames) != 1 || !strings.EqualFold(leafCert.DNSNames[0], task.Identifier) {
```
cpu: Is `strings.EqualFold` too liberal for this purpose? I'm not 100% certain that case folding is required/wanted here. Can you expand on why you chose this instead of a direct comparison?

va/va.go
```diff
@@ -254,96 +256,76 @@ func (va VAImpl) validateTLSSNI02(task *vaTask) *core.ValidationRecord {
...
-	return nil
+	// Verify key authorization in acmeValidation extension
+	expectedKeyAuthorization := task.Challenge.ExpectedKeyAuthorization(task.Account.Key)
+	h := sha256.Sum256([]byte(expectedKeyAuthorization))
+	for _, ext := range leafCert.Extensions {
+		if IdPeAcmeIdentifierV1.Equal(ext.Id) && ext.Critical {
+			if subtle.ConstantTimeCompare(h[:], ext.Value) == 1 {
+				return result
+			}
+			result.Error = acme.UnauthorizedProblem("Extension acmeValidationV1 value incorrect.")
```
cpu: I think it might be clearer if this error string matched the format of the `errText` above, e.g: "Incorrect validation certificate for %s challenge. Missing expected acmeValidation1 extension value"

va/va.go
mdebski: SANs are case insensitive by rfcs, I must have seen this in some other code (though I can't find EqualFold anywhere in boulder now).
[0] https://security.stackexchange.com/questions/150770/are-certificate-sans-case-sensitive?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa
[1] https://github.com/globalsign/certlint/blob/033c8b6ba713114a228cd06b0b6a745cbde3efbe/checks/certificate/subjectaltname/subjectaltname.go#L79

va/va.go
cpu: Agreed! Good catch :-)

cpu:
@rolandshoemaker Can you give this another :mag: ?

Issue #98, opened 2018-03-06
Implement order "ready" state


cpu:
Per https://tools.ietf.org/html/draft-ietf-acme-acme-10#section-7.1.6 

> Once all of the authorizations listed in the order object are in the "valid" state, the order transitions to the "ready" state.

felixfontein:
Isn't this about the change described in https://community.letsencrypt.org/t/acmev2-order-ready-status/62866? If yes, is there a reason this issue is still open?

cpu:
@felixfontein No reason except that I forgot to close it! Thanks for the ping.

PR #113, opened 2018-04-09
Set Order status field to invalid/valid based on authz status

### LINKED ISSUE #110: Orders with unvalidated and expired authorizations remain in state 'pending'
Pebble currently appears to set the expires field on a newly created order to 24 hours past when the order was created. The associated authorizations appear to get 1 hour before they expire. No problem there.

However, if the 1 hour time limit elapses on the authorizations without being validated, they seem to remain in a pending state forever. But you get an "Authorization expired" error if you try to POST to a challenge's URL.

I would think that querying the authorization's status returns "invalid" once the expiration time has been reached. I would think that also then sets the order status to "invalid" as well.

Have I misread the spirit of the spec? Or is there a way to generate new authorizations for an existing order?

munnerz:
This is an attempt to fix #110 and #98 

I did contemplate not ever setting the order field and always computing it in the memory store in GetOrderByID, but without a 'DeepCopy' function to create copies of orders, it would mean the orders status in the memory store could be updated in memory when a Get is performed. This seemed a bit perverse in my mind, but perhaps we should just do that instead.

/cc @cpu

### DIFF
--- a/wfe/wfe.go
+++ b/wfe/wfe.go
@@ -999,10 +999,10 @@ func (wfe *WebFrontEndImpl) FinalizeOrder(
 		return
 	}
 
-	// The existing order must be in a pending status to finalize it
-	if orderStatus != acme.StatusPending {
+	// The existing order must be in a ready status to finalize it
+	if orderStatus != acme.StatusReady {
 		wfe.sendError(acme.MalformedProblem(fmt.Sprintf(
-			"Order's status (%q) was not pending", orderStatus)), response)
+			"Order's status (%q) was not 'ready'", orderStatus)), response)
 		return
 	}
 
--- a/acme/common.go
+++ b/acme/common.go
@@ -7,6 +7,7 @@ const (
 	StatusPending    = "pending"
 	StatusInvalid    = "invalid"
 	StatusValid      = "valid"
+	StatusReady      = "ready"
 	StatusProcessing = "processing"
 
 	IdentifierDNS = "dns"
--- a/va/va.go
+++ b/va/va.go
@@ -197,7 +197,7 @@ func (va VAImpl) process(task *vaTask) {
 			}
 		}
 		if !anyNotValid {
-			order.Status = acme.StatusValid
+			order.Status = acme.StatusReady
 		}
 		order.Unlock()
 
--- a/va/va.go
+++ b/va/va.go
@@ -141,6 +141,7 @@ func (va VAImpl) process(task *vaTask) {
 	chal.ValidatedDate = now
 	chal.Validated = chal.ValidatedDate.Format(time.RFC3339)
 	authz := chal.Authz
+	order := authz.Order
 	chal.Unlock()
 
 	results := make(chan *core.ValidationRecord, concurrentValidations)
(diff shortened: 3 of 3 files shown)


wfe/wfe.go
```diff
@@ -999,10 +999,10 @@ func (wfe *WebFrontEndImpl) FinalizeOrder(
...
 
-	// The existing order must be in a pending status to finalize it
-	if orderStatus != acme.StatusPending {
+	// The existing order must be in a ready status to finalize it
+	if orderStatus != acme.StatusReady {
 		wfe.sendError(acme.MalformedProblem(fmt.Sprintf(
-			"Order's status (%q) was not pending", orderStatus)), response)
+			"Order's status (%q) was not 'ready'", orderStatus)), response)
 		return
 	}
```
munnerz: I've changed this in-line with the spec:

```
Once all of the
   authorizations listed in the order object are in the "valid" state,
   the order transitions to the "ready" state.  The order moves to the
   "processing" state after the client submits a request to the order's
   "finalize" URL and the CA begins the issuance process for the
   certificate.  Once the certificate is issued, the order enters the
   "valid" state.
```

Let me know if I've misinterpreted anything here!

munnerz: Looks like master doesn't mention 'ready' either: https://github.com/certbot/certbot/blob/master/acme/acme/messages.py#L139-L147

I'm not sure what Boulder does here 😬 I'd hope it supports 'ready'?

munnerz: Ahhh. The 'ready' state was newly added in v10 of the spec...

Previously (in 09):
```
status (required, string):  The status of this order.  Possible
      values are: "pending", "processing", "valid", and "invalid".
```

Pebble's README states it intends to implement an ACME v09 client. Is this current? Or does it aim to track the latest draft of the spec?

What version of the spec does Boulder implement, and what version does Letsencrypt's v2 staging and production endpoint implement? 😬

cpu: >  Or does it aim to track the latest draft of the spec?

It aims to track the latest.

> What version of the spec does Boulder implement, and what version does Letsencrypt's v2 staging and production endpoint implement?

Basically draft-10 but with a [couple noted divergences](https://community.letsencrypt.org/t/acme-v2-production-environment-wildcards/55578) (See "specification divergences"), one of which is this "Ready" status. I have a PR out on Boulder to fix that one: https://github.com/letsencrypt/boulder/pull/3614

cpu:
> One thought on the design of this PR: It might be easier to approach this like we did in Boulder and remove the status as a persisted order field and instead calculate it on retrieval (see SA.statusForOrder). I think the approach you took here would be OK to land as-is with the above as a follow-up. Mostly mentioning it in case you're interested. Another follow-up to consider is authz expiry, this should also update the order's status to invalid.

@munnerz I didn't hear back from you on the review feedback I left so I went ahead and implemented the design I mentioned above in https://github.com/letsencrypt/pebble/pull/128

I'm going to close this PR for now. It's collected up some conflicts and I think #128 is a more robust fix for  #110 and #98.

Thanks for the PR! I appreciate it!

PR #128, opened 2018-05-04
Handle order status as a property of assoc. authz status.

### LINKED ISSUE #110: Orders with unvalidated and expired authorizations remain in state 'pending'

cpu:
This PR updates the way Pebble handles an Order's status field to closer match the way we handle it in Boulder. This means the order's status is considered to be a calculated field derived based on the order's error/beganProcessing/certificate fields and the status/expires/error of the order's associated authorizations. Along the way the new "Ready" status was implemented. This PR also addresses the problem of the order status not being updated when an authorization is failed. Now, an order has a problem details field that is set when an authorization associated with the order has a failed challenge validation.

Resolves #110 and #98

### DIFF
--- a/core/types.go
+++ b/core/types.go
@@ -28,7 +28,7 @@ type Order struct {
 	CertificateObject    *Certificate
 }
 
-func (o Order) GetStatus(clk clock.Clock) (string, error) {
+func (o *Order) GetStatus(clk clock.Clock) (string, error) {
 	// Lock the order for reading
 	o.RLock()
 	defer o.RUnlock()
--- a/core/types.go
+++ b/core/types.go
@@ -10,6 +10,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/jmhodges/clock"
 	"github.com/letsencrypt/pebble/acme"
 	"gopkg.in/square/go-jose.v2"
 )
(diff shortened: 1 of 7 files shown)


core/types.go
```diff
@@ -23,9 +24,90 @@ type Order struct {
...
+			authzStatuses[acme.StatusExpired]++
+		}
+	}
+
+	// An order is invalid if **any** of its authzs are invalid
+	if authzStatuses[acme.StatusInvalid] > 0 {
+		return acme.StatusInvalid, nil
+	}
+
+	// An order is invalid if **any** of its authzs are expired
```
jsha: *is expired if

cpu: Good catch, thanks! Fixed in 29713fe

PR #130, opened 2018-05-12
Add certificate revocation call to the WFE.

calavera:
Adds an endpoint to the API to revoke certificates.
There are some caveats to consider:

1. It doesn't verify account ownership.
2. It doesn't verify the certificate issuer.
3. It validates the reason, but it doesn't do anything with it after.

See https://tools.ietf.org/html/draft-ietf-acme-acme-11#section-7.6

Signed-off-by: David Calavera <david.calavera@gmail.com>

### DIFF
--- a/db/memorystore.go
+++ b/db/memorystore.go
@@ -2,6 +2,7 @@ package db
 
 import (
 	"fmt"
+	"reflect"
 	"sync"
 
 	"github.com/letsencrypt/pebble/core"
@@ -169,4 +170,24 @@ func (m *MemoryStore) GetCertificateByID(id string) *core.Certificate {
 	m.RLock()
 	defer m.RUnlock()
 	return m.certificatesByID[id]
+}
+
+// GetCertificateByDER loops over all certificates to find the one that matches the provided DER bytes.
+// This method is linear and it's not optimized to give you a quick response.
+func (m *MemoryStore) GetCertificateByDER(der []byte) *core.Certificate {
+	m.RLock()
+	defer m.RUnlock()
+	for _, c := range m.certificatesByID {
+		if reflect.DeepEqual(c.DER, der) {
+			return c
+		}
+	}
+
+	return nil
+}
+
+func (m *MemoryStore) RevokeCertificate(cert *core.Certificate) {
+	m.Lock()
+	defer m.Unlock()
+	delete(m.certificatesByID, cert.ID)
 }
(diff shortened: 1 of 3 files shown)


db/memorystore.go
```diff
@@ -170,3 +171,23 @@ func (m *MemoryStore) GetCertificateByID(id string) *core.Certificate {
 	defer m.RUnlock()
 	return m.certificatesByID[id]
 }
+
+// GetCertificateByDER loops over all certificates to find the one that matches a DER.
```
cpu: nit: "a DER" -> "the provided DER bytes" ?

wfe/wfe.go
```diff
@@ -1618,3 +1620,52 @@ func uniqueLowerNames(names []string) []string {
 	sort.Strings(unique)
 	return unique
 }
+
+func (wfe *WebFrontEndImpl) RevokeCert(
```
cpu: I think the `RevokeCert` function deserves a comment that explains that it is presently implementing only **one** method of ACME revocation: Signing the revocation request by signing it with the certificate to be revoked's private key and embedding the certificate to be revoked's public key as a JWK in the JWS.

wfe/wfe.go
```diff
@@ -1618,3 +1620,52 @@ func uniqueLowerNames(names []string) []string {
...
+	ctx context.Context,
+	logEvent *requestEvent,
+	response http.ResponseWriter,
+	request *http.Request) {
+	body, _, prob := wfe.verifyPOST(ctx, logEvent, request, wfe.lookupJWK)
```
cpu: I think its worth highlighting the reason that `wfe.lookupJWK` is being used here since its an outlier compared to the usual `wfe.verifyPOST` calls used by other endpoints.

wfe/wfe.go
```diff
@@ -1618,3 +1620,52 @@ func uniqueLowerNames(names []string) []string {
...
+	err := json.Unmarshal(body, &revokeCertReq)
+	if err != nil {
+		wfe.sendError(
+			acme.MalformedProblem("Error unmarshaling certificate revocation JSON body"), response)
+		return
+	}
+
+	if revokeCertReq.Reason != nil {
+		r := *revokeCertReq.Reason
+		if r == 7 || r > 10 {
```
cpu: Can you make named constants for these? You can lift them from Boulder: https://github.com/letsencrypt/boulder/blob/e78a7bdb10a4897bd0636f00f63750afb081412d/revocation/reasons.go#L6-L20

wfe/wfe.go
cpu: It might also be a good idea to emphasize that Pebble's idea of "revocation" is to forget the certificate exists. There might be the expectation that the revocation request once processed would percolate to a CRL or an OCSP response and that's not the case for Pebble.

PR #131, opened 2018-05-30
README: Update draft # and cert revocation support.

cpu:


cpu:
> Can we instead document which revocation methods are available?

Sure, will iterate shortly.

Issue #134, opened 2018-06-10
Details on revocation


felixfontein:
I noticed that certificate recovation is now supported in Pebble. Great :-) I've noticed two things:

  1. ACME defines no behavior if one tries to revoke a certificate which has already been revoked. Boulder returns error type `urn:ietf:params:acme:error:malformed` (or `urn:acme:error:malformed` for ACME v1) with detail `Certificate already revoked`, while Pebble simply returns a 404. Are there plans to define a common behavior in the ACME protocol, or will this be left for implementations to handle as they want?

  2. When trying to revoke a certificate with its private key, Boulder does not require the Key ID field, while Pebble does: `{'type': 'urn:ietf:params:acme:error:malformedRequest', 'detail': 'Key ID (kid) in JWS header missing expected URL prefix', 'status': 400}` Which of these behaviors is correct? Or is revocation by certificate's private key simply not supported yet in Pebble?

cpu:
Hi @felixfontein,

> ACME defines no behavior if one tries to revoke a certificate which has already been revoked. Boulder returns error type urn:ietf:params:acme:error:malformed (or urn:acme:error:malformed for ACME v1) with detail Certificate already revoked, while Pebble simply returns a 404. 

Hmm. I think we should probably try to match what Boulder does and return a problem rather than simply an HTTP status code. Thanks for pointing that out!

> Are there plans to define a common behavior in the ACME protocol, or will this be left for implementations to handle as they want?

There aren't any active plans to define a common behaviour. You could start a discussion on the [ACME working group mailing list](https://mailarchive.ietf.org/arch/browse/acme/) about the topic if you would like. I do see that [ACME draft-12 Section 7.6](https://tools.ietf.org/html/draft-ietf-acme-acme-12#section-7.6) says "If the revocation fails, the server returns an error." with an example `urn:ietf:params:acme:error:unauthorized` problem response. Is more specificity about the potential errors needed?

> When trying to revoke a certificate with its private key, Boulder does not require the Key ID field, while Pebble does: {'type': 'urn:ietf:params:acme:error:malformedRequest', 'detail': 'Key ID (kid) in JWS header missing expected URL prefix', 'status': 400} Which of these behaviors is correct? Or is revocation by certificate's private key simply not supported yet in Pebble?

[...]

PR #137, opened 2018-06-11
WFE: Fix revocation authentication.

cpu:
This commit brings over more of Boulder's WFE2 revocation handling to
Pebble and supports **both** authenticating revocation requests with:
* a JWS with a KeyID specifing the account that previously issued the certificate
* a JWS carrying an embedded JWK containing the certificate to be revoked's public key.

Previously the revocation implementation specified it only
supported authenticating the request with the certificate's key and an
embedded JWK. In practice the revocation logic was still expecting a key
ID and authentication from the ACME account. It also did not verify that 
the certificate to be revoked was originally issued by the account that
authenticates the revocation request.

Resolves https://github.com/letsencrypt/pebble/issues/134

### DIFF
--- a/wfe/wfe.go
+++ b/wfe/wfe.go
@@ -1747,7 +1747,9 @@ func (wfe *WebFrontEndImpl) revokeCertByKeyID(
 			return nil
 		}
 		return acme.UnauthorizedProblem(
-			fmt.Sprintf("Account %q did not request issuance of the to-be-revoked certificate"))
+			fmt.Sprintf(
+				"Account %q did not request issuance of the to-be-revoked certificate",
+				existingAcct.ID))
 	}
 	return wfe.processRevocation(ctx, body, authorizedToRevoke, request, logEvent)
 }
--- a/wfe/wfe.go
+++ b/wfe/wfe.go
@@ -1748,7 +1748,7 @@ func (wfe *WebFrontEndImpl) revokeCertByKeyID(
 		}
 		return acme.UnauthorizedProblem(
 			fmt.Sprintf(
-				"Account %q did not request issuance of the to-be-revoked certificate",
+				"The certificate being revoked is not associated with account %q",
 				existingAcct.ID))
 	}
 	return wfe.processRevocation(ctx, body, authorizedToRevoke, request, logEvent)
--- a/wfe/wfe.go
+++ b/wfe/wfe.go
@@ -1615,7 +1615,9 @@ func (wfe *WebFrontEndImpl) RevokeCert(
 	logEvent *requestEvent,
 	response http.ResponseWriter,
 	request *http.Request) {
-	body, _, prob := wfe.verifyPOST(ctx, logEvent, request, wfe.lookupJWK)
+	// We use extractJWK rather than lookupJWK here because the request is
+	// authenticated with an embedded JWK not associated with an account/kid.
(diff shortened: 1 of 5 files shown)


jsha:
Since @calavera submitted #130, which added the account-based authentication for revocation, I assume their automated tests probably rely on that account-based authentication. So I think removing it would probably break their tests. Do you agree? If so, I think we should consider that a blocker on this PR, and only land this if we can also preserve account-based authentication.

cpu:
>  Do you agree? If so, I think we should consider that a blocker on this PR, and only land this if we can also preserve account-based authentication.

I agree, good point. I'll close this PR and resubmit with account based revocation implemented as well. :+1:

felixfontein:
Couldn't most parts of this PR be recycled to implement both checks? I.e. do something along
``` .go
	body, _, prob := wfe.verifyPOST(ctx, logEvent, request, wfe.lookupJWK)
	using_account_key = true;
	if prob != nil {
		// We use extractJWK rather than lookupJWK here because the request is
		// authenticated with an embedded JWK not associated with an account/kid.
		body, key, prob := wfe.verifyPOST(ctx, logEvent, request, wfe.extractJWK)
		if prob != nil {
			wfe.sendError(prob, response)
			return
		}
		using_account_key = false;
	}

	[...]
	
	// Check that the JWK that authenticated the request is equal to the
	// certificate's public key
	if !using_account_key && !keyDigestEquals(key, cert.Cert.PublicKey) {
		wfe.sendError(acme.UnauthorizedProblem(
			"JWK embedded in revocation request must be the same public key as the cert to be revoked"),
			response)
		return
	}
```
in `wfe/wfe.go`? I don't really know the code well enough (and also am not too familiar with Go) and don't have time to try this out right now, so take this with a grain of salt ;-)

cpu:
@felixfontein That approach is ~= what Boulder does. I have a branch mostly finished that ports the logic back to Pebble. I need to find time to test it and then I will submit a PR. Hopefully by ~EOD today.

cpu:
@jsha @calavera @felixfontein I think this PR is ready for :mag:'s again if you folks have time. I fixed the original account based revocation and added the certificate based revocation. It could certainly benefit from some real-world ACME client testing.

wfe/wfe.go
```diff
@@ -1615,48 +1661,176 @@ func (wfe *WebFrontEndImpl) RevokeCert(
...
+	// TODO(@cpu): Allow an account to revoke another account's certificate if
+	// the revoker account has valid authorizations for all of the names in the
+	// to-be-revoked certificate.
+	authorizedToRevoke := func(cert *core.Certificate) *acme.ProblemDetails {
+		if cert.AccountID == existingAcct.ID {
+			return nil
+		}
+		return acme.UnauthorizedProblem(
+			fmt.Sprintf(
+				"Account %q did not request issuance of the to-be-revoked certificate",
```
jsha: "The certificate being revoked is not associated with account %q". The "did not request issuance" is confusing because it sounds like Pebble is complaining that the revocation request isn't an issuance request.

cpu: WFM: Fixed in 6fe0ac0

felixfontein:
Thanks for implementing this! I can say that at least my client now successfully revoked a certificate issued by Pebble using its private key. (So far I tested that only against Let's Encrypt staging, and there it worked as well.)

Issue #33, opened 2017-07-07
Challenge hosts and ports should be configurable


shred:
In order to use Pebble for integration tests, there should be more configuration options for challenges.

At the moment, only the ports for http and tls-sni challenges can be configured. The host should be configurable, too, so all http and tls-sni verifications are always requested from the configured host, irregarding of the domain to be validated.

There are also no configurations for the dns challenge yet. Host name and port of the DNS server should be configurable, so Pebble can be configured to query a mock DNS server running on an unprivileged port.

cpu:
Hi @shred - This is great feedback, thank you!

>  The host should be configurable, too, so all http and tls-sni verifications are always requested from the configured host, irregarding of the domain to be validated.

This sounds like it would overlap with the mock DNS request. Would it be sufficient to provide the ability to specify a mock DNS server and then allow that DNS server to map the validation domain to whatever test host you want to be used for responding to the validation? Maybe I'm misunderstanding this part of your request.

shred:
If you ask me for my oppinion, then no, it doesn't overlap... :smile: For my integration tests, I wrote a very simple mock DNS server that only returns the requested TXT record. I also prefer to keep the http, tls-sni, and dns challenge integration tests separated (this is, for the http challenge I only start a test http server).

But you are right... It would be sufficient if the DNS host and port is configurable.

jsha:
You might be interested in `go get github.com/letsencrypt/boulder/test/dns-test-srv`. It handle both the TXT record challenges and the faking out returning localhost for HTTP and TLS-SNI challenges.

shred:
Thank you, @jsha. The source code of `dns-test-srv` is actually very helpful for extending my mock DNS server. So all I need is a way to configure Pebble's DNS settings.

cpu:
> So all I need is a way to configure Pebble's DNS settings.

I'll try to submit a PR for that in the next couple of days. Assigning this issue to myself as a reminder. Thanks!

cpu:
@shred Having looked into this a little more and talking with @jsha I think the best path forward is to set the system resolver to a fake DNS server. Unfortunately Go's `net` package doesn't provide any method for specifying a resolver to use. Boulder works around this by carrying a [large dependency](https://github.com/miekg/dns) and doing all of its own DNS lookups, wrapping the `net` interfaces as required. We're not super keen on having to do the same for Pebble since it seems reasonable to expect that a test server can be run in an environment with a test system DNS resolver.

[...]

PR #139, opened 2018-06-24
Custom DNS Server Address

ldez:
Add a way to add a custom DNS server address:
- add a flag `dnsserver` 

This allow to use custom port and custom address, ex:
- `192.168.0.56:5053`
- `8.8.8.8:53`

Related to #33

### DIFF
--- a/cmd/pebble/main.go
+++ b/cmd/pebble/main.go
@@ -1,8 +1,10 @@
 package main
 
 import (
+	"context"
 	"flag"
 	"log"
+	"net"
 	"net/http"
 	"os"
 
@@ -16,11 +18,12 @@ import (
 
 type config struct {
 	Pebble struct {
-		ListenAddress string
-		HTTPPort      int
-		TLSPort       int
-		Certificate   string
-		PrivateKey    string
+		ListenAddress    string
+		HTTPPort         int
+		TLSPort          int
+		Certificate      string
+		PrivateKey       string
+		DNSServerAddress string
 	}
 }
 
@@ -33,6 +36,10 @@ func main() {
 		"strict",
 		false,
 		"Enable strict mode to test upcoming API breaking changes")
+	resolverAddress := flag.String(
+		"dnsserver",
+		"",
+		"Define a custom DNS server address (ex: 192.168.0.56:5053 or 8.8.8.8:53). Override the configuration file.")
 	flag.Parse()
 	if *configFile == "" {
 		flag.Usage()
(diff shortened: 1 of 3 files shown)


cpu:
:wave: @ldez. Thanks for the PR! I won't have a chance to take a look at this until tomorrow. Just leaving a note so you know it hasn't been missed :-)

cmd/pebble/main.go
```diff
@@ -62,3 +80,13 @@ func main() {
 		muxHandler)
 	cmd.FailOnError(err, "Calling ListenAndServeTLS()")
 }
+
+func setupCustomDNSResolver(dnsResolverAddress string) {
+	net.DefaultResolver = &net.Resolver{
+		PreferGo: true,
+		Dial: func(ctx context.Context, network, address string) (net.Conn, error) {
```
cpu: nit: I think the `network` parameter can be replaced with `_` since you aren't using it in the `Dial` func.

cmd/pebble/main.go
```diff
@@ -62,3 +80,13 @@ func main() {
 		muxHandler)
 	cmd.FailOnError(err, "Calling ListenAndServeTLS()")
 }
+
+func setupCustomDNSResolver(dnsResolverAddress string) {
+	net.DefaultResolver = &net.Resolver{
+		PreferGo: true,
```
cpu: @ldez Do you know what happens in the case where Go's built-in resolver is not available? It seems like it might quietly fall back to the system resolver and the `net.Dialer` provided below will be ignored. Is that correct? I don't think its a big deal but I'm not sure what systems wouldn't have the Go resolver available.

cmd/pebble/main.go
```diff
@@ -33,6 +36,10 @@ func main() {
 		"strict",
 		false,
 		"Enable strict mode to test upcoming API breaking changes")
+	resolverAddress := flag.String(
```
cpu: I think its probably best to have this as only a config parameter instead of both a config parameter and a command line flag.

ldez: The command line flag is more easy to use in a container context: create a file with a dynamic IP is not really easy to do.

cpu: OK. I don't have strong preferences about whether it be a command line flag or a config parameter, just that it not be both :-)

If the command line flag is easier for the container context can you remove the config parameter?

cmd/pebble/main.go
ldez: yes if the dialer fails, it quietly fall back to the system resolver.
The Go resolver seems available for unix platform and windows.

cmd/pebble/main.go
ldez: it is done :smiley:

cmd/pebble/main.go
```diff
@@ -33,6 +35,10 @@ func main() {
...
+		"dnsserver",
+		"",
+		"Define a custom DNS server address (ex: 192.168.0.56:5053 or 8.8.8.8:53). Override the configuration file.")
```
cpu: I think the "Override the configuration file." part can be removed now :scissors: I edited the PR description to remove the configuration file reference already.

PR #141, opened 2018-06-26
Making sleep time configurable

### LINKED ISSUE #140: Possibility to decrease sleep time (without disabling it)
For integration tests, it would be nice if it is possible to decrease the random sleep intervals (to maybe at most 5 seconds), so that tests will take less long, without removing the sleep intervals completely (by setting `PEBBLE_VA_NOSLEEP` to `1`).

felixfontein:
This introduces a new environment variable `PEBBLE_VA_SLEEPTIME` which allows to set the sleep time to an arbitrary maximum value (instead of just 15). Only positive integers are accepted. (Fixes #140.)

BTW: I noticed that the waiting times were not between 1 and 15 seconds, but between 0 and 14 seconds. My patch doesn't change this, so if you set the maximum to `n`, you will get waits between 0 and `n - 1`. To change this, line 252 in `va.go` should be changed to `len := time.Duration(rand.Intn(va.sleepTime) + 1)`.

### DIFF
--- a/va/va.go
+++ b/va/va.go
@@ -46,9 +46,9 @@ const (
 	noSleepEnvVar = "PEBBLE_VA_NOSLEEP"
 
 	// sleepTimeEnvVar defines the environment variable name used to set the time
-	// VA should sleep between validation attempts (if not disabled). Set this
-	// e.g. to 5 when you invoke Pebble if you wish the delays to be between 1
-	// and 5 seconds (instead between 1 and 15 seconds):
+	// the VA should sleep between validation attempts (if not disabled). Set this
+	// e.g. to 5 when you invoke Pebble if you wish the delays to be between 0
+	// and 5 seconds (instead between 0 and 15 seconds):
 	//   PEBBLE_VA_SLEEPTIME=5 pebble
 	sleepTimeEnvVar = "PEBBLE_VA_SLEEPTIME"
 
@@ -247,8 +247,7 @@ func (va VAImpl) process(task *vaTask) {
 
 func (va VAImpl) performValidation(task *vaTask, results chan<- *core.ValidationRecord) {
 	if va.sleep {
-		// Sleep for a random amount of time between 1-15s (where 15 can be
-		// changed by modifying va.sleepTime)
+		// Sleep for a random amount of time between 0 and va.sleepTime seconds
 		len := time.Duration(rand.Intn(va.sleepTime))
 		va.log.Printf("Sleeping for %s seconds before validating", time.Second*len)
 		va.clk.Sleep(time.Second * len)
--- a/va/va.go
+++ b/va/va.go
@@ -45,6 +45,13 @@ const (
 	//   PEBBLE_VA_NOSLEEP=1 pebble
 	noSleepEnvVar = "PEBBLE_VA_NOSLEEP"
 
+	// sleepTimeEnvVar defines the environment variable name used to set the time
+	// VA should sleep between validation attempts (if not disabled). Set this
+	// e.g. to 5 when you invoke Pebble if you wish the delays to be between 1
+	// and 5 seconds (instead between 1 and 15 seconds):
+	//   PEBBLE_VA_SLEEPTIME=5 pebble
+	sleepTimeEnvVar = "PEBBLE_VA_SLEEPTIME"
+
 	// noValidateEnvVar defines the environment variable name used to signal that
 	// the VA should *not* actually validate challenges. Set this to 1 when you
 	// invoke Pebble if you wish validation to always succeed without actually
@@ -85,6 +92,7 @@ type VAImpl struct {
 	tlsPort     int
 	tasks       chan *vaTask
 	sleep       bool
+	sleepTime   int
 	alwaysValid bool
 }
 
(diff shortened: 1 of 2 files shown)


va/va.go
```diff
@@ -45,6 +45,13 @@ const (
 	//   PEBBLE_VA_NOSLEEP=1 pebble
 	noSleepEnvVar = "PEBBLE_VA_NOSLEEP"
 
+	// sleepTimeEnvVar defines the environment variable name used to set the time
+	// VA should sleep between validation attempts (if not disabled). Set this
```
cpu: nit: "the time VA" -> "the time the VA"

va/va.go
```diff
@@ -45,6 +45,13 @@ const (
...
+	// e.g. to 5 when you invoke Pebble if you wish the delays to be between 1
```
cpu: I think this comment should say "between 0 and 5 seconds" and "instead of between 0 and 15 seconds" since as you point out in your PR description a sleep of 0s may occur.

va/va.go
```diff
@@ -93,12 +101,13 @@ func New(
...
-		tlsPort:  tlsPort,
-		tasks:    make(chan *vaTask, taskQueueSize),
-		sleep:    true,
+		log:       log,
+		clk:       clk,
+		httpPort:  httpPort,
+		tlsPort:   tlsPort,
+		tasks:     make(chan *vaTask, taskQueueSize),
+		sleep:     true,
+		sleepTime: 15,
```
cpu: Can you make a `const` for the default value? Thanks!

va/va.go
```diff
@@ -110,6 +119,12 @@ func New(
 		va.log.Printf("Disabling random VA sleeps")
 	}
 
+	sleepTime := os.Getenv(sleepTimeEnvVar)
+	sleepTimeInt, err := strconv.Atoi(sleepTime)
+	if err == nil && sleepTimeInt >= 1 {
+		va.sleepTime = sleepTimeInt
```
cpu: Can you add a `va.log.Printf` that outputs the `va.sleepTime` when it is overridden?

va/va.go
```diff
@@ -232,8 +247,9 @@ func (va VAImpl) process(task *vaTask) {
 
 func (va VAImpl) performValidation(task *vaTask, results chan<- *core.ValidationRecord) {
 	if va.sleep {
-		// Sleep for a random amount of time between 1-15s
-		len := time.Duration(rand.Intn(15))
+		// Sleep for a random amount of time between 1-15s (where 15 can be
```
cpu: I think you can delete this comment outright. It's more confusing now with when the `va.sleepTime` might not always be 15s and I think the code is clear enough without the comment.

va/va.go
felixfontein: Thanks for noticing!

va/va.go
felixfontein: I've also adjusted the README.md, which also said "(from 1 to 15)".

va/va.go
felixfontein: I changed it to `// Sleep for a random amount of time between 0 and va.sleepTime seconds`

va/va.go
felixfontein: Done. I called it `defaultSleepTime`; if you want the `default` removed from the name, tell me.

va/va.go
felixfontein: Sure.

va/va.go
```diff
@@ -45,6 +45,18 @@ const (
...
 
+	// sleepTimeEnvVar defines the environment variable name used to set the time
+	// the VA should sleep between validation attempts (if not disabled). Set this
+	// e.g. to 5 when you invoke Pebble if you wish the delays to be between 0
+	// and 5 seconds (instead between 0 and 15 seconds):
+	//   PEBBLE_VA_SLEEPTIME=5 pebble
+	sleepTimeEnvVar = "PEBBLE_VA_SLEEPTIME"
+
+	// defaultSleepTime defines the default sleep time (in seconds) between
+	// validation attempts. Can be disabled resp. modified by the environment
```
cpu: Is "Can be disabled resp. modified by the environment" a typo? Should it read "Can be disabled or modified by the environment"

va/va.go
```diff
@@ -110,6 +124,13 @@ func New(
 		va.log.Printf("Disabling random VA sleeps")
 	}
 
+	sleepTime := os.Getenv(sleepTimeEnvVar)
+	sleepTimeInt, err := strconv.Atoi(sleepTime)
+	if err == nil && sleepTimeInt >= 1 {
```
cpu: Sorry I missed this on the first pass but I think this `if` clause should also include `!noSleep` to ensure that the sleep time isn't overridden/logged when it won't be used because `noSleepEnvVar` was present.

va/va.go
```diff
@@ -110,6 +124,13 @@ func New(
...
+		va.sleepTime = sleepTimeInt
+		va.log.Printf("Setting maximal random VA sleep time to %d seconds", va.sleepTime)
```
cpu: small nit: I think "maximum" reads a little bit more naturally than "maximal"

va/va.go
felixfontein: Yes, it should have been *or*. I've moved the *resp* to where the *and* was, I hope it is more correct now :)

va/va.go
felixfontein: I added `!va.sleep`, since `noSleepEnvVar` can be present and have the value `0` (or `false` or whatever's not on the explicit truish list), in which case `noValidateEnvVar` still makes sense.

va/va.go
felixfontein: I get these confused regularly... Thanks for pointing this out!

felixfontein:
@cpu, thanks for handling this!
Unfortunately, I did something stupid yesterday evening: the last change now actually prevents to change the sleeping time, **except** if sleeping is disabled. It should have been `va.sleep`, not `!va.sleep`.
I'll create a new PR to fix this...

PR #142, opened 2018-06-28
Fixing bad boolean condition introduced in #141

felixfontein:
This is a fixup for #141. With this fix, the new feature from that PR actually works as advertised...


felixfontein:
I wanted to try it out, and wondered for a few minutes why it didn't work. The fix was then pretty easy ;-)
Thanks for also handling this, and thanks for the reviews!

PR #143, opened 2018-07-01
Add a Dockerfile.

ldez:
This PR adds a simple Dockerfile.

In the case of running tests for a client with Docker, it avoids having to build the image during the execution of the tests or maintain a fork to build an image for the Docker Hub.

If this PR interests you, could you activate the automatic builds from the Docker Hub so that the image is built automatically? (this would allow to easily run the tests with a Pebble always up to date)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,23 @@
+FROM golang:1.10-alpine as builder
+
+RUN apk --update upgrade \
+&& apk --no-cache --no-progress add git mercurial bash gcc musl-dev curl tar \
+&& rm -rf /var/cache/apk/*
+
+WORKDIR /go/src/github.com/letsencrypt/pebble
+COPY . .
+
+RUN go get -u github.com/jmhodges/clock \
+&& go get -u gopkg.in/square/go-jose.v2
+
+RUN go install ./...
+
+## main
+FROM alpine:3.7
+
+RUN apk update && apk add --no-cache --virtual ca-certificates
+
+COPY --from=builder /go/bin/pebble /usr/bin/pebble
+COPY --from=builder /go/src/github.com/letsencrypt/pebble/test/ /test/
+
+ENTRYPOINT [ "/usr/bin/pebble" ]
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,4 @@
+.git
+pebble.exe
+pebble
+vendor/
--- a/README.md
+++ b/README.md
@@ -1,5 +1,8 @@
 # Pebble
 
+[![Build Status](https://travis-ci.org/letsencrypt/pebble.svg?branch=master)](https://travis-ci.org/letsencrypt/pebble)
+[![Docker Build Status](https://img.shields.io/docker/build/letsencrypt/pebble.svg)](https://hub.docker.com/r/letsencrypt/pebble/builds/)
+
 A miniature version of [Boulder](https://github.com/letsencrypt/boulder), Pebble
 is a small [ACME-12](https://tools.ietf.org/html/draft-ietf-acme-acme-12) test
 server not suited for use as a production CA.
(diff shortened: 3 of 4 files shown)


Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+## main
+FROM alpine:3.7
+
+RUN apk update && apk add --no-cache --virtual ca-certificates
+
+COPY --from=builder /go/bin/pebble /usr/bin/pebble
+COPY --from=builder /go/src/github.com/letsencrypt/pebble/test/ /test/
+
+ENTRYPOINT [ "/usr/bin/pebble" ]
```
cpu: I think this needs a `--config /test/config.json` no? Otherwise I think `pebble` will default to trying to read from `test/config/pebble-config.json` in the pwd.

Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+&& apk --no-cache --no-progress add git mercurial bash gcc musl-dev curl tar \
+&& rm -rf /var/cache/apk/*
+
+WORKDIR /go/src/github.com/letsencrypt/pebble
+COPY . .
+
+RUN go get -u github.com/jmhodges/clock \
+&& go get -u gopkg.in/square/go-jose.v2
+
+RUN go install ./...
```
cpu: Could you make the Builder script also run the unit tests? Is that an anti-pattern?

cpu:
@shred :wave: Would you be interested in taking a quick :eyes: at this PR? Would it help you remove any of the Pebble Docker definitions in the ACME4J `pom.xml`? https://github.com/shred/acme4j/blob/b9aeb816152d98752cf9f9be15be857865dda1f7/acme4j-it/pom.xml#L120-L170

Dockerfile
ldez: The purpose of a Dockerfile is not to replace a CI.
A Dockerfile must contain only what allows to build.

Dockerfile
cpu: Ahh, and PWD == `/` in this case so everything works :+1: 

Thanks for clarifying.

ldez:
Here is a small example with a docker-compose file:

```yml
pebble:
  image: letsencrypt/pebble
  command: -dnsserver ${CUSTOM_DNS_IP}:5053
  ports:
    - 14000:14000
  environment:
    - PEBBLE_VA_NOSLEEP=1
    - PEBBLE_WFE_NONCEREJECT=30
```

And with a Docker:
```shell
docker run -e "PEBBLE_VA_NOSLEEP=1"  letsencrypt/pebble -dnsserver 10.10.10.10:5053
```

ldez:
Documentation added in 3691aa9

shred:
@cpu Thanks for the pointer. This PR looks good, and I could use an automated build of Pebble. But I prefer to keep the current way for now, because with a few modifications to the pom file, I can also use my own Pebble fork for tests.

cpu:
> This PR looks good, and I could use an automated build of Pebble. But I prefer to keep the current way for now, because with a few modifications to the pom file, I can also use my own Pebble fork for tests.

@shred Thanks for taking a look, appreciated! Totally understand the desire to keep your setup as-is. If its working don't break it :-) You were just the first person I thought of that I knew was using Pebble with Docker and wanted to see if this PR made sense to you. Thanks!

cpu:
@ldez I will try to activate the automatic Dockerhub builds soon. I created a repository (https://hub.docker.com/r/letsencrypt/pebble/) but haven't been able to create an automatic build yet. I'm having a hard time getting the UI to let me create a build for a repo in the Let's Encrypt GH organization even though I'm a member.

Dockerfile
jsha: This should probably be `CMD` rather than `ENTRYPOINT`. The difference: If you pass an alternate command to `docker run`, it gets passed as an argument to `ENTRYPOINT`, but it overrides `CMD`. So for instance if you wanted to run `bash` in the context of this container it would not work right due to the `ENTRYPOINT`.

Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM golang:1.10-alpine as builder
+
+RUN apk --update upgrade \
+&& apk --no-cache --no-progress add git mercurial bash gcc musl-dev curl tar \
+&& rm -rf /var/cache/apk/*
+
+WORKDIR /go/src/github.com/letsencrypt/pebble
+COPY . .
+
+RUN go get -u github.com/jmhodges/clock \
```
jsha: This and the line below can be condensed into `go get ./...`, since `go get` fetches dependencies recursively.

Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM golang:1.10-alpine as builder
+
+RUN apk --update upgrade \
+&& apk --no-cache --no-progress add git mercurial bash gcc musl-dev curl tar \
```
jsha: What are mercurial, musl-dev, and tar used for?

felixfontein:
@cpu Are you planning to use tags for versioning, so it is possible to access a specific version (that would be a git commit ID, I guess) of Pebble?

cpu:
> @cpu Are you planning to use tags for versioning, so it is possible to access a specific version (that would be a git commit ID, I guess) of Pebble?

@felixfontein Do you want to start an issue to sort this out? I haven't given it a lot of thought and would be interested in discussing options.

felixfontein:
@cpu I've created a new issue: #146

PR #149, opened 2018-07-27
Distinguish between empty contact list and not specified contact list

### LINKED ISSUE #114: Attempts to clear contact on an account don't succeed and don't throw error.
I'm trying to implement a feature in my client to remove an account's contact information.

Per https://tools.ietf.org/html/draft-ietf-acme-acme-11#section-7.3.2

> If the client wishes to update this information in the future, it
   sends a POST request with updated information to the account URL.
   The server MUST ignore any updates to the "orders" field or any other
   fields it does not recognize.  If the server accepts the update, it
   MUST return a response with a 200 (OK) status code and the resulting
   account object.

I've tried two different methods to empty the contact field. Within the payload of the request: 

- `"contact":null`
- `"contact":[]`

Both of these return a 200 (OK) status as if they were accepted. But the resulting account object still has the old value for the contact field.

I tried testing against the Boulder v2 staging server as well. `"contact":[]` works and does clear the field. `"contact":null` works the same as Pebble and doesn't modify the account or throw an error.

It would be awesome if both methods worked. But I can settle for only `"contact":[]` working. Though, perhaps it would be more appropriate to generate an error on `"contact":null` in that case?

felixfontein:
When updating an account, this PR changes the behavior such that an empty contact list is treated differently than a not specified one (or one which is set to `null`, which cannot be distinguished easily from a not specified one). This allows clients to actually delete the contacts list.

Fixes #114.

I haven't got around to properly test this, so please either test this before merging, or wait for me to do proper testing ;-)

### DIFF
--- a/wfe/wfe.go
+++ b/wfe/wfe.go
@@ -669,7 +669,7 @@ func (wfe *WebFrontEndImpl) UpdateAccount(
 
 	// if this update contains no contacts or deactivated status,
 	// simply return the existing account and return early.
-	if len(updateAcctReq.Contact) == 0 && updateAcctReq.Status != acme.StatusDeactivated {
+	if updateAcctReq.Contact == nil && updateAcctReq.Status != acme.StatusDeactivated {
 		err = wfe.writeJsonResponse(response, http.StatusOK, existingAcct)
 		if err != nil {
 			wfe.sendError(acme.InternalErrorProblem("Error marshalling account"), response)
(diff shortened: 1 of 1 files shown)


felixfontein:
Ok; I've tested it, and it seems to work fine.

felixfontein:
@cpu good to hear ;-) Thanks for testing and handling!

PR #151, opened 2018-07-27
Adding account key roll-over

### LINKED ISSUE #150: Account key roll-over
Pebble does not support account key rollover.

(I know this is stated in the README, but I think it would be nice to have an issue for this as well :-) )

felixfontein:
Adds [ACME-13](https://tools.ietf.org/html/draft-ietf-acme-acme-13#section-7.3.6) account key roll-over to Pebble.

Fixes #150.

### DIFF
--- a/db/memorystore.go
+++ b/db/memorystore.go
@@ -18,12 +18,11 @@ import (
 
 // Error when key change fails because new account key has already been in use
 type ExistingAccountError struct {
-	msg             string
 	MatchingAccount *core.Account
 }
 
-func (e *ExistingAccountError) Error() string {
-	return e.msg
+func (e ExistingAccountError) Error() string {
+	return fmt.Sprintf("New public key is already in use by account %s", e.MatchingAccount.ID)
 }
 
 
@@ -145,7 +144,7 @@ func (m *MemoryStore) ChangeAccountKey(acct *core.Account, newKey *jose.JSONWebK
 	}
 
 	if otherAccount, present := m.accountsByKeyID[newKeyID]; present {
-		return &ExistingAccountError{"New public key is already in use", otherAccount}
+		return ExistingAccountError{otherAccount}
 	}
 
 	delete(m.accountsByKeyID, oldKeyID)
--- a/db/memorystore.go
+++ b/db/memorystore.go
@@ -71,6 +71,9 @@ func (m *MemoryStore) GetAccountByKey(key crypto.PublicKey) (*core.Account, erro
 	return m.accountsByKeyID[keyID], nil
 }
 
+// Note that this function should *NOT* be used for key changes. It assumes
+// the public key associated to the account does not change. Use ChangeAccountKey
+// to change the account's public key.
 func (m *MemoryStore) UpdateAccountByID(id string, acct *core.Account) error {
 	m.Lock()
 	defer m.Unlock()
--- a/db/memorystore.go
+++ b/db/memorystore.go
@@ -116,6 +116,31 @@ func (m *MemoryStore) AddAccount(acct *core.Account) (int, error) {
 	return len(m.accountsByID), nil
 }
 
+func (m *MemoryStore) ChangeAccountKey(acct *core.Account, newKey *jose.JSONWebKey) error {
+	m.Lock()
+	defer m.Unlock()
+
+	oldKeyID, err := keyToID(acct.Key)
+	if err != nil {
+		return err
+	}
+
+	newKeyID, err := keyToID(newKey)
+	if err != nil {
+		return err
+	}
+
+	if _, present := m.accountsByKeyID[newKeyID]; present {
+		return fmt.Errorf("New public key is already in use")
+	}
+
+	delete(m.accountsByKeyID, oldKeyID)
+	acct.Key = newKey
+	m.accountsByKeyID[newKeyID] = acct
+	m.accountsByID[acct.ID] = acct
+	return nil
+}
+
 func (m *MemoryStore) AddOrder(order *core.Order) (int, error) {
 	m.Lock()
 	defer m.Unlock()
--- a/db/memorystore.go
+++ b/db/memorystore.go
@@ -116,29 +116,29 @@ func (m *MemoryStore) AddAccount(acct *core.Account) (int, error) {
 	return len(m.accountsByID), nil
 }
 
-func (m *MemoryStore) ChangeAccountKey(acct *core.Account, newKey *jose.JSONWebKey) error {
+func (m *MemoryStore) ChangeAccountKey(acct *core.Account, newKey *jose.JSONWebKey) (error, *core.Account) {
 	m.Lock()
 	defer m.Unlock()
 
 	oldKeyID, err := keyToID(acct.Key)
 	if err != nil {
-		return err
+		return err, nil
 	}
 
 	newKeyID, err := keyToID(newKey)
 	if err != nil {
-		return err
+		return err, nil
 	}
 
-	if _, present := m.accountsByKeyID[newKeyID]; present {
-		return fmt.Errorf("New public key is already in use")
+	if otherAccount, present := m.accountsByKeyID[newKeyID]; present {
+		return fmt.Errorf("New public key is already in use"), otherAccount
 	}
 
 	delete(m.accountsByKeyID, oldKeyID)
 	acct.Key = newKey
 	m.accountsByKeyID[newKeyID] = acct
 	m.accountsByID[acct.ID] = acct
-	return nil
+	return nil, nil
 }
 
 func (m *MemoryStore) AddOrder(order *core.Order) (int, error) {
--- a/db/memorystore.go
+++ b/db/memorystore.go
@@ -16,6 +16,17 @@ import (
 	"github.com/letsencrypt/pebble/core"
 )
 
+// Error when key change fails because new account key has already been in use
+type ExistingAccountError struct {
+	msg             string
+	MatchingAccount *core.Account
+}
+
+func (e *ExistingAccountError) Error() string {
+	return e.msg
+}
+
+
 // Pebble keeps all of its various objects (accounts, orders, etc)
 // in-memory, not persisted anywhere. MemoryStore implements this in-memory
 // "database"
@@ -119,29 +130,29 @@ func (m *MemoryStore) AddAccount(acct *core.Account) (int, error) {
 	return len(m.accountsByID), nil
 }
 
-func (m *MemoryStore) ChangeAccountKey(acct *core.Account, newKey *jose.JSONWebKey) (error, *core.Account) {
+func (m *MemoryStore) ChangeAccountKey(acct *core.Account, newKey *jose.JSONWebKey) error {
 	m.Lock()
 	defer m.Unlock()
 
 	oldKeyID, err := keyToID(acct.Key)
 	if err != nil {
-		return err, nil
+		return err
 	}
 
 	newKeyID, err := keyToID(newKey)
 	if err != nil {
-		return err, nil
+		return err
 	}
 
 	if otherAccount, present := m.accountsByKeyID[newKeyID]; present {
-		return fmt.Errorf("New public key is already in use"), otherAccount
+		return &ExistingAccountError{"New public key is already in use", otherAccount}
 	}
 
 	delete(m.accountsByKeyID, oldKeyID)
 	acct.Key = newKey
 	m.accountsByKeyID[newKeyID] = acct
 	m.accountsByID[acct.ID] = acct
-	return nil, nil
+	return nil
 }
 
 func (m *MemoryStore) AddOrder(order *core.Order) (int, error) {
--- a/db/memorystore.go
+++ b/db/memorystore.go
@@ -16,7 +16,8 @@ import (
 	"github.com/letsencrypt/pebble/core"
 )
 
-// Error when key change fails because new account key has already been in use
+// ExistingAccountError is an error type indicating when an operation fails
+// because the MatchingAccount has a key conflict.
 type ExistingAccountError struct {
 	MatchingAccount *core.Account
 }
--- a/wfe/wfe.go
+++ b/wfe/wfe.go
@@ -776,16 +776,16 @@ func (wfe *WebFrontEndImpl) KeyRollover(
 	// Check account ID
 	prefix := wfe.relativeEndpoint(request, acctPath)
 	if !strings.HasPrefix(innerContent.Account, prefix) {
-		wfe.sendError(acme.MalformedProblem("Key ID (account) in inner JWS body missing expected URL prefix"), response)
+		wfe.sendError(acme.MalformedProblem(fmt.Sprintf("Key ID (account) in inner JWS body missing expected URL prefix (provided account value: %q)", innerContent.Account)), response)
 		return
 	}
 	accountID := strings.TrimPrefix(innerContent.Account, prefix)
 	if accountID == "" {
-		wfe.sendError(acme.MalformedProblem("No key ID (account) in inner JWS body"), response)
+		wfe.sendError(acme.MalformedProblem(fmt.Sprintf("No key ID (account) in inner JWS body (provided account value: %q)", innerContent.Account)), response)
 		return
 	}
 	if accountID != existingAcct.ID {
-		wfe.sendError(acme.MalformedProblem("Key roll-over inner JWS body contains wrong account ID"), response)
+		wfe.sendError(acme.MalformedProblem(fmt.Sprintf("Key roll-over inner JWS body contains wrong account ID (provided account value: %q)", innerContent.Account)), response)
 		return
 	}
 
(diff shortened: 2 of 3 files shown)


felixfontein:
Ok, there's a major downside with this: currently, account IDs are computed from the account key, so a key change also changes the account ID. This is not a good idea, I think, and clients storing the account ID will have problems with this.

Idea:
 1. Put this PR on hold;
 2. Create another PR to detach account IDs from keys.

For 2., the account ID should probably be some random string. I guess the DB should do the mapping between keys and account IDs. Is that OK? If yes, I can start working on that.

Another (minor) point with this PR: it doesn't check whether the new key is already used for an account.

cpu:
:wave: Hi @felixfontein, thanks for the PR & accompanying thoughts.

> Ok, there's a major downside with this: currently, account IDs are computed from the account key, so a key change also changes the account ID. This is not a good idea, I think, and clients storing the account ID will have problems with this.

Hmmm. That is a good point! I don't think we had this consequence in mind when we made the choice to identify accounts this way.

> For 2., the account ID should probably be some random string. I guess the DB should do the mapping between keys and account IDs. Is that OK? If yes, I can start working on that.

Yes, this sounds good. I'm in agreement with you that holding out on this PR until the related account ID problem is fixed is a good course of action. Thanks so much for offering to work on a fix :-)

> Another (minor) point with this PR: it doesn't check whether the new key is already used for an account.

I think that's probably worth addressing before we land this code since it could end up surprising folks and performing this check is a MUST for server operators in the draft specification.

felixfontein:
@cpu I rebased against a squashed version of #154 (Using account IDs different from key IDs) and adjusted the PR to behave correctly ;) I added the missing existing account public key check, and also a check for the URL parameters of the inner and outer JWS being the same. I think the PR now behaves as expected :)

As soon as #154 is merged, I'll rebase against the new master to get rid of the commits from #154.

cpu:
> As soon as #154 is merged, I'll rebase against the new master to get rid of the commits from #154.

@felixfontein All set for :point_up: 

I'll try to give this a review pass today but it might slip until next week. Thanks again!

felixfontein:
@cpu Great! I've just rebased it to current `master`, so it's ready for review :)

db/memorystore.go
```diff
@@ -116,6 +119,31 @@ func (m *MemoryStore) AddAccount(acct *core.Account) (int, error) {
 	return len(m.accountsByID), nil
 }
 
+func (m *MemoryStore) ChangeAccountKey(acct *core.Account, newKey *jose.JSONWebKey) (error, *core.Account) {
```
cpu: Can you swap the order of the `*core.Account` and `error` returns? It's common Go convention to always return the `error` type as the last return value.

db/memorystore.go
```diff
@@ -116,6 +119,31 @@ func (m *MemoryStore) AddAccount(acct *core.Account) (int, error) {
...
+
+	if otherAccount, present := m.accountsByKeyID[newKeyID]; present {
+		return fmt.Errorf("New public key is already in use"), otherAccount
+	}
+
+	delete(m.accountsByKeyID, oldKeyID)
+	acct.Key = newKey
+	m.accountsByKeyID[newKeyID] = acct
+	m.accountsByID[acct.ID] = acct
+	return nil, nil
```
cpu: I think this should be just `return nil` once the refactoring I describe happens.

wfe/wfe.go
```diff
@@ -712,6 +715,102 @@ func (wfe *WebFrontEndImpl) UpdateAccount(
...
+		wfe.sendError(acme.MalformedProblem(err.Error()), response)
+		return
+	}
+
+	innerHeaderURL, ok := parsedInnerJWS.Signatures[0].Header.ExtraHeaders[jose.HeaderKey("url")].(string)
+	if !ok || len(innerHeaderURL) == 0 {
+		wfe.sendError(acme.MalformedProblem("JWS header parameter 'url' required for inner JWS."), response)
+	}
+	if innerHeaderURL != outerHeaderURL {
+		wfe.sendError(acme.MalformedProblem("JWS header parameter 'url' differ for inner and outer JWS."), response)
```
cpu: nit: I think "differ" should be plural, e.g. "JWS header parameter 'url' differs for inner and outer JWS."

wfe/wfe.go
```diff
@@ -712,6 +715,102 @@ func (wfe *WebFrontEndImpl) UpdateAccount(
...
+	// TODO(@cpu): `checkAlgorithm()`
+
+	innerPayload, err := parsedInnerJWS.Verify(newPubKey)
+	if err != nil {
+		wfe.sendError(acme.MalformedProblem("JWS verification error"), response)
+		return
+	}
+
+	var innerContent struct {
+		Account string `json:"account"`
```
cpu: You can get rid of the `json` tags on both of these fields since their name matches the JSON keys exactly and you aren't specifying any additional options in the tags.

wfe/wfe.go
```diff
@@ -712,6 +715,102 @@ func (wfe *WebFrontEndImpl) UpdateAccount(
...
+		OldKey  jose.JSONWebKey `json:"oldKey"`
+	}
+	err = json.Unmarshal([]byte(innerPayload), &innerContent)
+	if err != nil {
+		wfe.sendError(acme.MalformedProblem("Error unmarshaling key roll-over innermost JSON body"), response)
```
cpu: I think we use the term "inner JWS body" in Boulder. Can you use that here as well instead of "innermost JSON body"?

wfe/wfe.go
```diff
@@ -712,6 +715,102 @@ func (wfe *WebFrontEndImpl) UpdateAccount(
...
+		return
+	}
+
+	// Check account ID
+	prefix := wfe.relativeEndpoint(request, acctPath)
+	if !strings.HasPrefix(innerContent.Account, prefix) {
+		wfe.sendError(acme.MalformedProblem("Key ID (account) in innermost JSON body missing expected URL prefix"), response)
```
cpu: ditto: re innermost JSON body

wfe/wfe.go
```diff
@@ -712,6 +715,102 @@ func (wfe *WebFrontEndImpl) UpdateAccount(
...
+		return
+	}
+	accountID := strings.TrimPrefix(innerContent.Account, prefix)
+	if accountID == "" {
+		wfe.sendError(acme.MalformedProblem("No key ID (account) in innermost JSON body"), response)
```
cpu: Can you include the raw `innerContent.Account` as a `%q` format specifier somewhere in the error message? That will make debugging easier for client developers.

cpu: (also another place to fix "innermost JSON body")

wfe/wfe.go
```diff
@@ -712,6 +715,102 @@ func (wfe *WebFrontEndImpl) UpdateAccount(
...
+		return
+	}
+	if accountID != existingAcct.ID {
+		wfe.sendError(acme.MalformedProblem("Key roll-over innermost JSON contains wrong account ID"), response)
```
cpu: Can you include the raw innerContent.Account as a %q format specifier somewhere in the error message here too? (Also another "innermost JSON" to replace)

wfe/wfe.go
```diff
@@ -712,6 +715,102 @@ func (wfe *WebFrontEndImpl) UpdateAccount(
...
+		return
+	}
+
+	// Verify inner key
+	if !keyDigestEquals(innerContent.OldKey, *existingAcct.Key) {
+		wfe.sendError(acme.MalformedProblem("Key roll-over innermost JSON contains wrong old key"), response)
```
cpu: another "innermost JSON" to fix.

db/memorystore.go
```diff
@@ -116,6 +119,31 @@ func (m *MemoryStore) AddAccount(acct *core.Account) (int, error) {
...
+		return err, nil
+	}
+
+	newKeyID, err := keyToID(newKey)
+	if err != nil {
+		return err, nil
+	}
+
+	if otherAccount, present := m.accountsByKeyID[newKeyID]; present {
+		return fmt.Errorf("New public key is already in use"), otherAccount
```
cpu: In general it's not good Go practice to return both an error and an instance of the non-error return type. 

It would be better here to return just an `error` from this function. Return only `nil` when everything goes OK and return `existingAccountError{otherAccount.ID}` for a key conflict. You would then define your own `existingAccountError` type that implements the error interface, holds an account ID, and prints a good message about there being an existing account with the given ID.

In the calling code you can specifically check whether `err` is of type `existingAccountError`, and if it is, cast it to that type and extract the existing account ID to return as a `Location` header.

It's a bit more work but I think it will be a cleaner design overall

db/memorystore.go
cpu: Re-reading this I understand now that the `*core.Account` return was being used for key-conflicts. I think you should remove this return entirely and just return an `error` and use the `existingAccountError` I describe in my other comment for that case.

wfe/wfe.go
```diff
@@ -712,6 +715,102 @@ func (wfe *WebFrontEndImpl) UpdateAccount(
...
+		return
+	}
+
+	// Ok, now change account key
+	err, otherAccount := wfe.db.ChangeAccountKey(existingAcct, newPubKey)
+	if err != nil {
```
cpu: This is where you will want to check whether `err` is of the specific type that you defined for a key conflict, and handle that accordingly.

wfe/wfe.go

wfe/wfe.go

wfe/wfe.go

db/memorystore.go
felixfontein: I guess it must be `ExistingAccountError`, otherwise I cannot use it both in `db` and `wfe`.

cpu: @felixfontein Yup! You're correct. It should be an exported type for that reason. Sorry I missed that :-)

felixfontein:
@cpu Thanks for your review! I've now refactored the key already in use error as you suggested. The next step will be shortening the `wfe.KeyRollover` function.

db/memorystore.go
felixfontein: @cpu Great! I'm learning a bit more of Go every time I work on this ;)

felixfontein:
@cpu I think I'm done refactoring. Can you check how close this is to what you want? :)

db/memorystore.go
```diff
@@ -16,6 +16,17 @@ import (
...
 )
 
+// Error when key change fails because new account key has already been in use
+type ExistingAccountError struct {
+	msg             string
+	MatchingAccount *core.Account
+}
+
+func (e *ExistingAccountError) Error() string {
+	return e.msg
```
cpu: I think you can remove the `msg` field from this struct entirely since the only place `ExistingAccountError` is used is for a new key conflict and the message for that case is always the same. If you remove `msg` you can implement `Error()` with: 
`return fmt.Sprintf("New key is already in use by account %s", e.ID)`

db/memorystore.go
```diff
@@ -16,6 +16,17 @@ import (
 	"github.com/letsencrypt/pebble/core"
 )
 
+// Error when key change fails because new account key has already been in use
```
cpu: The Godoc convention is to start a type's comment with the name of the type, like:
> // ExistingAccountError is an error type indicating when an operation fails because the `MatchingAccount` has a key conflict.

db/memorystore.go
```diff
@@ -116,6 +130,31 @@ func (m *MemoryStore) AddAccount(acct *core.Account) (int, error) {
...
+		return err
+	}
+
+	newKeyID, err := keyToID(newKey)
+	if err != nil {
+		return err
+	}
+
+	if otherAccount, present := m.accountsByKeyID[newKeyID]; present {
+		return &ExistingAccountError{"New public key is already in use", otherAccount}
```
cpu: If you refactor the `ExistingAccountError` as suggested above this will end up just being `return ExistingAccountError{otherAccount}` (I believe you don't need to be using `&{}` here to take a reference)

wfe/wfe.go
```diff
@@ -499,64 +502,80 @@ func (wfe *WebFrontEndImpl) verifyPOST(
...
 }
 
-func (wfe *WebFrontEndImpl) verifyJWS(
+// Checks parsed JWS whether it matches the given public key, does some general
+// checks, and extracts the URL header parameter.
+func (wfe *WebFrontEndImpl) verifyJWSLight(
 	pubKey *jose.JSONWebKey,
 	parsedJWS *jose.JSONWebSignature,
-	request *http.Request) ([]byte, *jose.JSONWebKey, *acme.ProblemDetails) {
+	jwsPrefix string) ([]byte, string, *acme.ProblemDetails) {
```
cpu: Are you sure that you need the `jwsPrefix` argument that you're adding to this function? It looks like `verifyJWS` always sets it to `""` and only `wfe.KeyRollover` uses it. I think it would be better to remove this argument and have `wfe.KeyRollover` make a new malformed problem with the correct message prefix when there is a non-nil err response from calling `verifyJWSLight` (or whatever it ends up being called!)

wfe/wfe.go
```diff
@@ -712,6 +731,109 @@ func (wfe *WebFrontEndImpl) UpdateAccount(
...
+func (wfe *WebFrontEndImpl) verifyKeyRollover(
+	innerPayload []byte,
+	existingAcct *core.Account,
+	newKey *jose.JSONWebKey,
+	request *http.Request) *acme.ProblemDetails {
+	var innerContent struct {
+		Account string
+		OldKey  jose.JSONWebKey
+	}
+	err := json.Unmarshal([]byte(innerPayload), &innerContent)
```
cpu: I don't think you need the `[]byte()` cast on `innerPayload` since its already a `[]byte` argument

wfe/wfe.go
```diff
@@ -712,6 +731,109 @@ func (wfe *WebFrontEndImpl) UpdateAccount(
...
+	}
+
+	innerPayload, innerHeaderURL, prob := wfe.verifyJWSLight(newPubKey, parsedInnerJWS, "Inner ")
+	if prob != nil {
+		wfe.sendError(prob, response)
+		return
+	}
+
+	if innerHeaderURL != outerHeaderURL {
+		wfe.sendError(acme.MalformedProblem("JWS header parameter 'url' differs for inner and outer JWS."), response)
```
cpu: I think you missed a `return` after this `wfe.sendError` and control flow will continue when the inner header and outer header don't match

wfe/wfe.go
```diff
@@ -499,64 +502,80 @@ func (wfe *WebFrontEndImpl) verifyPOST(
...
+		return nil, "", nil, prob
 	}
 
 	return wfe.verifyJWS(pubKey, parsedJWS, request)
 }
 
-func (wfe *WebFrontEndImpl) verifyJWS(
+// Checks parsed JWS whether it matches the given public key, does some general
+// checks, and extracts the URL header parameter.
+func (wfe *WebFrontEndImpl) verifyJWSLight(
```
cpu: Sorry, I think I forgot to hit submit to include this comment in my last review:

Can we brainstorm a new name for this function? What about splitting it up into two functions named `verifyJWSSignature` and `extractJWSURL`?  I think that will be more descriptive than `verifyJWSLight`.

db/memorystore.go
felixfontein: Done.

db/memorystore.go
felixfontein: Done.

db/memorystore.go
felixfontein: The `&` is needed, otherwise go complains with:
```
db/memorystore.go:148:30: cannot use ExistingAccountError literal (type ExistingAccountError) as type error in return argument:
	ExistingAccountError does not implement error (Error method has pointer receiver)
```
I had to replace
```
func (e *ExistingAccountError) Error() string { ...
```
by
```
func (e ExistingAccountError) Error() string { ...
```
then it worked :)

wfe/wfe.go
felixfontein: Yes, it also works without it.

wfe/wfe.go
felixfontein: Definitely! Thanks for noticing this!

wfe/wfe.go
felixfontein: Yes, please :) I tentatively named it `verifyJWSLight` while I wasn't sure what exactly will end up in it, and since it is a subset of what `verifyJWS` is doing ;-)

felixfontein: I think I'll name the first one `verifyJWSSignatureAndAlgorithm` (at least for now), as a reminder that the code for checking the algorithm is still missing ;-)

felixfontein: I did the refactoring in 15befc4c28430080ad0ef5dc8d825b54f48cf529. I think it's now much better than before.

felixfontein:
@cpu Since the commit list is getting longer and longer, do you want me to squash the existing ones into one, or a small number of commits?

felixfontein:
(I see you're already squashing when merging, but if that helps you reviewing, I can already squash them before.)

cpu:
> @cpu Since the commit list is getting longer and longer, do you want me to squash the existing ones into one, or a small number of commits?

@felixfontein That's ok :-) I'm happy to squash on merge. I (lightly) prefer having the individual commits from the review process.

PR #153, opened 2018-07-31
Allow to retrieve root certificate by GET /root

### LINKED ISSUE #152: How to get hold of root certificate?
As part of integration tests, I'm verifying the generated certificate chain. However, since #148, I also need the root certificate to validate the chain.

Is there a way to get hold of it? (I think there currently isn't.)

It would be really nice to be able to get the root certificate from somewhere (i.e. Pebble writes it to disk, or provides an endpoint where it can be read with a `GET`).

(ACME itself doesn't provide a way to retrieve the root certificate for an order AFAIK; is that correct? If not, that would also solve this ;-) )

felixfontein:
Provides a `GET` endpoint to retrieve the CA's root certificate. Fixes #152.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,9 +79,6 @@ correctly.
 pebble -config ./test/config/pebble-config.json
 ```
 
-Note that the CA's root certificate is regenerated on every launch. It can be
-retrieved via `https://localhost:14000/root`.
-
 ### Docker Image
 
 With a docker-compose file:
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -79,6 +79,9 @@ correctly.
 pebble -config ./test/config/pebble-config.json
 ```
 
+Note that the CA's root certificate is regenerated on every launch. It can be
+retrieved via `https://localhost:14000/root`.
```
cpu: @felixfontein Could you expand on this section briefly to mention:
1. When you might need the root (such as for an end-to-end test of the generated certificates)
2. That adding the Pebble CA certificate to a trust store should only be done on **testing** systems.

For 2. you might find the "Root Certificate" text on this page helpful to crib from: https://letsencrypt.org/docs/staging-environment/

cpu: It might also be worth emphasizing that the Pebble root CA != the Pebble HTTPS API certificate's root CA.

wfe/wfe.go
```diff
@@ -232,6 +233,17 @@ func (wfe *WebFrontEndImpl) sendError(prob *acme.ProblemDetails, response http.R
...
 
+func (wfe *WebFrontEndImpl) RootCert(
+	ctx context.Context,
+	logEvent *requestEvent,
+	response http.ResponseWriter,
+	request *http.Request) {
+
+	response.Header().Set("Content-Type", "application/pem-certificate-chain; charset=utf-8")
+	response.WriteHeader(http.StatusOK)
+	_, _ = response.Write(wfe.ca.GetRootCert().PEM())
```
cpu: Can you add a check that `wfe.ca.GetRootCert() != nil` and return an error if it is? I don't think there is a way to get Pebble running and have that field be nil but it would be nice to avoid the nil panic on `wfe.ca.GetRootCert().PEM()` just in case.

README.md
felixfontein: I moved the text to a new section at the end of the document. I've also tried to adjust it as you requested.

felixfontein:
Thanks for reviewing this :)

felixfontein:
Thanks for handling this, @cpu!

PR #154, opened 2018-07-31
Using account IDs different from key IDs

felixfontein:
This PR changes the way account IDs are created. They used to be the SHA256 of the account's public key before. Now, they are numeric IDs (incremented by one for every new account).

This is a prerequisite for correctly implementing account key roll-over, for which the account's ID must no longer depend on the account's public key (see #151).

### DIFF
--- a/db/memorystore.go
+++ b/db/memorystore.go
@@ -242,10 +242,8 @@ func (m *MemoryStore) RevokeCertificate(cert *core.Certificate) {
 
 /*
  * keyToID produces a string with the hex representation of the SHA256 digest
- * over a provided public key. We use this for acme.Account ID values
- * because it makes looking up a account by key easy (required by the spec
- * for retreiving existing account), and becauase it makes the reg URLs
- * somewhat human digestable/comparable.
+ * over a provided public key. We use this to associate public keys to
+ * acme.Account objects, and to ensure every account has a unique public key.
  */
 func keyToID(key crypto.PublicKey) (string, error) {
 	switch t := key.(type) {
--- a/db/memorystore.go
+++ b/db/memorystore.go
@@ -1,9 +1,16 @@
 package db
 
 import (
+	"crypto"
+	"crypto/sha256"
+	"crypto/x509"
+	"encoding/hex"
 	"fmt"
 	"reflect"
+	"strconv"
 	"sync"
+
+	"gopkg.in/square/go-jose.v2"
 
 	"github.com/jmhodges/clock"
 	"github.com/letsencrypt/pebble/core"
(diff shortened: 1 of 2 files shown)


db/memorystore.go
```diff
@@ -206,3 +239,29 @@ func (m *MemoryStore) RevokeCertificate(cert *core.Certificate) {
 	defer m.Unlock()
 	delete(m.certificatesByID, cert.ID)
 }
+
+/*
+ * keyToID produces a string with the hex representation of the SHA256 digest
+ * over a provided public key. We use this for acme.Account ID values
```
cpu: I don't think this comment is accurate anymore w.r.t using this for the acme Account ID, right? Now it's only being used to ensure that duplicate keys aren't used across accounts.

wfe/wfe.go
```diff
@@ -1728,14 +1687,12 @@ func (wfe *WebFrontEndImpl) revokeCertByKeyID(
 		return prob
 	}
 
-	keyID, err := keyToID(key)
+	existingAcct, err := wfe.db.GetAccountByKey(key)
 	if err != nil {
-		return acme.MalformedProblem(err.Error())
+		return acme.MalformedProblem("Cannot obtain key ID from public key")
```
cpu: Can you also include the `err` in the malformed problem message here?

wfe/wfe.go
```diff
@@ -1728,14 +1687,12 @@ func (wfe *WebFrontEndImpl) revokeCertByKeyID(
...
 	}
-
-	existingAcct := wfe.db.GetAccountByID(keyID)
 	if existingAcct == nil {
-		return acme.UnauthorizedProblem(fmt.Sprintf("Account with keyID %q does not exist", keyID))
+		return acme.UnauthorizedProblem("Account with public key does not exist")
```
cpu: for this error message, I think it might be clearer as: "No account found corresponding to public key authenticating this request", what do you think?

db/memorystore.go
felixfontein: Yes, it is. I've updated it.

wfe/wfe.go

wfe/wfe.go
felixfontein: Yes, that's much better. I'm using your text now.

felixfontein:
@cpu Thanks for your review!

cpu:
Thanks @felixfontein! On to #151 !

felixfontein:
@cpu Thanks again for reviewing and handling! :)

felixfontein:
Ooops, there was a bug in here: see #155

PR #155, opened 2018-08-03
#154: Forgot to update all account references on update

felixfontein:
@cpu I noticed one bug in #154: on account update, only `m.accountsByID[id]` was updated, but not `m.accountsByKeyID[keyID]` (with corresponding `keyID`). This leads to strange update behavior, with changes suddenly disappearing (depending on how you look at them, i.e. either via account public key or via account URI).

(I noticed that my test for #149 stopped working, resp. started to behave really strange.)

### DIFF
--- a/db/memorystore.go
+++ b/db/memorystore.go
@@ -77,7 +77,12 @@ func (m *MemoryStore) UpdateAccountByID(id string, acct *core.Account) error {
 	if m.accountsByID[id] == nil {
 		return fmt.Errorf("account with ID %q does not exist", id)
 	}
+	keyID, err := keyToID(acct.Key)
+	if err != nil {
+		return err
+	}
 	m.accountsByID[id] = acct
+	m.accountsByKeyID[keyID] = acct
 	return nil
 }
 


cpu:
I think this still leaves a case where an account registered with keyA that rolls-over to keyB would still leave a `m.accountsByKeyID[keyAHash]` entry that would prevent any other accounts using keyA. I think that's fine for now and not a grave concern for Pebble.

felixfontein:
@cpu Thanks for merging!
I don't think there is a problem for key roll-over, since in #151, the old `m.accountsBKeyID[keyAHash]` entry is deleted. (In #151 there's an extra method for updating account keys; `UpdateAccountByID` is not used for changing the key. I guess there should be some comment explicitly stating that.)

felixfontein:
(I added f6db21f17e6389af332cb949ad46e72718ac7a55 in #151 for this.)

cpu:
> I don't think there is a problem for key roll-over, since in #151, the old m.accountsBKeyID[keyAHash] entry is deleted. 

Aha! perfect! That's great. Thanks for confirming.

PR #156, opened 2018-08-10
Return existing account on newAccount with key in use

### LINKED ISSUE #92: NewAccount doesn't return an Account resource if the account already exists
This causes checks to ensure the account is valid (after the account has already been created) to fail if they check the status of the Account (as they likely will want to).

We should return `existingAcct` here to do so: https://github.com/letsencrypt/pebble/blob/master/wfe/wfe.go#L624

From the ACME draft:

```
7.3.1.  Finding an Account URL Given a Key

   If the server already has an account registered with the provided
   account key, then it MUST return a response with a 200 (OK) status
   code and provide the URL of that account in the Location header
   field.  This allows a client that has an account key but not the
   corresponding account URL to recover the account URL.

   If a client wishes to find the URL for an existing account and does
   not want an account to be created if one does not already exist, then
   it SHOULD do so by sending a POST request to the new-account URL with
   a JWS whose payload has an "onlyReturnExisting" field set to "true"
   ({"onlyReturnExisting": true}).  If a client sends such a request and
   an account does not exist, then the server MUST return an error
   response with status code 400 (Bad Request) and type
   "urn:ietf:params:acme:error:accountDoesNotExist".
```

This doesn't actually seem clear to me - should the ACME server **only** respond with a Location header pointing to the Account resource, or should it also include the Account resource itself?
Given `onlyReturnExisting` exists, it 'feels' like newAccount should always return the full account resource if it exists, and this would make developing clients easier. The ACME server could respond with 200 if the resource already exists, or 201 for created (as per current spec)

felixfontein:
Fixes #92.

### DIFF
--- a/wfe/wfe.go
+++ b/wfe/wfe.go
@@ -747,7 +747,7 @@ func (wfe *WebFrontEndImpl) NewAccount(
 		// the account and a 200 OK response
 		acctURL := wfe.relativeEndpoint(request, fmt.Sprintf("%s%s", acctPath, existingAcct.ID))
 		response.Header().Set("Location", acctURL)
-		_ = wfe.writeJsonResponse(response, http.StatusOK, nil)
+		_ = wfe.writeJsonResponse(response, http.StatusOK, existingAcct)
 		return
 	} else if existingAcct == nil && newAcctReq.OnlyReturnExisting {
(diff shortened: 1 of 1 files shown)


felixfontein:
@cpu You're welcome! If implementing features would always be this simple... :)

PR #157, opened 2018-08-13
Improve certificate revocation

felixfontein:
This PR adds code to store all revoked certificates, so that the new [draft-14 error](https://tools.ietf.org/html/draft-ietf-acme-acme-14#section-6.6) `alreadyRevoked` can be returned (see [page 56](https://tools.ietf.org/html/draft-ietf-acme-acme-14#page-56)) instead of a generic malformed problem if an already revoked certificate is to be revoked again.

Updates #136 to newest draft.

### DIFF
--- a/acme/problems.go
+++ b/acme/problems.go
@@ -17,6 +17,7 @@ const (
 	unsupportedContactErr  = errNS + "unsupportedContact"
 	accountDoesNotExistErr = errNS + "accountDoesNotExist"
 	badRevocationReasonErr = errNS + "badRevocationReason"
+	alreadyRevokedErr      = errNS + "alreadyRevoked"
 )
 
 type ProblemDetails struct {
@@ -136,6 +137,14 @@ func UnsupportedMediaTypeProblem(detail string) *ProblemDetails {
 func BadRevocationReasonProblem(detail string) *ProblemDetails {
 	return &ProblemDetails{
 		Type:       badRevocationReasonErr,
+		Detail:     detail,
+		HTTPStatus: http.StatusBadRequest,
+	}
+}
+
+func AlreadyRevokedProblem(detail string) *ProblemDetails {
+	return &ProblemDetails{
+		Type:       alreadyRevokedErr,
 		Detail:     detail,
 		HTTPStatus: http.StatusBadRequest,
 	}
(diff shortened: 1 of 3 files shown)


cpu:
Hi @felixfontein - I haven't forgotten about this PR! I will be reviewing it this week for sure :-)

felixfontein:
@cpu Cool, thanks! In fact, earlier today, I was pondering whether I should ping you and ask about it, but then decided to wait until tomorrow ;)

cpu:
I'll push a fresh docker image with this commit in the near future. I tried now and there's a Gopkg.in <-> Github outage that is causing the `go get` in the `docker build` to fail to retrieve the go-jose dependency.

felixfontein:
@cpu Thanks for merging this! :) I assumed you got sidetracked by the POST-as-GET discussion, so I didn't ping you ;)

felixfontein:
BTW, what do you think is the correct error when one tries to revoke an already revoked certificate with the wrong private key? I noticed this morning that Boulder returns `alreadyRevoked` instead of `unauthorized`; I have no idea what Pebble does, though.

jsha:
Good question! I think that's kinda "dealer's choice," since the spec doesn't mandate either one. Probably `alreadyRevoked` is ideal, since it would be a shame for the user who gets `unauthorized` to spend time hunting down the correct private key, only to find the certificate already revoked. If they get `alreadyRevoked` right away they can know that they don't have to do anything more.

felixfontein:
Sounds good! I was just a bit confused as I revoked a set of certs in a loop (where one cert showed up twice, first with the wrong key and then later with the correct key), and after seeing that some failed with `unauthorized`, I tried again but then everything worked fine. I first thought that there might be a bug in my client, until I noticed that I revoked one cert twice, and so on the second run the wrong key didn't matter anymore...

jsha:
Sounds like a great case study in why `alreadyRevoked` is more informative. :-D

cpu:
> I'll push a fresh docker image with this commit in the near future. 

Published `letsencrypt/pebble:2018-09-12` today with this commit.

felixfontein:
I'm very glad indeed that we now have `alreadyRevoked` :)

PR #158, opened 2018-08-16
TLS-ALPN-01: Adjust IdPeAcmeIdentifier to value in draft-05

felixfontein:
This updates `id-pe-acmeIdentifier-v1` with OID `1.3.6.1.5.5.7.1.30.1` ([draft-01](https://tools.ietf.org/html/draft-ietf-acme-tls-alpn-01#page-4)) to `id-pe-acmeIdentifier` with OID `1.3.6.1.5.5.7.1.31` ([draft-05](https://tools.ietf.org/html/draft-ietf-acme-tls-alpn-05#page-4)).

CC @rolandshoemaker

### DIFF
--- a/cmd/pebble/main.go
+++ b/cmd/pebble/main.go
@@ -59,7 +59,7 @@ func main() {
 	clk := clock.New()
 	db := db.NewMemoryStore(clk)
 	ca := ca.New(logger, db)
-	va := va.New(logger, clk, c.Pebble.HTTPPort, c.Pebble.TLSPort)
+	va := va.New(logger, clk, c.Pebble.HTTPPort, c.Pebble.TLSPort, *strictMode)
 
 	wfe := wfe.New(logger, clk, db, va, ca, *strictMode)
 	muxHandler := wfe.Handler()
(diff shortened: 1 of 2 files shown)


felixfontein:
@cpu I've also been thinking about this (supporting both OIDs), but decided to just use the new OID to nudge people who use Pebble for tests to switch to the new OID ASAP.

But TBH, I complete forgot about `-strict`. Now I'm somewhat undecided :) I guess I'll implement whatever you prefer.

cpu:
> But TBH, I complete forgot about -strict. Now I'm somewhat undecided :) I guess I'll implement whatever you prefer.

Let's use `-strict` for this in case someone has already implemented TLS-ALPN-01 in their integration tests using Pebble. Thanks @felixfontein !

felixfontein:
@cpu Non-strict mode now supports the old identifier, as requested. If strict mode is disabled and extensions with both identifiers appear, the first one will be taken (as in Boulder).

felixfontein:
Cool! This allows me to adjust my TLS-ALPN-01 code to use the draft-05 OID, without the tests screaming ;)
@cpu @jsha thanks a lot for reviewing this!

PR #159, opened 2018-08-20
WFE: Fix typo in finalize error msg

cpu:


wfe/wfe.go
```diff
@@ -818,7 +818,7 @@ func (wfe *WebFrontEndImpl) KeyRollover(
 		wfe.sendError(acme.MalformedProblem("Inner JWS verification error"), response)
 		return
 	}
-
```
felixfontein: Sorry for adding the spaces!

cpu: TBH I didn't even notice that I removed it in this branch! Might have been an accident :-)

felixfontein: I think your editor is doing this, as you checked the same thing in in #160 then ;)

Issue #167, opened 2018-09-25
POST-as-GET fails with malformedRequest


shred:
I am trying to read the current state of an authorization by sending a POST-as-GET request to e.g. `https://localhost:14000/authZ/Xn3qYREpJjrDo7QcGL4hoPK2YbixtF6LUeiZoaIgk3s`. As specified in the draft, the payload is an empty string.

Instead of returning the authorization, Pebble returns a 400 Bad Request with the following problem:
```
{
  "type":"urn:ietf:params:acme:error:malformedRequest",
  "detail":"Malformed authorization update: unexpected end of JSON input",
  "status":400
}
```

eggsampler:
I'm getting the same problem too. After a little investigating I believe it's due to the way jose.v2 decodes the payload, specifically this:

https://github.com/square/go-jose/blob/ef984e69dd356202fd4e4910d4d9c24468bdf0b8/encoding.go#L147-L149

It appears to not differentiate between a missing payload or an empty payload which causes the nil check in pebble to always return false, and thus postAsGet being false:

https://github.com/letsencrypt/pebble/blob/b09dc60e31611024b0e856f7633cd64aa77a707c/wfe/wfe.go#L646

Which then means the requests are being interpreted as a POST not POST-AS-GET, and so it's attempting to parse a non-existent JSON body.

cpu:
This is exactly why I was opposed to using the empty body as the indicator of a POST-as-GET. It's very clear that it causes implementation headaches. 

I will revert https://github.com/letsencrypt/pebble/pull/162/commits/7da91f1b59f16837aaa74b1e1aeb5b9160e38253

Biting my tongue to avoid any more comments about the state of the protocol based on this late breaking IESG feedback..........................................

cpu:
https://github.com/letsencrypt/pebble/pull/168

felixfontein:
As expected, it works again for my implementation! \o/

PR #168, opened 2018-09-26
WFE: Fix POST-as-GET

cpu:
Like I [described on-list](https://mailarchive.ietf.org/arch/msg/acme/HWXOB39xv0TqW6tbILlBnCgMFCQ) this POST-as-GET empty body check is difficult to implement correctly and I messed it up during the implementation PR based on review feedback. This reverts that review feedback and puts the code back into a state that inter-operates with clients.

Resolves https://github.com/letsencrypt/pebble/issues/167

### DIFF
--- a/wfe/wfe.go
+++ b/wfe/wfe.go
@@ -643,7 +643,7 @@ func (wfe *WebFrontEndImpl) verifyJWS(
 	}
 
 	return &authenticatedPOST{
-		postAsGet: len(payload) == 0 && payload != nil,
+		postAsGet: string(payload) == "",
 		body:      payload,
 		url:       headerURL,
 		jwk:       pubKey}, nil


cpu:
Pushed a new docker image for this: `letsencrypt/pebble:2018-09-26`.

jsha:
It looks like my previous concern about distinguishing between `"payload": null` and `"payload": ""` at the ACME layer was overblown. Null payloads are already [forbidden by the JWS spec](https://tools.ietf.org/html/rfc7515#section-7.2.1):

> The "payload" member MUST be present and contain the value BASE64URL(JWS Payload).

And go-jose enforces this, as demonstrated by this small program:

```
package main

import "log"
import "github.com/square/go-jose"

func main() {
	content := ` { "protected": "", "payload": null, "signature": "" } `
	_, err := jose.ParseSigned(content)
	if err != nil {
		log.Fatal(err)
	}
}
```

So simply checking for an empty string at the ACME layer is sufficient to accept only the inputs we want.

PR #171, opened 2018-10-02
Allow POST-as-GET for account retrieval

felixfontein:
If I understand [draft-15](https://tools.ietf.org/html/draft-ietf-acme-acme-15) correctly, the account URL also must allow POST-as-GET (for account information retrieval). The current Pebble version does not support that yet, i.e. you need to pass `{}` as the JWT payload (that isn't anymore mentioned explicitly in the draft, as I assume POST-as-GET is the new preferred method).

This is an attempt to allow POST-as-GET also for account URLs.

### DIFF
--- a/wfe/wfe.go
+++ b/wfe/wfe.go
@@ -746,6 +746,10 @@ func (wfe *WebFrontEndImpl) UpdateAccount(
 	// if this update contains no contacts or deactivated status,
 	// simply return the existing account and return early.
 	if updateAcctReq.Contact == nil && updateAcctReq.Status != acme.StatusDeactivated {
+		if wfe.strict == true && postData.postAsGet == false {
+			wfe.sendError(acme.MalformedProblem("Use POST-as-GET to retrieve account data instead of doing an empty update"), response)
+			return
+		}
 		err := wfe.writeJsonResponse(response, http.StatusOK, existingAcct)
 		if err != nil {
 			wfe.sendError(acme.InternalErrorProblem("Error marshalling account"), response)
--- a/wfe/wfe.go
+++ b/wfe/wfe.go
@@ -722,23 +722,31 @@ func (wfe *WebFrontEndImpl) UpdateAccount(
 		Contact []string `json:"contact"`
 		Status  string   `json:"status,omitempty"`
 	}
-	err := json.Unmarshal(postData.body, &updateAcctReq)
-	if err != nil {
-		wfe.sendError(
-			acme.MalformedProblem("Error unmarshaling account update JSON body"), response)
-		return
-	}
-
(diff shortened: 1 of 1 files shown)


cpu:
> The current Pebble version does not support that yet, i.e. you need to pass {} as the JWT payload (that isn't anymore mentioned explicitly in the draft, as I assume POST-as-GET is the new preferred method).

+1 That's my understanding as well.

Thanks for the PR! I will give this a :mag: and test shortly.

wfe/wfe.go
```diff
@@ -722,23 +722,31 @@ func (wfe *WebFrontEndImpl) UpdateAccount(
...
+		existingAcct, prob = wfe.getAcctByKey(postData.jwk)
+		if prob != nil {
+			wfe.sendError(prob, response)
+			return
+		}
 	}
 
 	// if this update contains no contacts or deactivated status,
 	// simply return the existing account and return early.
 	if updateAcctReq.Contact == nil && updateAcctReq.Status != acme.StatusDeactivated {
```
cpu: I think it would be sensible to return an error if we reach the "retrieve an account because there was no update body" case but `wfe.strict == true && postData.postAsGet == false` because this indicates the client is using the old way of querying account information but wants Pebble to be strict about that sort of thing.

Does that make sense?

felixfontein: I was thinking about that myself, but decided against it, because the draft doesn't say one should not retrieve the account data this way. From how I understand it, this is still a valid way, even though it is not explicitly advertised any more. I'm kind of torn, so I let you decide :)

cpu: IMO: If the draft doesn't describe fetching the account that way then it shouldn't be supported. I think `-strict` should reject deprecated behaviour.

wfe/wfe.go
```diff
@@ -722,23 +722,35 @@ func (wfe *WebFrontEndImpl) UpdateAccount(
...
-		err = wfe.writeJsonResponse(response, http.StatusOK, existingAcct)
+		if wfe.strict == true && postData.postAsGet == false {
```
golangcibot: should omit comparison to bool constant, can be simplified to wfe.strict

wfe/wfe.go
felixfontein: I've added a commit which does that. I've used `acme.MalformedProblem()`; or should I use a different error class instead?

cpu: `acme.MalformedProblem()` sounds correct to me. Thanks @felixfontein !

cpu:
Pushed `letsencrypt/pebble:2018-10-10` to dockerhub with this commit.

felixfontein:
@cpu thanks for reviewing and merging!

PR #172, opened 2018-10-03
malformedRequest -> malformed

felixfontein:
The error is called `urn:ietf:params:acme:error:malformed`, not `urn:ietf:params:acme:error:malformedRequest` according to [the table in Section 6.5 of draft-15 (and earlier ones as well)](https://tools.ietf.org/html/draft-ietf-acme-acme-15#section-6.7).

See also ietf-wg-acme/acme#454.


felixfontein:
Yes, I've noticed that. I was wondering why my fallback to the old account update didn't work, and noticed that Pebble behaved differently from Boulder :)

PR #178, opened 2018-11-30
Lower default nonce reject & VA sleep time.

cpu:
In a POST-as-GET world there are a lot more POST requests which means a lot more chances for bad nonces. The existing default rejection rate of 15% is now too high in practice. 

Similarly, a longer default VA sleep time means clients poll resources waiting for status changes longer, which means more POST-as-GET requests, which means more POSTs which means more chance for nonce rejections again.

The upstream `acme` module gaining support for POST-as-GET but only retrying requests a maximum of **one** time for badNonce errors means without these changes our CI consistently fails after bad luck & many requests ensures two or more concurrent badNonce errors (See: https://github.com/certbot/certbot/issues/6542)


jsha:
Great find!

PR #180, opened 2018-12-03
Do not return deactivated accounts in newAccount

### LINKED ISSUE #179: Behavior for deactivated accounts
While refactoring some code, I noticed the following:
 - when doing a POST-as-GET for an existing, but deactivated account, an error is returned (403, "urn:ietf:params:acme:error:unauthorized");
 - when doing a `newAccount` call for the account private key, the account object is returned with `"status" == "deactivated"`.

According to the [current draft](https://tools.ietf.org/html/draft-ietf-acme-acme-16#section-7.3.6), "Once an account is deactivated, the server MUST NOT accept further requests authorized by that account's key.", I would argue that the behavior for `newAccount` is wrong.

felixfontein:
Returns a specific unauthorized problem when `newAccount` is called with a public key matching a deactivated account. This is a compromise on contradicting conditions in [7.3.6](https://tools.ietf.org/html/draft-ietf-acme-acme-16#section-7.3.6) (*"Once an account is deactivated, the server MUST NOT accept further requests authorized by that account's key."*) and [7.3.1](https://tools.ietf.org/html/draft-ietf-acme-acme-16#section-7.3.1) (*"If the server receives a newAccount request signed with a key for which it already has an account registered with the provided account key, then it MUST return a response with a 200 (OK) status code and provide the URL of that account in the Location header field. The body of this response represents the account object as it existed on the server before this request."*) of [draft-16](https://tools.ietf.org/html/draft-ietf-acme-acme-16).

I decided not to add the `Location` header in this case. This is debatable as well, though, and easy to change.

Fixes #179.

### DIFF
--- a/wfe/wfe.go
+++ b/wfe/wfe.go
@@ -941,11 +941,18 @@ func (wfe *WebFrontEndImpl) NewAccount(
 	// Lookup existing account to exit early if it exists
 	existingAcct, _ := wfe.db.GetAccountByKey(postData.jwk)
 	if existingAcct != nil {
-		// If there is an existing account then return a Location header pointing to
-		// the account and a 200 OK response
-		acctURL := wfe.relativeEndpoint(request, fmt.Sprintf("%s%s", acctPath, existingAcct.ID))
-		response.Header().Set("Location", acctURL)
-		_ = wfe.writeJsonResponse(response, http.StatusOK, existingAcct)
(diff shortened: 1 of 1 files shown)


felixfontein:
@cpu thanks for reviewing and merging :)

cpu:
@felixfontein Of course :-)

I pushed a new Docker tag for this commit: `letsencrypt/pebble:2018-12-06`

PR #181, opened 2018-12-04
pebble-challtestsrv: small binary for mock DNS & ACME challenges.

cpu:
Boulder has a nice handy [`challtestsrv` package and command](https://github.com/letsencrypt/boulder/tree/9e39680e3f78c410e2d780a7badfe200a31698eb/test/challtestsrv) used for integration tests. Its small and useful enough that the library portion has been promoted to a first-class repo: https://github.com/letsencrypt/challtestsrv 

The stand-alone binary with an HTTP management interface can come live in the Pebble repo where more folks can use it without pulling in all of Boulder. I've heard from a few ACME client developers that this would be useful to them. It is possible we could achieve the same thing by leaving the binary in the Boulder repo using the updated code that doesn't import other things from Boulder. Moving it out of the repo will help us commit to working on abstractions that make tests cleaner. This also makes it quick and easy to have a full Pebble environment with mock DNS without needing to install tools from other repos.

The dependency on the letsencrypt/challtestsrv package does require pulling in a dep. on `github.com/miekg/dns` (and vendoring it) but I think its a fair tradeoff.

The provided Dockerfile is now split into two dockerfiles (see `docker/` directory): one for `pebble` and one for `pebble-challtestsrv`. They are both updated to use Go 1.11, to build with the vendored modules instead of fetching them at build time, and to use the latest Alpine base image. 

A new `docker-compose.yml` example is included that starts up a `pebble-challtestsrv` container and a `pebble` container that uses the former as its DNS server. The README is updated to explain the usage briefly.


cpu:
I think I might extract the `challtestsrv` package to its own repo. Pebble can vendor that package to provide `pebble-challtestsrv`. I think this will be cleaner for both Pebble and Boulder.

cpu:
Closing this for now while I pursue a first-class repo for the challtestsrv package (https://github.com/letsencrypt/challtestsrv/pull/1)

I will re-open this when I can vendor that module into Pebble instead of putting the package in this repo.

cpu:
> I will re-open this when I can vendor that module into Pebble instead of putting the package in this repo.

This was done in 2e8f228

Issue #182, opened 2018-12-11
pebble-challtestsrv: Add flag for setting default ipv4/6 addresses on startup


eggsampler:
The new challtestsrv setup is pretty much perfect for consistent and easy testing across both boulder and pebble imo (thanks @cpu 👍). But it would be great if the pebble-challtestsrv binary supported an option to set the default ipv4 address via a command line flag.

At the moment, when using docker-compose you need to post to the /set-default-ipv4 path every startup with the pebble IP address when testing with internal/randomly generated domain names. Alternatively, manually adding the IP address for each a record you're using. I think it would be easier if the binary had a flag and was given the pebble IP by default in the `docker-compose.yml`.

Essentially, the same way pebble is given the pebble-challtestsrv address for dns queries, https://github.com/letsencrypt/pebble/blob/80ae43598b324483279cdc281ea98ca8b6f7ae32/docker-compose.yml#L7
Something similar could be passed to pebble-challtestsrv, ie:
`command: pebble-challtestsrv -defaultipv4 10.30.50.2`

cpu:
@eggsampler Thanks for the feedback. That makes sense to me :+1: I would be happy to take a crack at implementing this.

PR #183, opened 2018-12-11
challtestsrv: add -defaultIPv4 and -defaultIPv6 cli args.

cpu:
This provides an easy way to set the default A/AAAA mock IP addresses without at start-up without needing to use the management interface.

Resolves https://github.com/letsencrypt/pebble/issues/182

### DIFF
--- a/cmd/pebble-challtestsrv/README.md
+++ b/cmd/pebble-challtestsrv/README.md
@@ -16,6 +16,10 @@ redirect behaviour for HTTP-01 challenge validation.
 
 ```
 Usage of pebble-challtestsrv:
+  -defaultIPv4 string
+    Default IPv4 address for mock DNS responses to A queries (default "127.0.0.1")
+  -defaultIPv6 string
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -4,7 +4,7 @@ services:
     build:
       context: .
       dockerfile: docker/pebble/Dockerfile
-    command: pebble -config /test/config/pebble-config.json -strict -dnsserver 10.30.50.3:8053
+    command: pebble -config /test/config/pebble-config.json -strict -dnsserver 10.30.50.3:8053 -defaultIPv6 "" -defaultIPv4 10.30.50.3
```
eggsampler: I believe the -defaultIPv6/4 flags should be on the pebble-challtestsrv binary (not pebble).

cpu: :laughing: You're 100% correct. I slapped this into a commit after I opened the PR and should have paid closer attention. Thanks for catching that! Fixed in 6adc5db

PR #187, opened 2018-12-27
Publish automatically Docker images.

### LINKED ISSUE #146: Version numbers/identifiers for Docker image tags
It would be nice if there would be some kind of version numbers or identifiers for Pebble which can be used to identify specific versions, for example for tags in the official Pebble repository (https://hub.docker.com/r/letsencrypt/pebble/tags/). Having only latest is not so useful if you want to target a specific version :-)

Ideas for version identifiers:
 * commit IDs (the ones from `master`) (I'm using that right now for a [ACME testing container](https://github.com/felixfontein/ansible-acme-test/));
 * dates (version is built from the latest commit to `master` on that day);
 * some integer or "regular" version number (which someone has to increment for every commit/set of commits which should define a new version). This requires manual work (or some clever scripting, with storing the version number somewhere).

I think I'd prefer the first solution, since it is the simplest to implement. Any suggestions/thoughts/ideas?

ldez:
Publishes images for:
- pebble
- pebble-challtestsrv

Publishes image tags:
- `latest`: the latest build
- commit hash: each commit have an image
- tag: each tag have an image

I'm not sure about the commit hash, maybe it's not a good idea.

Also, I can sync latest and the tag if needed.

The PR require that you add env vars in Travis: `DOCKER_USER` and `DOCKER_PASS`.

Fixes #146 

ping @shred :wink:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,9 @@ go:
 env:
   - GO111MODULE=on
 
+services:
+  - docker
+
 before_install:
   - git clone https://github.com/certbot/certbot
   - cd certbot
@@ -33,3 +36,10 @@ before_script:
 script:
   - go vet ./...
   - REQUESTS_CA_BUNDLE=./test/certs/pebble.minica.pem python ./test/chisel2.py example.letsencrypt.org elpmaxe.letsencrypt.org
+
+deploy:
+  - provider: script
+    script: sh .travis/publish.sh
+    skip_cleanup: true
+    on:
+      repo: letsencrypt/pebble
--- a/.travis/publish.sh
+++ b/.travis/publish.sh
@@ -1,0 +1,39 @@
+#!/usr/bin/env bash
+set -e
+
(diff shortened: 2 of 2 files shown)


cpu:
Thanks for the automation PR @ldez :-) I probably won't have a chance to review this until next week but its on my TODO list now :spiral_notepad: :ballot_box_with_check:

ldez:
The new behavior:

- latest: the latest tagged version
- tag: each tag have an image
- no image per commit hash

cpu:
I created a new team under the Let's Encrypt dockerhub organization that only has read/write permissions for the `pebble` image and the `pebble-challtestsrv` image. I created a dedicated `pebblebot` docker hub account and added it to the new team and configured Travis with the `DOCKER_USER` and `DOCKER_PASS` env vars matching the `pebblebot` account. I _think_ that's all we need here! I'll work on creating an initial release tag shortly.

PR #191, opened 2019-01-07
WFE: Return Status 200 for HEAD to new-nonce endpoint.

cpu:
Previously we mistakenly returned status 204 (no content) for all
requests to new-nonce, including HEAD. This status should only be used
for GET requests.

This is the same problem & fix as was reported on the Boulder repo:
https://github.com/letsencrypt/boulder/issues/3989

Similar to https://github.com/letsencrypt/boulder/issues/3989 because this
may be a breaking API change HTTP status 200 (OK) is only returned for
HEAD requests when Pebble is running in `-strict` mode.

### DIFF
--- a/wfe/wfe.go
+++ b/wfe/wfe.go
@@ -360,11 +360,13 @@ func (wfe *WebFrontEndImpl) Nonce(
 	logEvent *requestEvent,
 	response http.ResponseWriter,
 	request *http.Request) {
-	statusCode := http.StatusOK
+	statusCode := http.StatusNoContent
 	// The ACME specification says GET requets should receive http.StatusNoContent
-	// 	// and HEAD requests should receive http.StatusOK.
-	if request.Method == "GET" {
-		statusCode = http.StatusNoContent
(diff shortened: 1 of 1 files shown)


jsha:
Wait, GETs and HEADs are supposed to return different statuses? I missed that in the first go-round. Doesn't that break the HTTP spec?

cpu:
> Wait, GETs and HEADs are supposed to return different statuses? I missed that in the first go-round. Doesn't that break the HTTP spec?

¯\\\_(ツ)_/¯

This traces back to https://github.com/ietf-wg-acme/acme/pull/371 and a suggestion by @martinthomson:
> Note also that 204 for HEAD newNonce isn't necessary. HEAD responses always don't have a body.

I think whether or not it breaks the HTTP spec doesn't really matter - the feedback is too late. The ACME spec is clear on this and unless its fixed in EDIT48 or if there is an errata created it will be what client developers expect from compliant servers.

cpu:
Pushed `letsencrypt/pebble:2019-01-07` with this commit to Dockerhub.

PR #193, opened 2019-01-11
Add goreleaser.

ldez:
It's a proposal, this PR does the same thing as #187 (with more tags) but also publishes binary files.

Goreleaser is simple to use and configure.

- allow to publish automatically docker images
  - for pebble and pebble-challtestsrv
  - tags: `latest`, `vx.y.z`, `vx.y`, `vx`
- allow to publish archive (tar/zip) for the binaries
    - linux
    - windows
    - darwin
    - freebsd
    - openbsd

Goreleaser can do more, but I think it's a little overkill for pebble :wink:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,5 @@ _testmain.go
 *.exe
 *.test
 *.prof
+
+dist/
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,0 +1,111 @@
+project_name: pebble
+
+builds:
+  - binary: pebble
+    main: ./cmd/pebble/
+    env:
+    - GO111MODULE=on
+    - CGO_ENABLED=0
+
+    goos:
+    - windows
+    - darwin
+    - linux
+    - freebsd
+    - openbsd
+    goarch:
+    - amd64
+    - 386
+    - arm
+    - arm64
+    goarm:
+    - 7
+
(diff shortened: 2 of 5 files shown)


cpu:
@ldez quick Q: will this PR also depend on `DOCKER_USER` and `DOCKER_PASS` env vars being present for `goreleaser` to be able to push the images (similar to #187)? I think it will be best if I can create a separate dockerhub user for the CI credentials and grant it permission to just the Pebble repository.

Overall I like the idea of easy binary releases for Pebble and it seems like `goreleaser` is a pretty common choice for this :+1:  I'll check with the other LE folks to see if they think the tool is an acceptable dependency.

ldez:
I had forgotten a line in `travis.yml`:
```
- test -n "$TRAVIS_TAG" && docker login -u="$DOCKER_USERNAME" -p="$DOCKER_PASSWORD"
```

So you can use the env vars names you want to use a dedicated account.

cpu:
Talking this over with @jsha he convinced me that `goreleaser` is probably overkill for Pebble and we should keep it simple to start. 

@ldez with that in mind I'd like to close this PR in favour of #187 - I hope that's OK with you :-)

ldez:
No problem, it was mostly a PR to discuss. :smile_cat:

PR #200, opened 2019-01-31
Leave out port on HTTP request host header if it's the standard one

DanielMorsing:
During testing with nginx-ingress and http->https redirects, an issue
was discovered. Pebble does its http requests with the Host header
set to Host: domain:httpPort. This would confuse nginx-ingress into
redirecting to https://domain:httpPort/.wellknown if mandatory TLS
upgrade was enabled, causing failures once pebble tried to do a TLS
handshake on port 80.

Fix by trimming the port off the Host header if it is port 80.

### DIFF
--- a/va/va.go
+++ b/va/va.go
@@ -488,6 +488,14 @@ func (va VAImpl) fetchHTTP(identifier string, token string) ([]byte, string, *ac
 		return nil, url.String(), acme.MalformedProblem(
 			fmt.Sprintf("Invalid URL %q\n", url.String()))
 	}
+	// trim the :80 port of the url off the host header
+	// if we're using the standard port.
+	// Certain HTTP routers (notably kubernetes nginx-ingress)
+	// gets confused by it
+	if va.httpPort == 80 {
+		httpRequest.Host = identifier
+	}
+
 	httpRequest.Header.Set("User-Agent", userAgent())
 	httpRequest.Header.Set("Accept", "*/*")
 


va/va.go
```diff
@@ -488,6 +488,14 @@ func (va VAImpl) fetchHTTP(identifier string, token string) ([]byte, string, *ac
 		return nil, url.String(), acme.MalformedProblem(
 			fmt.Sprintf("Invalid URL %q\n", url.String()))
 	}
+	// trim the :80 port of the url off the host header
```
jsha: Let's move this above where we create the URL; we should create a variable `host` that contains either the host:port or host alone, depending on the port. Then we can stick that directly into the `url.URL` when we create it.

jsha: This comment was from an earlier draft of feedback, sorry I forgot to delete it. Let's resolve the spec issue before we move forward with this edit.

DanielMorsing:
The component that "breaks" is a nginx-ingress, the integration between nginx and kubernetes. When set up with mandatory TLS redirects, it will take a request in the form of
```
GET /.wellknown/ HTTP/1.1
Host: foo.example.org:80
```
and redirect it to https://foo.example.org:80/.wellknown/, meaning it's directing pebble to do a TLS connection on port 80, which will fail the handshake.

After reading what you said though, I'm thinking this fix is a bit of a bandaid, so might abandon it and try to work with the nginx-ingress folks. From reading the RFCs, I still can't tell if `Host: foo.example.org:80` is a distinct identity from `Host: foo.example.org`, and whether it would be better to leave the port number off the Host header regardless of which port was used to establish a connection

jsha:
Great, thanks! It sounds like this might actually show a broader bug in nginx-ingress: If you have something on `http://example.com:5555`, mandatory TLS redirects might redirect to `https://example.com:5555`, which is guaranteed not to work. So maybe mandatory TLS redirects should only operate on port 80, and always send you to `https://example.com/` (i.e. without the port).

jsha:
I'm going to close this for now; thanks for working on it, and keep in touch. :-)

Issue #192, opened 2019-01-07
Make current `-strict` behaviour default


cpu:
Pebble is carrying some breaking changes that more closely reflect the ACME specification but could break clients. To make Pebble less annoying to use in CI we gate these changes behind the `-strict` feature flag so that clients can make a choice about whether they'd like a stable(ish) ACME server for CI, or to find compatibility problems in their client before they become an issue with Let's Encrypt's production servers.

At some point in the near future we should make a flag day to convert the `-strict` behaviours to default. Most of them have been present for some time and I think its reasonable to switch them to default soon.

At present these are the `-strict` changes that would become default:

- [ ] GET requests to Order, Authz, Challenge and Certificate resources will fail. POST-as-GET will be required.
- [ ] HEAD requests to new nonce will return status 200 not status 204.
- [ ] POST requests will be required to have the correct Content-Type header: `application/jose+json`
- [ ] Using a POST with an empty JWS body to fetch Account resources will fail. A POST-AS-GET request should be used instead.
- [ ] POSTs to Challenge resources that contain a key authorization in the JWS body will fail. The JWS body should be the trivial JSON payload `{}`.
- [ ] TLS-ALPN-01 challenge response certificates using the obsolete OID (`1.3.6.1.5.5.7.1.30.1`) will fail validation. The current OID should (`1.3.6.1.5.5.7.1.31`) be used.

cpu:
I think if we switch to semver (https://github.com/letsencrypt/pebble/issues/146) this would be a natural point at which to bump major version numbers.

jsha:
Let's go ahead and choose this flag day as February 15. We'll merge a PR to master making these behaviors the default, and bump to v2.0.0. Please post here if you expect this to cause big problems for you.

Klaus-Tockloth:
The current Certbot release (0.30.2) isn't working (for me) in strict mode (option -strict):

```
{
   "type": "urn:ietf:params:acme:error:malformed",
   "detail": "Challenge response body contained legacy KeyAuthorization field, POST body should be `{}`",
   "status": 400
}
```

jsha:
Thanks for checking! I've [told the Certbot team](https://github.com/certbot/certbot/issues/6755)

cpu:
> Let's go ahead and choose this flag day as February 15. We'll merge a PR to master making these behaviors the default, and bump to v2.0.0.

[...]

PR #203, opened 2019-02-14
Promote -strict behaviors to the default

### LINKED ISSUE #206: Validate POST before looking up authz
Per comment in https://github.com/letsencrypt/pebble/pull/203/files/c23f6987fad8dc8448a70d266bc89ce7d2ebef7b#r257281428

Also fix this for challenges per https://github.com/letsencrypt/pebble/pull/203/files/c23f6987fad8dc8448a70d266bc89ce7d2ebef7b#r257280132

jsha:
This change keeps the -strict flag and related plumbing, in case we need it again in the future.

Since this rejects challenge POSTs with `keyAuthorization`, and Certbot master still sends that,
I've updated the Travis config to check out a branch of Certbot that is tweaked to not send it.

Resolves https://github.com/letsencrypt/pebble/issues/192
Resolves #206

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ services:
   - docker
 
 before_install:
-  - git clone https://github.com/certbot/certbot
+  - git checkout -b no-keyauthorization https://github.com/certbot/certbot
   - cd certbot
   - ./certbot-auto --os-packages-only -n
   - ./tools/venv.py
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ services:
   - docker
 
 before_install:
+  # TODO(#204): Change back to cloning master once Certbot omits # keyAuthorization.
   - git clone --depth=1 -b no-keyauthorization https://github.com/certbot/certbot
   - cd certbot
   - ./certbot-auto --os-packages-only -n
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ services:
   - docker
 
 before_install:
-  # TODO(#204): Change back to cloning master once Certbot omits # keyAuthorization.
+  # TODO(#204): Change back to cloning master once Certbot omits keyAuthorization.
   - git clone --depth=1 -b no-keyauthorization https://github.com/certbot/certbot
   - cd certbot
   - ./certbot-auto --os-packages-only -n
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ services:
   - docker
 
 before_install:
-  - git checkout -b no-keyauthorization https://github.com/certbot/certbot
+  - git clone --depth=1 -b no-keyauthorization https://github.com/certbot/certbot
   - cd certbot
   - ./certbot-auto --os-packages-only -n
   - ./tools/venv.py
--- a/va/va.go
+++ b/va/va.go
@@ -396,11 +396,6 @@ func (va VAImpl) validateTLSALPN01(task *vaTask) *core.ValidationRecord {
 	for _, ext := range leafCert.Extensions {
 		if ext.Critical {
 			hasAcmeIdentifier := IdPeAcmeIdentifier.Equal(ext.Id)
-			// For backwards compatibility, check old identifier
-			// as well if strict mode is not enabled
-			if !va.strict && IdPeAcmeIdentifierV1Obsolete.Equal(ext.Id) {
-				hasAcmeIdentifier = true
-			}
 			if hasAcmeIdentifier {
 				var extValue []byte
 				if _, err := asn1.Unmarshal(ext.Value, &extValue); err != nil {
--- a/va/va.go
+++ b/va/va.go
@@ -73,11 +73,6 @@ const (
 // (https://tools.ietf.org/html/draft-ietf-acme-tls-alpn-04#page-4)
 var IdPeAcmeIdentifier = asn1.ObjectIdentifier{1, 3, 6, 1, 5, 5, 7, 1, 31}
 
-// This is the identifier defined in draft-01. This is only supported for backwards
-// compatibility.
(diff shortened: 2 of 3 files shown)


.travis.yml
```diff
@@ -16,7 +16,7 @@ services:
   - docker
 
 before_install:
-  - git clone https://github.com/certbot/certbot
+  - git clone --depth=1 -b no-keyauthorization https://github.com/certbot/certbot
```
cpu: Can you file a Pebble issue to link with a TODO here to revert to Certbot master when they've fixed this upstream?

va/va.go
```diff
@@ -396,11 +396,6 @@ func (va VAImpl) validateTLSALPN01(task *vaTask) *core.ValidationRecord {
 	for _, ext := range leafCert.Extensions {
 		if ext.Critical {
 			hasAcmeIdentifier := IdPeAcmeIdentifier.Equal(ext.Id)
-			// For backwards compatibility, check old identifier
-			// as well if strict mode is not enabled
-			if !va.strict && IdPeAcmeIdentifierV1Obsolete.Equal(ext.Id) {
```
cpu: You can also delete `IdPeAcmeIdentifierV1Obsolete` - nothing else uses it AFAICT

wfe/wfe.go
```diff
@@ -1365,7 +1354,7 @@ func (wfe *WebFrontEndImpl) Order(
 	request *http.Request) {
 
 	var account *core.Account
-	if request.Method == "GET" && wfe.strict {
+	if request.Method == "GET" {
```
cpu: I don't think this check is meaningful anymore because the handle func is only registered for "POST" now:
> wfe.HandleFunc(m, orderPath, wfe.Order, "POST")

wfe/wfe.go
```diff
@@ -1919,21 +1890,15 @@ func (wfe *WebFrontEndImpl) Certificate(
...
 	request *http.Request) {
-	if request.Method == "GET" && wfe.strict {
-		response.Header().Set("Allow", "POST")
-		wfe.sendError(acme.MethodNotAllowed(), response)
+	postData, prob := wfe.verifyPOST(ctx, logEvent, request, wfe.lookupJWK)
+	if prob != nil {
+		wfe.sendError(prob, response)
+		return
+	}
+	_, prob = wfe.validPOSTAsGET(postData)
```
cpu: I think we should be capturing the `acct` return from `validPOSTAsGet` and checking that the account specified owns the order corresponding to the certificate serial being fetched. It looks like this was missed in the original code and might need a new way to map from the certificate serial to the order. Maybe better as a follow-up issue? If you agree will you file the ticket?

wfe/wfe.go
```diff
@@ -1679,15 +1660,6 @@ func (wfe *WebFrontEndImpl) Challenge(
...
-	// Unauthenticated GETs to challenges are only allowed when not running in
-	// strict mode.
-	if request.Method == "GET" && wfe.strict {
-		response.Header().Set("Allow", "POST")
-		wfe.sendError(acme.MethodNotAllowed(), response)
-		return
-	}
-
 	chalID := strings.TrimPrefix(request.URL.Path, challengePath)
 	chal := wfe.db.GetChallengeByID(chalID)
```
cpu: Maybe unrelated to this PR but it looks like we're fetching the challenge (& returning a 404 if it isn't found) before validating the POST. I think it would be better to do the opposite.

wfe/wfe.go
```diff
@@ -1679,15 +1660,6 @@ func (wfe *WebFrontEndImpl) Challenge(
 	logEvent *requestEvent,
 	response http.ResponseWriter,
 	request *http.Request) {
-
-	// Unauthenticated GETs to challenges are only allowed when not running in
-	// strict mode.
```
cpu: I think the diff to this function missed some complexity that could be reduced further down at L1700 (Github won't let me leave a comment on the line):
```
	// If there was an account authenticating this GET request then make sure it
	// owns the challenge.
	if account != nil && chal.Authz.Order.AccountID != account.ID {
```
There shouldn't be any case where `account == nil` with mandatory POST-as-GET right?

wfe/wfe.go
```diff
@@ -1586,14 +1575,6 @@ func (wfe *WebFrontEndImpl) Authz(
...
-
-	// Only allow GET when not being strict
-	if request.Method == "GET" && wfe.strict {
-		response.Header().Set("Allow", "POST")
-		wfe.sendError(acme.MethodNotAllowed(), response)
-		return
-	}
-
 	authzID := strings.TrimPrefix(request.URL.Path, authzPath)
 	authz := wfe.db.GetAuthorizationByID(authzID)
```
cpu: Maybe unrelated to this PR but it looks like we're fetching the authz (& returning a 404 if it isn't found) before validating the POST. I think it would be better to do the opposite.

wfe/wfe.go
cpu: I think this also means the `else if` that follows can be simplified.

.travis.yml
```diff
@@ -16,7 +16,8 @@ services:
   - docker
 
 before_install:
-  - git clone https://github.com/certbot/certbot
+  # TODO(#204): Change back to cloning master once Certbot omits # keyAuthorization.
```
cpu: tiny nit. 
```suggestion
  # TODO(#204): Change back to cloning master once Certbot omits keyAuthorization.
```

wfe/wfe.go
```diff
@@ -1365,23 +1354,17 @@ func (wfe *WebFrontEndImpl) Order(
 	request *http.Request) {
 
 	var account *core.Account
```
cpu: I think you can delete this explicit `var` & the `account = acct` assignment on L1367 by changing L1362 to `account, prob := wfe.validPOSTAsGET(postData)`.

PR #212, opened 2019-02-22
Connect AppVeyor to Pebble project

adferrand:
This PR is the prerequisite to #208, it is about activating a Windows pipeline for Pebble. This PR contains only a dummy pipeline that is here to confirm that everything is wired. The actual logic for this pipeline will be in the next PR.

AppVeyor is a CI provider that uses Windows VMs to execute specific pipelines for this platform. We use it in the Certbot project. Everything is configured through an `appveyor.yml` file, similar to `.travis.yml`.

I put here a link to the equivalent PR that I did for Certbot when we connected AppVeyor for it: certbot/certbot#6361. It contains the detailed instructions for the steps I am about to enumerate here:

Basically, activating AppVeyor consists in:
* creating a `letsencrypt` account on https://www.appveyor.com/,
* associating the `letsencrypt` GitHub account using OAuth,
* create an AppVeyor project for the `letsencrypt/pebble` GitHub project
* merging this PR

Once done, I will check that the dummy pipeline is correctly triggered for the `master` branch. Then I will update my PR against `master` to trigger the first full pipeline logic for a PR push.

PR #213, opened 2019-02-22
Configure AppVeyor to deliver native Windows containers of Pebble and Challtestsrv

### LINKED ISSUE #208: Design of Windows CI/CD pipeline for Pebble
The general idea of this design proposition, is to setup an AppVeyor CI/CD pipeline that ultimately publish on Docker Hub a Windows Container for Pebble, and another one for Pebble-Challtestsrv.

Indeed my recent tests to build Certbot on Windows showed that Pebble can run on Windows out-of-the-box, and so is a suitable integration tests ACME CA server for this platform. The preferred way to distribute the software would be Windows Containers available on Docker Hub side by side to their Linux counterparts.

I see two general approaches for the AppVeyor CI:
1) Create an equivalent pipeline to Travis CI for Linux
2) Focus only on the Docker Hub publication

For 1), it means to translate each step of the Travis CI/CD pipeline to the AppVeyor CI/CD one:
* setup hosts and environment variables
* clone Certbot, setup its venv, enable it
* go install pebble and start the generated binary
* run chisel2.py
* build from source the docker, and publish it

The way AppVeyor is configured is really close to the Travis approach, consuming a `appveyor.yml` YAML file with typical steps of run, build, test and deploy. So the technology allows to follow the same general approach. In the details, apart from Go which is built to be cross-platform out-of-the-box:
* Certbot is already working on Windows, on a satisfactory test-grade level (almost all unit tests are passing, in `acme` module is already fully compatible)
* from what I see, `chisel2.py` has no platform specific code
* Docker is available and running on standard AppVeyor Windows platform

On the cons, the full build logic needs to be maintained on both AppVeyor and Travis. On the pros, Pebble is fully tested on Windows before its publication to Docker Hub.

For 2), the idea would be to first poll the relevant build status from Travis, using its API (see https://docs.travis-ci.com/user/developer/#api-v3). When the final status (succeed/failed) is obtained, AppVeyor proceed this way:
* in case of Travis build failure, the AppVeyor build is failed
* in case of Travis build succeed, AppVeyor build and publish the Windows Docker, in a really similar way to the last step of the solution 1).

On the cons, Pebble is not tested on Windows, only relying on the results obtained on Linux. On the pros, AppVeyor config contains less logic, and does not need to be maintained in parallel to the Travis on. Also, the tests are relying on Certbot for Linux, which is at a production-grade level.

I prefer personnally 1) for the Windows testability, but I wanted to propose an alternative with less AppVeyor knowledge commitment.

As last words, you can see on by fork a running example of AppVeyor building and pushing Pebble + Pebble-Challtestsrv Windows Docker upon GIT push, and operable relevant Dockerfiles for Windows.

@jsha, @cpu, I am open and available for discussions!

adferrand:
Fixes #208. As a prerequisite, #212 must be merged before.

This PR configure an AppVeyor pipeline to build and publish native Windows containers of Pebble and Challtestsrv on Docker Hub. This pipeline execute the same logic that his Linux counterpart on Travis CI: go build, go test using chisel2 and `acme` module from Certbot, then docker build + docker push.

The CI part (build + test) is executed for every commit on every branch (including pull requests). The CD part (docker deploy) is executed only for tags pushed on Pebble repository. When triggered, the CD will publish two native Windows dockers on the `letsencrypt/pebble` and `letsencrypt/pebble-challtestsrv` Docker Hub projects:
* the pinned release tag `X.X.X-nanoserver-sac2016`, that corresponds to the tag released on the GIT repository (Linux counterpart is `X.X.X`)
* the rolling release tag `nanoserver-sac2016` that is the latest pinned release (Linux counterpart is `latest`)

Aside AppVeyor configuration, I also modified chisel2 to be runnable on both Python 2.x and Python 3.x, because Certbot needs to run on Python 3.x on Windows. I also made some lights corrections on the Travis CI pipeline to make it live nicely with AppVeyor.

Once this PR have been merged, one thing will need to be done from the Pebble maintainers that I cannot do myself. Indeed, two secured environment variables are required by this pipeline: `DOCKER_USER` and `DOCKER_PASS` that are the Docker Hub credentials used to push the image triggered by a GIT tag push, to trigger a release.

These credentials need to be set with the login/password of a user authorized to push to `letsencrypt/pebble` and `letsencrypt/pebble-challtestsrv` Docker Hub projects. To do so, go to the AppVeyor Pebble project > Settings > Environment, and add the two environment variables (name + value) in the "Environment variables" section. **Be sure to check the lock "Toggle variable encryption" for each environment variable.**

This can be done later after this PR have been merged, because as I said, the credentials are required only when a tag, and so a new release of Pebble, is pushed.

Of course I am available for any help required to execute the steps here.


appveyor.yml
```diff
@@ -0,0 +1,36 @@
+image: Visual Studio 2017
+
+hosts:
+  example.letsencrypt.org: 127.0.0.1
+  elpmaxe.letsencrypt.org: 127.0.0.1
+
+environment:
+  DOCKER_USER:
+    secure: E0lafFBkoStDl0EOWJ7RJg==
```
ldez: It's better to define these values (`DOCKER_USER`, `DOCKER_PASS`) inside the Appveyor interface.

appveyor-secure-env

.travis.yml
```diff
@@ -1,6 +1,7 @@
 language: go
 
 sudo: true
+dist: xenial
```
ldez: could you explain this change?

.travis.yml
```diff
@@ -30,16 +31,14 @@ install:
 
 before_script:
   - pebble &
-  # Wait for pebble to come up
-  - until </dev/tcp/localhost/14000 ; do sleep 0.1 ; done
```
ldez: could you explain this change?

docker/pebble-challtestsrv/Dockerfile.windows
```diff
@@ -0,0 +1,17 @@
+FROM golang:1.11-nanoserver-sac2016 as builder
```
ldez: `windows.Dockerfile` seems a better (for highlighting code at least) and more standard name than `Dockerfile.windows`.

Same thing for `Dockerfile.linux` -> `linux.Dockerfile`

docker/pebble/Dockerfile.windows
```diff
@@ -0,0 +1,18 @@
+FROM golang:1.11-nanoserver-sac2016 as builder
```
ldez: `windows.Dockerfile` seems a better (for highlighting code at least) and more standard name than `Dockerfile.windows`.

Same thing for `Dockerfile.linux` -> `linux.Dockerfile`

.travis.yml
adferrand: By default Travis uses Ubuntu Trusty, that have a very old version of Python 2.7, the 2.7.6. This version does not offer an easy way to skip the ssl check required to use the polling on Pebble directory url I add in this PR. 

Xenial ships with Python 2.7.11, that have this support.

.travis.yml
adferrand: If Linux has a decent way to poll on a network socket, this is not the case for Windows. In order to have a consistent way to have a uptime check on Pebble on both platform, I moved this process directly on Chisel, making this instruction useless.

As a side note, this migration allows to have a self contained Chisel script that could work on a remote Pebble instance.

appveyor.yml
adferrand: You are totally right ! It makes the configuration more portable, for the same level of security. I updated the code, and the PR description.

docker/pebble-challtestsrv/Dockerfile.windows
adferrand: Indeed. Done.

docker/pebble/Dockerfile.windows
adferrand: Indeed. Done.

cpu:
@adferrand I created a conflict with `appveyor.yml` when I merged #212 that needs to be resolved on this brnach.

One other question from reading the PR desc (I haven't started on reviewing the diff yet):
> * the pinned release tag `X.X.X-nanoserver-sec2016`, that corresponds to the tag released on the GIT repository (Linux counterpart is `X.X.X`)
> * the rolling release tag `nanoserver-sec2016` that is the latest pinned release (Linux counterpart is `latest`)

What is the significance of the "nanoserver-sec2016" suffix?

adferrand:
First, I made a typo, this is `nanoserver-sac2016`, not `nanoserver-sec2016`. Fixed it.

So, `nanoserver-sac2016` means: a Windows container running on the Nano Server installation option of Windows Server 2016 update through the Semi-Annual Channel (SAC). Pretty complex name instead of simply `windows-2016` similarly to `debian-stretch` and other `centos-7`, you would say.

Because, on the contrary to Linux containers, where basically, any decent kernel (3.10+, released in 2013) will be able to run any Docker, this is not the case on Windows. For instance, if you use a Windows Server Core 1809, you will be able to run your Windows container only on the latest Windows 10 version, released 4 months ago (the ... `1809`)

So I choose a decent tag (more on this after), and so take exactly the same name than the `golang` image flavor, on which Pebble is compiled. At this time, the `1.11-nanoserver-sac2016`.

Now about the choice, by itself:
* why `nanoserver`: because it is the lightest edition of Windows Server, specifically designed for Docker containers.
* why `2016`: because Windows Server 2016 make this container runable on any Windows 10/Windows Server 2016+ machine
* why `sac`: because the Semi-Annual Channel (SAC) as opposed to the Long Term Support Channel (LTSC) gives decent updates to take advantages of the continual improvement of Docker for Windows. I think it is a good fit for an integration tool like Pebble. Also, LTSC is not available for Nano Server edition in `golang` container.

adferrand:
And merging conflict solved.

cpu:
> And merging conflict solved.

Thanks!

> So, `nanoserver-sac2016` means: a Windows container running on the Nano Server installation option of Windows Server 2016 updated through the Semi-Annual Channel (SAC).

Ahhh! Thank you for the explanation. I was totally unfamiliar with this terminology before now. I agree with your rationale for selecting this particular matrix of base OS and update channel :+1:

appveyor.yml
```diff
@@ -1,4 +1,32 @@
...
+  - venv3\Scripts\activate.bat
+  - cd ..
+
+build_script:
+  - go install -v -mod=vendor ./...
+  - ps: $PebbleProcess = Start-Process C:\gopath\bin\pebble.exe -PassThru
 
 test_script:
-  - ps: Write-Host "Hello, world!"
+  - go vet ./...
```
cpu:
```suggestion
  - go vet -mod=vendor ./...
```

docker/pebble-challtestsrv/windows.Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+RUN go install -v ./cmd/pebble-challtestsrv/...
+
+## main
+FROM mcr.microsoft.com/windows/nanoserver:sac2016
+
+COPY --from=builder /gopath/bin/pebble-challtestsrv.exe /gopath/bin/pebble-challtestsrv.exe
+
+RUN powershell.exe -Command $path = $env:path + ';c:\gopath\bin'; Set-ItemProperty -Path 'HKLM:\SYSTEM\CurrentControlSet\Control\Session Manager\Environment\' -Name Path -Value $path
+
+CMD [ "/pebble-challtestsrv" ]
```
cpu: It's possible I'm not understanding the `RUN` invocation and Windows filesystem mechanics but does this `"/pebble-challtestsrv"` need to be `"/pebble-challtestsrv.exe"`?

test/chisel2.py
```diff
@@ -44,6 +51,19 @@
 SET_TXT = "http://localhost:8055/set-txt"
 CLEAR_TXT = "http://localhost:8055/clear-txt"
 
+def wait_for_acme_server():
+    """Wait for directory URL set in the DIRECTORY env variable to respond"""
+    ctx = ssl.create_default_context()
+    ctx.check_hostname = False
+    ctx.verify_mode = ssl.CERT_NONE
```
cpu: Can you help me understand why `ctx.verify_mode = ssl.CERT_NONE` is needed here?

Is it possible you could poll the directory using `requests` instead of `urlopen` so that the `REQUESTS_CA_BUNDLE` env var takes effect?

.ci/publish_linux.sh
```diff
@@ -1,13 +1,6 @@
 #!/usr/bin/env bash
 set -e
 
-if [ "${TRAVIS_PULL_REQUEST}" = "false" ]; then
```
cpu: Can you explain why its OK to delete this check? It seems like we would still want the linux publish script to not publish based on builds for pull requests.

cpu:
Apologies, I also added some CI work that will need to be merged into this branch as part of https://github.com/letsencrypt/pebble/commit/08b2d7a6116126bda2fa1fab09dfb24800a79959 - I think the changes should be platform independent.

.ci/publish_linux.sh
adferrand: Because `deploy` step in `travis.yml` is set to `on.tags: true`. This means that this step will be executed only on a tag push, not during a PR, or even a normal push on `master`. So the condition in the bash script is redundant.

You can see effective execution of the pipeline to a push to `master` on my fork here: https://travis-ci.org/letsencrypt/pebble/builds/498856123

You will see that the deploy step is not executed, with an explicit message about that.

appveyor.yml
adferrand: For my curiosity, what means this option, as its effect to the build is not obvious to me?

docker/pebble-challtestsrv/windows.Dockerfile
adferrand: In Windows, file extension is more closer to a metadata about the file type, than a part of the file name. In particular for `.exe` files, it instructs Windows that the file is an executable: then it can be called with or without the `.exe` extension with the exact same behavior.

test/chisel2.py
adferrand: Well, you are right, it is the simplest option. I did not used it in the first place, because there is no formal mechanism (like a setuptools layout) to say that `requests` module is required for chisel script.


So it happens to be here because Certbot and its dependencies require it. However, the propability that `requests` would stop to be a transitive dependency to Certbot is equivalent to Certbot been rewritten in Java, I think we can rely on it.

.ci/publish_linux.sh
cpu: That makes sense! Thanks :+1:

docker/pebble-challtestsrv/windows.Dockerfile
cpu: That makes sense!

For my own curiosity could you help me understand the `RUN powershell.exe` line and how `/pebble-challtestsrv` ends up being resolved to an absolute path? It looks like the powershell fragment is required to add the `$GOPATH/bin` to the system path, but what does the registry key being set accomplish? 

Thanks for schooling me on all this Windows stuff :-)

adferrand: Again, this is because Windows is quite complicated about this. Here, this line with powershell does two things:
* an equivalent of classical Linux operation `PATH=/some/path/:$PATH` to make immediately available a new path resolution on the current shell
* an insertion in the registry to make this updated PATH modification available to any new process

The second step is necessary because in Windows, calling a shell from another shell does not use a fork approach, that would end to a complete copy of the environment (including PATH) from the parent process to the child process. It uses a thread approach: in this case, the child shell is a new thread, with an independent fresh new environment, so without the updated PATH. The fresh environment is constructed from the registry, so it gets the updated PATH from it thanks to the second operation.

test/chisel2.py
cpu: Gotcha, that rationale makes sense but I think using `requests` from `chisel2.py` as an implicit dependency is fine for now.

I'd like to find time at some point in the future to clean up the duplication between Boulder's `chisel.py`,`chisel2.py` and Pebble's copy of `chisel2.py` but that's a separate concern. I think the code complexity is high enough that it deserves a `requirements.txt` file and being treated as a first class Python project on its own.

> However, the propability that `requests` would stop to be a transitive dependency to Certbot is equivalent to Certbot been rewritten in Java, I think we can rely on it.

:laughing: I for one am excited about the new `AbstractCertbotACMEProviderInterfaceFactory`

adferrand:
Migration of the lint/test/coverage is done for AppVeyor. The Windows pipeline has now the same exact steps and actions than its Linux equivalent.

To make the coverage work, a secured variable named `COVERALLS_TOKEN` must be added in the AppVeyor account the same way than `DOCKER_USER` and `DOCKER_PASS`. Its value must be the repo token that can be found on https://coveralls.io/github/letsencrypt/pebble/settings.

adferrand:
With certbot/certbot#6758 that has been merged, Certbot master branch can be used again with Pebble. I updated the AppVeyor configuration here. PR #217 takes care of Travis configuration.

cpu:
> @cpu do you still have the credentials for pebblebot handy, to add them to AppVeyor? Or should we generate a separate pebblebotwindows users?

I have the `pebblebot` creds encrypted in my password manager. I think its reasonable to use the same account for AppVeyor. I'll try adding them.

cpu:
> I think its reasonable to use the same account for AppVeyor. I'll try adding them

@jsha Can you verify my Github/AppVeyor user has access to the LE project you set up? I only see Certbot stuff (`https://ci.appveyor.com/account/certbot/projects`) and my personal repos (`https://ci.appveyor.com/account/cpu/projects`). There doesn't appear to be any connection to the Let's Encrypt github org or the Pebble repo.

jsha:
I fixed this - I had associated the Pebble project in AppVeyor with the Certbot user. I removed it and re-added it, and I added permissions on the Pebble project for the "Boulder Developers" group. It should be ready for you to put in the credentials for pebblebot in the env settings.

cpu:
> It should be ready for you to put in the credentials for pebblebot in the env settings.

I'm afraid I still only see Certbot projects and my own projects. There's nothing for the LE org. Do you want to try and fix that or should I email you GPG encrypted pebblebot creds?

cpu:
I kicked the Travis task that failed because it looked like a transient error fetching tools:
```
go: golang.org/x/crypto@v0.0.0-20181203042331-505ab145d0a9: git fetch -f origin refs/heads/*:refs/heads/* refs/tags/*:refs/tags/* in /home/travis/gopath/pkg/mod/cache/vcs/de5fd3af413a4f3f969455ae522b4002fcb7bb4c158f339396dfc77710c9007d: exit status 128:
	fatal: remote error: Access denied to IP 35.192.187.174
```

PR #220, opened 2019-03-05
Only run goveralls if $GOVERALLS_TOKEN is present

jsha:
This makes running in AppVeyor simpler.


jsha:
Ah, good point. Looks like we could use `APPVEYOR`: https://www.appveyor.com/docs/environment-variables/. I'll do that.

adferrand:
Superseeded by my latest modification in #213, that removes goverall execution in AppVeyor.

PR #221, opened 2019-03-07
Support IP address identifiers

orangepizza:
I made I_think_this_works version, can't find client that support pebble and easy enough to mass with
it's compiled and tested for signs cert wih normal domains, and I checked WFE parts with modified client, but couldn't make the clients work with actual CSR with IP Address. , but when csr passed with shortcut (with some hack), it will sign it properlly( I think)

TLS-ALPN-01 doesn't implanted yet


wfe/wfe.go
```diff
@@ -1489,22 +1498,50 @@ func (wfe *WebFrontEndImpl) FinalizeOrder(
...
-		if name != csrNames[i] {
+	for i, name := range orderDNSs {
+		if name != csrDNSs[i] {
 			wfe.sendError(acme.UnauthorizedProblem(
 				fmt.Sprintf("CSR is missing Order domain %q", name)), response)
 			return
 		}
 	}
+	for i, IP := range orderIPs {
+		if csrIPs[i].Equal(net.ParseIP(IP)) == false {
```
golangcibot: S1002: should omit comparison to bool constant, can be simplified to `!csrIPs[i].Equal(net.ParseIP(IP))` (from `gosimple`)

orangepizza:
I think there should be better ways to streamline commit history...

cpu:
Thanks for the PR @orangepizza ! I might not get a chance to review this until next week but I wanted to mention that I saw and appreciate it :-)

> I think there should be better ways to streamline commit history...

You could rebase and squash your commits locally and then force push your branch but honestly I wouldn't worry about it. We always use a squash-and-merge when accepting PRs that will collapse all of your individual commits in this branch into one commit brought into master.

cpu:
@orangepizza There are also some `golangci-lint` errors failing this branch in [Travis](https://travis-ci.org/letsencrypt/pebble/builds/503014241):

```
$ golangci-lint run
ca/ca.go:14: File is not `goimports`-ed (goimports)
	"time"
va/va.go:91: File is not `goimports`-ed (goimports)
	Identifier string
wfe/wfe.go:1187: File is not `goimports`-ed (goimports)
		}	else {
		// Non-wildcard authorizations get all of the enabled challenge types
		enabledChallenges = []string{acme.ChallengeHTTP01, acme.ChallengeTLSALPN01, acme.ChallengeDNS01}
wfe/wfe.go:1022:62: `produciton` is a misspelling of `production` (misspell)
		//ip check, we'll need to check for exclude private IPs on produciton, but I think this works for pebble
		                                                           ^
The command "golangci-lint run" exited with 1.
```

If you could take a crack at fixing those before I dive in to review that would be awesome. Thanks again!

orangepizza:
about last added tls-alpn:
I thought this was enough but section 6 blocks this - SNI of request should be rdns arpa address, not literal ip address. I wonder how should I make it works?

felixfontein:
I had some problems with IPv6 addresses, more specifically `::1`: it sometimes works, and sometimes doesn't. When it doesn't work, it produces an error during the `finalize-order` call: `{'type': 'urn:ietf:params:acme:error:unauthorized', 'detail': 'CSR is missing Order IP \"::1\"', 'status': 403}"`.

felixfontein:
@orangepizza about rdns address: I don't know if the go standard libs have an utility for producing them, but manually constructing them from the IP addresses isn't that hard. I've done it here: https://github.com/ansible/ansible/pull/53228/files#diff-e25d9f5862c36c63135426f2412eb0b2R541

orangepizza:
> 
> 
> @orangepizza about rdns address: I don't know if the go standard libs have an utility for producing them, but manually constructing them from the IP addresses isn't that hard. I've done it here: https://github.com/ansible/ansible/pull/53228/files#diff-e25d9f5862c36c63135426f2412eb0b2R541

I found net.reverseaddr in document but some reason go says it doesnt exsit. :(
https://golang.org/pkg/net/?m=all#reverseaddr

wfe/wfe.go
```diff
@@ -1048,9 +1048,19 @@ func (wfe *WebFrontEndImpl) verifyOrder(order *core.Order) *acme.ProblemDetails
 	}
 	// Check that all of the identifiers in the new-order are DNS type
 	for _, ident := range idents {
+		//ip check, we'll need to check for exclude private IPs on production, but I think this works for pebble
```
felixfontein: There's no need for this check in Pebble.

wfe/wfe.go
```diff
@@ -1489,22 +1498,50 @@ func (wfe *WebFrontEndImpl) FinalizeOrder(
...
+	var orderDNSs, orderIPs []string
+	for _, ident := range orderIdentifiers {
+		switch ident.Type {
+		case acme.IdentifierDNS:
+			orderDNSs = append(orderDNSs, ident.Value)
+		case acme.IdentifierIP:
+			orderIPs = append(orderIPs, ident.Value)
+		default:
+			wfe.sendError(acme.UnauthorizedProblem(
+				fmt.Sprintf("Order includes ilegal ident type %s", ident.Type)), response)
```
felixfontein:
```suggestion
				fmt.Sprintf("Order includes illegal ident type %s", ident.Type)), response)
```

wfe/wfe.go
```diff
@@ -1489,22 +1498,50 @@ func (wfe *WebFrontEndImpl) FinalizeOrder(
...
+		if name != csrDNSs[i] {
 			wfe.sendError(acme.UnauthorizedProblem(
 				fmt.Sprintf("CSR is missing Order domain %q", name)), response)
 			return
 		}
 	}
+	for i, IP := range orderIPs {
+		if !csrIPs[i].Equal(net.ParseIP(IP)) {
+			wfe.sendError(acme.UnauthorizedProblem(
+				fmt.Sprintf("CSR is missing Order IP %q", IP)), response)
```
felixfontein: I think the reason for me getting this error is that you don't sort the list of IP addresses. (`uniqueLowerNames()` sorts the list of DNS names, that's why the same loop for DNS names works fine.) Thus if there is more than one IP address in the list, and the order of the IPs in the CSR differs from the order of the IPs in the ACME order differ, you'll get this error.

wfe/wfe.go
```diff
@@ -1753,12 +1790,14 @@ func (wfe *WebFrontEndImpl) validateAuthzForChallenge(authz *core.Authorization)
...
 	ident := authz.Identifier
-	if ident.Type != acme.IdentifierDNS {
+	switch ident.Type {
+	case acme.IdentifierDNS:
+	case acme.IdentifierIP:
+	default:
 		return nil, acme.MalformedProblem(
-			fmt.Sprintf("Authorization identifier was type %s, only %s is supported",
-				ident.Type, acme.IdentifierDNS))
+			fmt.Sprintf("Authorization identifier was type %s, only %s,%s is supported",
```
felixfontein:
```suggestion
			fmt.Sprintf("Authorization identifier was type %s, only %s and %s are supported",
```

felixfontein:
> I found net.reverseaddr in document but some reason go says it doesnt exsit. :(
> https://golang.org/pkg/net/?m=all#reverseaddr

From how I understand go, this is an internal function (because it starts with a lowercase letter) and isn't exported, so you cannot use it outside of the `net` package. (Look at the URL to that page, it has `?m=all` in there, which apparently shows also private methods of that package. If you remove that from the URL, you'll get the list of exported methods, and that doesn't include `reverseaddr`.)

cpu:
> I don't know if the go standard libs have an utility for producing them, but manually constructing them from the IP addresses isn't that hard. I've done it here: 

If its helpful I wrote some Go code to do this task in the opposite direction last week: 
https://github.com/zmap/zlint/blob/edc14b276d77a052891716312a2b580e4bbb87c4/lints/lint_subject_contains_reserved_arpa_ip.go#L109-L154

It shouldn't be too hard to take that and go the other direction from an IP to the rDNS labels.

wfe/wfe.go
orangepizza: yes so I didn't included it there

wfe/wfe.go
orangepizza: I was assumed that both csr domain name and order doesn't sorted, and thought order should've follow the SAN's order in csr to be vaild. add a sort at both order and csr IP list

wfe/wfe.go
felixfontein: I would remove this comment, or heavily rewrite it, since in the current way it's pretty confusing (especially who "I" is).

felixfontein: (In this PR, it's clear, but once this is merged, it's confusing :) )

va/va.go
```diff
@@ -537,3 +542,26 @@ func (va VAImpl) fetchHTTP(identifier string, token string) ([]byte, string, *ac
...
 	return body, url.String(), nil
 }
+func alpnaddr(addr string) (arpa string, err error) {
+	const hexDigit = "0123456789abcdef"
+	ip := net.ParseIP(addr)
+	if ip == nil {
+		return "", &net.DNSError{Err: "unrecognized address", Name: addr}
+	}
+	if ip.To4() != nil {
+		return string(uint(ip[15])) + "." + string(uint(ip[14])) + "." + string(uint(ip[13])) + "." + string(uint(ip[12])) + ".in-addr.arpa.", nil
```
felixfontein: Should there really be the trailing dot (for `in-addr.arpa.`)? I mean, for regular DNS names, the trailing dot isn't there as well, right?

orangepizza: techtically the draft itself has trailing dot
For example if the IP address being validated is
   2001:db8::1 the SNI HostName should contain "1.0.0.0.0.0.0.0.0.0.0.0.
   0.0.0.0.0.0.0.0.0.0.0.0.8.b.d.0.1.0.0.2.ip6.arpa.

felixfontein: Technically, you should also write `example.com.` instead of `example.com`. But for certificates, one leaves the final dot away.

orangepizza: but draft itself explicitly included the dot and have word 'should' so ...

felixfontein: Hmm, interesting. The draft explicitly mentions the trailing dot in its example: https://tools.ietf.org/html/draft-ietf-acme-ip-05#section-6
@cpu what do you think, should the dot be there or not?

orangepizza:
While I'm implementing SNI requirement of draft I doubt it's good idea to set sni this way. I think it may be able to broken like TLS-SNI-01 did.
at worst case of shared hosting that broken at TLS-SNI that anybody can upload any cert with any domain name, and backend webserver updated to version that "supports" tls-alpn protocal, couldn't anyone upload a alpn cert with rDNS name and claim a cert for server's ip address that he doesn't control?

felixfontein:
The backend web server explicitly has to know about the TLS-ALPN-01 protocol. If it implements that wrong, it's a problem of the webserver / the provider, not of the protocol. The big difference between TLS-ALPN-01 and TLS-SNI-01 is that TLS-SNI-01 didn't require explicit webserver support, as opposed to TLS-ALPN-01.

va/va.go
felixfontein: I guess it doesn't really matters, since this isn't put into the certificate, but "only" used in the HTTP protocol while validating the challenge.

ca/ca.go
```diff
@@ -156,12 +157,14 @@ func (ca *CAImpl) newIntermediateIssuer() error {
...
-func (ca *CAImpl) newCertificate(domains []string, key crypto.PublicKey, accountID string) (*core.Certificate, error) {
+func (ca *CAImpl) newCertificate(domains []string, ips []net.IP, key crypto.PublicKey, accountID string) (*core.Certificate, error) {
 	var cn string
 	if len(domains) > 0 {
 		cn = domains[0]
+	} else if len(ips) > 0 {
+		cn = ips[0].String()
 	} else {
-		return nil, fmt.Errorf("must specify at least one domain name")
+		return nil, fmt.Errorf("must specify at least one domain name or IPs")
```
cpu: small plurality nit:
```suggestion
		return nil, fmt.Errorf("must specify at least one domain name or IP address")
```

va/va.go
```diff
@@ -87,13 +87,17 @@ func certNames(cert *x509.Certificate) string {
 		names = append(names, cert.Subject.CommonName)
 	}
 	names = append(names, cert.DNSNames...)
+	for i := range cert.IPAddresses {
```
cpu: I don't think its worthwhile updating `certNames` to also include IP addresses in the joined names, it's only used by `validateTLSALPN01` and only when presented with an incorrect validation certificate. Can you back this diff out?

va/va.go
```diff
@@ -87,13 +87,17 @@ func certNames(cert *x509.Certificate) string {
...
+	}
 	return strings.Join(names, ", ")
 }
 
 type vaTask struct {
-	Identifier string
-	Challenge  *core.Challenge
-	Account    *core.Account
+	Identifier     string
+	IdentifierType string
```
cpu: I think it makes sense to add a new type to the `acme` package called `Identifier` that holds both a `Value` (called`Identifier` here) and a `Type` (called `IdentifierType` here) field. The `vaTask` can be updated to have a single `Identifier` field of type `acme.Identifier`.

va/va.go
```diff
@@ -151,11 +155,12 @@ func New(
 	return va
 }
 
-func (va VAImpl) ValidateChallenge(ident string, chal *core.Challenge, acct *core.Account) {
+func (va VAImpl) ValidateChallenge(ident string, chal *core.Challenge, acct *core.Account, identType string) {
```
cpu: If the `acme.Identifier` type I mentioned above is added it could be used to replace the two `ident` and `identType` arguments with one `acme.Identifier` argument to simplify this function signature a little bit.

va/va.go
```diff
@@ -341,14 +346,19 @@ func (va VAImpl) validateDNS01(task *vaTask) *core.ValidationRecord {
 func (va VAImpl) validateTLSALPN01(task *vaTask) *core.ValidationRecord {
 	portString := strconv.Itoa(va.tlsPort)
 	hostPort := net.JoinHostPort(task.Identifier, portString)
-
+	var serverNameIdentifier string
+	if task.IdentifierType == acme.IdentifierDNS {
+		serverNameIdentifier = task.Identifier
+	} else {
+		serverNameIdentifier, _ = alpnaddr(task.Identifier)
```
cpu: This is the only call site for the `alpnaddr` function and it throws away the `error` return. I think either `alpnaddr` should be updated to not return an `error` or this call site should be changed to handle the error in some way rather than assign it to `_`.

va/va.go
```diff
@@ -374,7 +384,14 @@ func (va VAImpl) validateTLSALPN01(task *vaTask) *core.ValidationRecord {
 	leafCert := certs[0]
 
 	// Verify SNI - certificate returned must be issued only for the domain we are verifying.
-	if len(leafCert.DNSNames) != 1 || !strings.EqualFold(leafCert.DNSNames[0], task.Identifier) {
+	namematch := false
+	switch task.IdentifierType {
```
cpu: Above you used an `if`/`else` to switch on identifier type, here you're using `switch`. Can you update one to be consistent with the other?

va/va.go
```diff
@@ -374,7 +384,14 @@ func (va VAImpl) validateTLSALPN01(task *vaTask) *core.ValidationRecord {
 	leafCert := certs[0]
 
 	// Verify SNI - certificate returned must be issued only for the domain we are verifying.
-	if len(leafCert.DNSNames) != 1 || !strings.EqualFold(leafCert.DNSNames[0], task.Identifier) {
+	namematch := false
```
cpu: nit: I usually prefer to have the implicit default instead of an explicit initialization for `bool` vars that will be flipped in later stages of execution.
```suggestion
	var namematch bool
```

va/va.go
```diff
@@ -374,7 +384,14 @@ func (va VAImpl) validateTLSALPN01(task *vaTask) *core.ValidationRecord {
...
+	switch task.IdentifierType {
+	case acme.IdentifierDNS:
+		namematch = len(leafCert.DNSNames) == 1 && strings.EqualFold(leafCert.DNSNames[0], task.Identifier)
+	case acme.IdentifierIP:
+		namematch = len(leafCert.IPAddresses) == 1 && leafCert.IPAddresses[0].Equal(net.ParseIP(task.Identifier))
```
cpu: I'm slightly nervous about this `namematch` implementation for IP addresses. `net.ParseIP` can return `nil` if the IP is invalid. Can you document the assumption that the `task.Identifier` value has already been validated or handle the case where the `parse` returns nil?

va/va.go
```diff
@@ -465,10 +482,11 @@ func (va VAImpl) validateHTTP01(task *vaTask) *core.ValidationRecord {
 // purpose HTTP function
 func (va VAImpl) fetchHTTP(identifier string, token string) ([]byte, string, *acme.ProblemDetails) {
 	path := fmt.Sprintf("%s%s", acme.HTTP01BaseURL, token)
+	portString := strconv.Itoa(va.httpPort)
 
 	url := &url.URL{
 		Scheme: "http",
-		Host:   fmt.Sprintf("%s:%d", identifier, va.httpPort),
+		Host:   net.JoinHostPort(identifier, portString),
```
cpu: I think this diff was required to handle adding a port to an IP address (particularly IPv6) in the correct way. Is that accurate?

va/va.go
```diff
@@ -524,3 +542,26 @@ func (va VAImpl) fetchHTTP(identifier string, token string) ([]byte, string, *ac
 
 	return body, url.String(), nil
 }
+func alpnaddr(addr string) (arpa string, err error) {
```
cpu: A few nits here:

1) Can you add an empty newline between this function's start and the previous function's end?
2) I think `alpnaddr` isn't the most descriptive name. I think it's computing an SNI value to send in a TLS-ALPN-01 validation request specifically for IP address validation, is that correct? If so then I think the new name needs to capture that its specific to ACME-IP.
3) Can you add a Godoc comment on this function? It's a significant bit of code and could use the added context/description.
4) Please don't use named return values (`arpa`, `err`). We avoid those in Boulder/Pebble as a matter of style. Thanks!

va/va.go
```diff
@@ -524,3 +542,26 @@ func (va VAImpl) fetchHTTP(identifier string, token string) ([]byte, string, *ac
...
+	const hexDigit = "0123456789abcdef"
```
cpu: You should be able to use `fmt` and an `%x` parameter to avoid needing to do this hex conversion yourself.

va/va.go
```diff
@@ -524,3 +542,26 @@ func (va VAImpl) fetchHTTP(identifier string, token string) ([]byte, string, *ac
...
+	ip := net.ParseIP(addr)
+	if ip == nil {
+		return "", &net.DNSError{Err: "unrecognized address", Name: addr}
+	}
+	if ip.To4() != nil {
+		return strconv.Itoa(int(ip[15])) + "." + strconv.Itoa(int(ip[14])) + "." + strconv.Itoa(int(ip[13])) + "." + strconv.Itoa(int(ip[12])) + ".in-addr.arpa.", nil
```
cpu: I'm surprised by the indexes being used here for a 4 byte IP address. Can you add a comment explaining how the IP `[]byte` slice is represented and why this is the correct construction for the reverse name?

This line is also quite long and would be more readable split up. Thanks!

va/va.go
```diff
@@ -524,3 +542,26 @@ func (va VAImpl) fetchHTTP(identifier string, token string) ([]byte, string, *ac
...
+	// Add it, in reverse, to the buffer
+	for i := len(ip) - 1; i >= 0; i-- {
+		v := ip[i]
+		buf = append(buf, hexDigit[v&0xF],
+			'.',
+			hexDigit[v>>4],
+			'.')
+	}
+	// Append "ip6.arpa." and return (buf already has the final .)
+	buf = append(buf, "ip6.arpa."...)
```
cpu: Can you add some RFC references to justify the way these addresses are being constructed?

va/va.go
```diff
@@ -524,3 +542,26 @@ func (va VAImpl) fetchHTTP(identifier string, token string) ([]byte, string, *ac
...
+	}
+	if ip.To4() != nil {
+		return strconv.Itoa(int(ip[15])) + "." + strconv.Itoa(int(ip[14])) + "." + strconv.Itoa(int(ip[13])) + "." + strconv.Itoa(int(ip[12])) + ".in-addr.arpa.", nil
+	}
+	// Must be IPv6
+	buf := make([]byte, 0, len(ip)*4+len("ip6.arpa."))
+	// Add it, in reverse, to the buffer
+	for i := len(ip) - 1; i >= 0; i-- {
+		v := ip[i]
+		buf = append(buf, hexDigit[v&0xF],
```
cpu: If my comment about using `%x` doesn't help simplify this logic could you please comment the bitwise operations to make it easier to read through the implementation?

wfe/wfe.go
```diff
@@ -1048,9 +1049,18 @@ func (wfe *WebFrontEndImpl) verifyOrder(order *core.Order) *acme.ProblemDetails
 	}
 	// Check that all of the identifiers in the new-order are DNS type
 	for _, ident := range idents {
+		if ident.Type == acme.IdentifierIP {
+			ip := net.ParseIP(ident.Value)
+			if ip == nil {
+				return acme.MalformedProblem(fmt.Sprintf(
+					"Order included illegal IP address value: %q\n",
```
cpu: nit:
```suggestion
					"Order included malformed IP type identifier value: %q\n",
```

wfe/wfe.go
```diff
@@ -1048,9 +1049,18 @@ func (wfe *WebFrontEndImpl) verifyOrder(order *core.Order) *acme.ProblemDetails
...
+					ident.Value))
+			}
+			continue
+		}
 		if ident.Type != acme.IdentifierDNS {
 			return acme.MalformedProblem(fmt.Sprintf(
-				"Order included non-DNS type identifier: type %q, value %q",
+				"Order included illegal type identifier: type %q, value %q",
```
cpu:
```suggestion
				"Order included unsupported type identifier: type %q, value %q",
```

wfe/wfe.go
```diff
@@ -1199,8 +1209,14 @@ func (wfe *WebFrontEndImpl) makeChallenges(authz *core.Authorization, request *h
...
 		chals = []*core.Challenge{chal}
 	} else {
-		// Non-wildcard authorizations get all of the enabled challenge types
-		enabledChallenges := []string{acme.ChallengeHTTP01, acme.ChallengeTLSALPN01, acme.ChallengeDNS01}
+		// IP addresses get HTTP-01 and TLS-ALPN challenges
+		var enabledChallenges []string
+		if authz.Identifier.Value == acme.IdentifierIP {
+			enabledChallenges = []string{acme.ChallengeHTTP01, acme.ChallengeTLSALPN01}
+		} else {
+			// Non-wildcard authorizations get all of the enabled challenge types
```
cpu: I think this comment needs to be updated
```suggestion
			// Non-wildcard, non-IP identifier authorizations get all of the enabled challenge types
```

wfe/wfe.go
```diff
@@ -1436,7 +1443,9 @@ func (wfe *WebFrontEndImpl) FinalizeOrder(
 	orderAccountID := existingOrder.AccountID
 	orderStatus := existingOrder.Status
 	orderExpires := existingOrder.ExpiresDate
-	orderNames := existingOrder.Names
+	// not using ordernames now, can't split dns and ips
+	//	orderNames := existingOrder.Names
```
cpu: Please don't leave commented code in-file. It should be removed entirely or left as-is. In this case I think it should stay. I think this change will introduce a regression that would allow an ACME client to create an order with duplicated identifiers. Can you expand on why you weren't able to make this work with IP address identifiers, I'm not sure I understand based on the comment text.

wfe/wfe.go
```diff
@@ -1489,22 +1498,59 @@ func (wfe *WebFrontEndImpl) FinalizeOrder(
...
+	var orderIPs []net.IP
+	for _, ident := range orderIdentifiers {
+		switch ident.Type {
+		case acme.IdentifierDNS:
+			orderDNSs = append(orderDNSs, ident.Value)
+		case acme.IdentifierIP:
+			orderIPs = append(orderIPs, net.ParseIP(ident.Value))
+		default:
+			wfe.sendError(acme.UnauthorizedProblem(
+				fmt.Sprintf("Order includes illegal ident type %s", ident.Type)), response)
```
cpu: I try to avoid "illegal" since its somewhat vague.
```suggestion
				fmt.Sprintf("Order includes unknown identifier type %s", ident.Type)), response)
```

wfe/wfe.go
```diff
@@ -1489,22 +1498,59 @@ func (wfe *WebFrontEndImpl) FinalizeOrder(
...
+		}
+	}
+	//first sort IPs
```
cpu: nit: always include a space between `//` and the comment text.
```suggestion
	// first sort IPs
```

wfe/wfe.go
```diff
@@ -1048,9 +1049,18 @@ func (wfe *WebFrontEndImpl) verifyOrder(order *core.Order) *acme.ProblemDetails
 	}
 	// Check that all of the identifiers in the new-order are DNS type
```
cpu: I think this comment needs updating

wfe/wfe.go
```diff
@@ -1116,12 +1126,12 @@ func (wfe *WebFrontEndImpl) makeAuthorizations(order *core.Order, request *http.
 	// Lock the order for reading
 	order.RLock()
 	// Create one authz for each name in the order's parsed CSR
-	for _, name := range order.Names {
+	for _, name := range order.Identifiers {
```
cpu: I think before this step Pebble also needs to ensure there aren't any duplicate IP address identifiers in the order.

wfe/wfe.go
```diff
@@ -1489,22 +1498,59 @@ func (wfe *WebFrontEndImpl) FinalizeOrder(
...
+		default:
+			wfe.sendError(acme.UnauthorizedProblem(
+				fmt.Sprintf("Order includes illegal ident type %s", ident.Type)), response)
+		}
+	}
+	//first sort IPs
+	sort.Slice(orderIPs, func(i, j int) bool {
+		return bytes.Compare(orderIPs[i], orderIPs[j]) < 0
+	})
+	//and make uniqueLowerNames for DNSNames
```
cpu:
```suggestion
	// and make uniqueLowerNames for DNSNames
```

wfe/wfe.go
```diff
@@ -1489,22 +1498,59 @@ func (wfe *WebFrontEndImpl) FinalizeOrder(
...
+	orderDNSs = uniqueLowerNames(orderDNSs)
 	// Check that the CSR has the same number of names as the initial order contained
-	csrNames := uniqueLowerNames(parsedCSR.DNSNames)
-	if len(csrNames) != len(orderNames) {
+	csrDNSs := uniqueLowerNames(parsedCSR.DNSNames)
+	csrIPs := parsedCSR.IPAddresses
+	//sort this too
```
cpu:
```suggestion
	// sort this too
```

wfe/wfe.go
```diff
@@ -1489,22 +1498,59 @@ func (wfe *WebFrontEndImpl) FinalizeOrder(
...
+	sort.Slice(csrIPs, func(i, j int) bool {
+		return bytes.Compare(csrIPs[i], csrIPs[j]) < 0
+	})
+	if len(csrDNSs) != len(orderDNSs) {
+		wfe.sendError(acme.UnauthorizedProblem(
+			"Order includes different number of DNSnames than CSR specifies"), response)
```
cpu:
```suggestion
			"Order includes different number of DNSnames identifiers than CSR specifies"), response)
```

wfe/wfe.go
```diff
@@ -1489,22 +1498,59 @@ func (wfe *WebFrontEndImpl) FinalizeOrder(
...
+		return
+	}
+	if len(csrIPs) != len(orderIPs) {
 		wfe.sendError(acme.UnauthorizedProblem(
-			"Order includes different number of names than CSR specifies"), response)
+			"Order includes different number of IP addresses than CSR specifies"), response)
```
cpu:
```suggestion
			"Order includes different number of IP address identifiers than CSR specifies"), response)
```

wfe/wfe.go
```diff
@@ -1753,12 +1799,14 @@ func (wfe *WebFrontEndImpl) validateAuthzForChallenge(authz *core.Authorization)
 	defer authz.RUnlock()
 
 	ident := authz.Identifier
-	if ident.Type != acme.IdentifierDNS {
+	switch ident.Type {
```
cpu: Given that there is no body for most of the cases in this switch I think I'd prefer to see it written as an `if` statement like:

`if ident.Type != acme.IdentifierDNS && ident.Type != acme.IdentifierIP {`

orangepizza:
first round of applying requested change done, can you review this again?

orangepizza:
I think I fixed things that you requested, @cpu. and you review this again?

va/va.go
```diff
@@ -524,3 +540,30 @@ func (va VAImpl) fetchHTTP(identifier string, token string) ([]byte, string, *ac
...
+//this reverseaddr funntion is from net/dnsclient.go form golang
+func reverseaddr(addr string) string {
+	const hexDigit = "0123456789abcdef"
+	ip := net.ParseIP(addr)
+	if ip == nil {
+		return ""
+	}
+	// Apperently IP type in net package saves all ip in ipv6 formant, from biggest byte to smallest. we need last 4 bytes, so ip[15] to ip[12]
+	if ip.To4() != nil {
+		return strconv.Itoa(int(ip[15])) + "." + strconv.Itoa(int(ip[14])) + "." + strconv.Itoa(int(ip[13])) + "." + strconv.Itoa(int(ip[12])) + ".in-addr.arpa."
```
rolandshoemaker: This would be somewhat cleaner as `fmt.Sprintf("%d.%d.%d.%d.in-addr.arpa.", ip[15], ip[14], ip[13], ip[12])`.

va/va.go
```diff
@@ -524,3 +540,30 @@ func (va VAImpl) fetchHTTP(identifier string, token string) ([]byte, string, *ac
...
+	}
+	// Must be IPv6
+	buf := make([]byte, 0, len(ip)*4+len("ip6.arpa."))
```
rolandshoemaker: Rather than doing buffer based string construction like this it'd be somewhat simpler to just use a `[]string` and append each two quads to it in the loop below then do the final construction using `fmt.Sprintf("%s.ip6.arpa.", strings.Join(quads, "."))`.

va/va.go
```diff
@@ -524,3 +540,30 @@ func (va VAImpl) fetchHTTP(identifier string, token string) ([]byte, string, *ac
...
+	// Add it, in reverse, to the buffer
+	// each part of slice has two quads, so manual hex conversion is needed and fmt %x doesn't help
```
rolandshoemaker: I'm not sure this is true, you can use `fmt.Sprintf("%x.%x", x&0f, x>>4)` which would be significantly simpler.

wfe/wfe.go
```diff
@@ -1046,11 +1047,21 @@ func (wfe *WebFrontEndImpl) verifyOrder(order *core.Order) *acme.ProblemDetails
 	if len(idents) == 0 {
 		return acme.MalformedProblem("Order did not specify any identifiers")
 	}
-	// Check that all of the identifiers in the new-order are DNS type
+	// Check that all of the identifiers in the new-order are DNS or IPaddress type
+	// Validity check of ipaddresses are done here.
 	for _, ident := range idents {
+		if ident.Type == acme.IdentifierIP {
+			ip := net.ParseIP(ident.Value)
```
rolandshoemaker: This and line below can be condensed into `if ip := net.ParseIP(ident.Value); ip == nil {...}`.

va/va.go
orangepizza: I can't find good name for this []string. it's single byte, but upper and lower is reversed.

orangepizza:
this will be all of the changes you requested.. I changed the buffer from []byte to []string, but I still think it's still a buffer, so left it's name as is.

va/va.go
```diff
@@ -524,3 +540,25 @@ func (va VAImpl) fetchHTTP(identifier string, token string) ([]byte, string, *ac
 
 	return body, url.String(), nil
 }
+
+//this reverseaddr funntion is from net/dnsclient.go form golang
```
cpu: nit: You're missing a leading space after the `//` here and also have a couple small typos:
``` suggestion
// reverseaddr function is borrowed from net/dnsclient.go[0] and the Go std library.
// [0]: https://golang.org/src/net/dnsclient.go
```

va/va.go
```diff
@@ -524,3 +540,25 @@ func (va VAImpl) fetchHTTP(identifier string, token string) ([]byte, string, *ac
...
+		return ""
+	}
+	// Apperently IP type in net package saves all ip in ipv6 formant, from biggest byte to smallest. we need last 4 bytes, so ip[15] to ip[12]
+	if ip.To4() != nil {
+		return fmt.Sprintf("%d.%d.%d.%d.in-addr.arpa.", ip[15], ip[14], ip[13], ip[12])
+	}
+	// Must be IPv6
+	buf := make([]string, 0, len(ip)+1)
+	// Add it, in reverse, to the buffer
+	// each part of slice has two quads, so manual hex conversion is needed and fmt %x doesn't help
```
cpu: I think this comment about "fmt %x doesn't help" isn't accurate anymore and needs to be removed or updated. You're using @rolandshoemaker's suggestion and have `fmt.Sprintf("%x.%x", ...` in the loop body.

va/va.go
```diff
@@ -524,3 +540,25 @@ func (va VAImpl) fetchHTTP(identifier string, token string) ([]byte, string, *ac
...
+	for i := len(ip) - 1; i >= 0; i-- {
+		buf = append(buf, fmt.Sprintf("%x.%x", ip[i]&0x0F, ip[i]>>4))
+	}
+	// Append "ip6.arpa." and return (buf already has the final .) see RFC3152 for how this address constructed.
```
cpu: nit: without quotes at first glance it seems like this comment has a typo
```suggestion
	// Append "ip6.arpa." and return (buf already has the final '.') see RFC3152 for how this address is constructed.
```

wfe/wfe.go
```diff
@@ -1489,22 +1496,56 @@ func (wfe *WebFrontEndImpl) FinalizeOrder(
...
+	var orderDNSs []string
+	var orderIPs []net.IP
+	for _, ident := range orderIdentifiers {
+		switch ident.Type {
+		case acme.IdentifierDNS:
+			orderDNSs = append(orderDNSs, ident.Value)
+		case acme.IdentifierIP:
+			orderIPs = append(orderIPs, net.ParseIP(ident.Value))
+		default:
+			wfe.sendError(acme.UnauthorizedProblem(
```
cpu: I think this should be a `MalformedProblem` instead of `UnauthorizedProblem`.

wfe/wfe.go
```diff
@@ -1489,22 +1496,56 @@ func (wfe *WebFrontEndImpl) FinalizeOrder(
...
+				fmt.Sprintf("Order includes unknown identifier type %s", ident.Type)), response)
+		}
+	}
+	// sort and deduplicate Names
+	orderDNSs = uniqueLowerNames(orderDNSs)
+	orderIPs = uniqueIPs(orderIPs)
```
cpu: I think the DNS SAN and IP address identifier deduplication for the order should be done in `NewOrder`. Only the CSR needs to be deduplicated at finalization point. The storage should only create orders in normalized form.

wfe/wfe.go
```diff
@@ -1955,6 +1995,22 @@ func uniqueLowerNames(names []string) []string {
 	return unique
 }
 
+// uniqueIPs returns the set of all unique IP addresses in the input.
+// The returned names will be sorted in assending order.
```
cpu: I think this comment says "names" when it should be "IP addresses" and there's another small typo here. 

```suggestion
// The returned IP addresses will be sorted in ascending order.
```

wfe/wfe.go
```diff
@@ -1955,6 +1995,22 @@ func uniqueLowerNames(names []string) []string {
...
+// this internally call UniqueLowerNames as map doesn't support net.IP as key
```
cpu: I think this function would be clearer/faster if you built a map using the `ip.String()` as the key and didn't rely on `uniqueLowerNames`. That will also avoid having to re-parse the IP back from the string.

E.g. something like:
```
uniqMap := make(map[string]net.IP)
for _, ip := range IPs {
  uniqMap[ip.String()] = ip
}
results := make([]net.IP, len(uniqMap))
for k, v := range uniqMap {
  results = append(result, v)
}
return results
```

cpu:
@orangepizza There's also a few conflicts (`va/va.go`, `wfe/wfe.go`) that need to be resolved against master. Thanks!

wfe/wfe.go
orangepizza: I try moving them to NewOrder, but it's look like order is not preserved in storage, failing at CSR test.

orangepizza:
other then it still does sorting at finalizeorder (doing it at initalizeorder showhow breaks order, and it's been there before I started. I think I fixed this round of requested changes.

cpu:
@orangepizza Thanks, I'll take a look again shortly.

felixfontein:
There's a new draft: https://tools.ietf.org/html/draft-ietf-acme-ip-06 There doesn't seem to be any content changes relevant to this PR (and the implementations), though.

BTW, is there a chance this will get merged anytime soon?

cpu:
> BTW, is there a chance this will get merged anytime soon?

Sorry for the delay :sweat: I know this one has lagged. I think its ready to merge soon, I'd just like to get a second review from @jsha first. Hopefully this week or next.

wfe/wfe.go
```diff
@@ -1450,22 +1477,57 @@ func (wfe *WebFrontEndImpl) FinalizeOrder(
...
+				fmt.Sprintf("Order includes unknown identifier type %s", ident.Type)), response)
+		}
+	}
+	// looks like this spliting doesn't preserve order or Identifiers, so sort them again.
+	orderDNSs = uniqueLowerNames(orderDNSs)
+	orderIPs = uniqueIPs(orderIPs)
+
+	csrDNSs := uniqueLowerNames(parsedCSR.DNSNames)
+	csrIPs := uniqueIPs(parsedCSR.IPAddresses)
+	// sort and deduplicate CSR SANs
```
jsha: I think this comment was meant for two lines above.

wfe/wfe.go
```diff
@@ -1450,22 +1477,57 @@ func (wfe *WebFrontEndImpl) FinalizeOrder(
 		return
 	}
 
+	// spliting order identifiers per types
```
jsha:
```suggestion
	// split order identifiers per types
```

wfe/wfe.go
```diff
@@ -1450,22 +1477,57 @@ func (wfe *WebFrontEndImpl) FinalizeOrder(
...
+		case acme.IdentifierDNS:
+			orderDNSs = append(orderDNSs, ident.Value)
+		case acme.IdentifierIP:
+			orderIPs = append(orderIPs, net.ParseIP(ident.Value))
+		default:
+			wfe.sendError(acme.MalformedProblem(
+				fmt.Sprintf("Order includes unknown identifier type %s", ident.Type)), response)
+		}
+	}
+	// looks like this spliting doesn't preserve order or Identifiers, so sort them again.
```
jsha: I'm pretty sure the way you wrote the split code *does* preserve order. Can you produce a concrete example where it doesn't?

wfe/wfe.go
```diff
@@ -1450,22 +1477,57 @@ func (wfe *WebFrontEndImpl) FinalizeOrder(
...
+	orderDNSs = uniqueLowerNames(orderDNSs)
+	orderIPs = uniqueIPs(orderIPs)
+
+	csrDNSs := uniqueLowerNames(parsedCSR.DNSNames)
+	csrIPs := uniqueIPs(parsedCSR.IPAddresses)
+	// sort and deduplicate CSR SANs
+	sort.Slice(csrIPs, func(i, j int) bool {
```
jsha: Since `uniqueLowerNames` sorts its output I think it makes sense for `uniqueIPs` to do the same. Would you move this sort inside `uniqueIPs`?

orangepizza: yes I moved it

wfe/wfe.go
orangepizza: wording was wrong:  looks like saving order to database breaks order of identifiers, for exemple https://travis-ci.org/letsencrypt/pebble/builds/537102621?utm_source=github_status&utm_medium=notification

wfe/wfe.go
cpu: Ok, that makes sense. Thanks for trying!

felixfontein:
Thanks everyone for getting this done! :)

felixfontein:
This PR contained a little bug, I've created a fix in #237.

felixfontein:
There seems to be another bug in this PR, dns-01 validation doesn't seem to work anymore. When I want to validate challenges for `example.com`, TXT lookup fails with `lookup _acme-challenge.{dns example.com}: no such host`.

felixfontein:
(I verified that it still worked with the previous commit before this PR was merged.)

cpu:
> There seems to be another bug in this PR, dns-01 validation doesn't seem to work anymore.

Just to close the loop for anyone that stumbles onto this discussion. Both bugs flagged by @felixfontein have been fixed (by @felixfontein :trophy:!) in #238 and #237.

PR #231, opened 2019-04-12
dep: remove jmhodges/clock dependency.

cpu:
It was inherited from Boulder but was never used anywhere in Pebble.


jsha:
I'm fine with this for now, since I think we shouldn't have dependencies and fields we're not currently using. But I'd like to stay open to the possibility we'll bring it back in the future. The ability to manipulate Boulder's notion of time has been very useful for our integration tests, and I suspect some clients will want to do similar in Pebble.

cpu:
>  The ability to manipulate Boulder's notion of time has been very useful for our integration tests, and I suspect some clients will want to do similar in Pebble.

Maybe! I'm open to bringing it back (or something similar) in the future if there are tests being written that need it. Pebble doesn't do authorization reuse or ACME-CAA or any of the things that jump to mind where we've had to use fake time in Boulder. It would probably just be useful for testing expiration and there's likely a way to do that without needing all of the complexity of passing around a fake clock.

jsha:
Yep! The main thing I'm thinking of here is that clients need to test how they behave when a certificate is near expiration. But I guess most clients are probably doing that by manipulating the *client* clock, and it doesn't really matter if they manipulate the *server* clock (though if they get back a certificate that doesn't have expected dates they might get into a renewal loop).

Issue #230, opened 2019-04-08
Test Pebble with --race and concurrent client requests


cpu:
There's been at least one (#228) bug where Pebble wasn't properly locking resources to prevent data races. I think this has largely flown under the radar because the the very small `chisel` test in CI is single-threaded and we're installing Pebble in CI without `--race`: https://github.com/letsencrypt/pebble/blob/194da59119430fff75e3249f07e109bb2ba69c24/.travis.yml#L37

Here's a short-term list of things to do:

1. Use `--race` in the `go install` during CI.
1. Run the Boulder `load-generator` against Pebble briefly after the chisel issuance.
1. Fix any bugs it shakes out.

Item 2 wasn't possible up until ~last week (https://github.com/letsencrypt/boulder/issues/4086) but the timing works out well and is probably the fastest way to jam concurrent ACME client requests into the CI pipeline.

It would probably make sense to revisit the code with an eye for this problem and consider some refactoring that could be done to make the locking throughout the codebase less onerous.

Issue #228, opened 2019-04-08
Data race in WFE


bluecmd:
Hi,

I'm running into this data race warning:

```
WARNING: DATA RACE
Write at 0x00c000132cd8 by goroutine 26:
  github.com/u-root/u-bmc/vendor/github.com/letsencrypt/pebble/va.VAImpl.setAuthzValid()
      /home/bluecmd/go/src/github.com/u-root/u-bmc/vendor/github.com/letsencrypt/pebble/va/va.go:188 +0x26e
  github.com/u-root/u-bmc/vendor/github.com/letsencrypt/pebble/va.VAImpl.process()
      /home/bluecmd/go/src/github.com/u-root/u-bmc/vendor/github.com/letsencrypt/pebble/va/va.go:256 +0x899

Previous read at 0x00c000132cd8 by goroutine 15:
  github.com/u-root/u-bmc/vendor/github.com/letsencrypt/pebble/wfe.(*WebFrontEndImpl).Authz()
      /home/bluecmd/go/src/github.com/u-root/u-bmc/vendor/github.com/letsencrypt/pebble/wfe/wfe.go:1615 +0x40b
  github.com/u-root/u-bmc/vendor/github.com/letsencrypt/pebble/wfe.(*WebFrontEndImpl).Authz-fm()
      /home/bluecmd/go/src/github.com/u-root/u-bmc/vendor/github.com/letsencrypt/pebble/wfe/wfe.go:1540 +0x73
  github.com/u-root/u-bmc/vendor/github.com/letsencrypt/pebble/wfe.(*WebFrontEndImpl).HandleFunc.func1()
      /home/bluecmd/go/src/github.com/u-root/u-bmc/vendor/github.com/letsencrypt/pebble/wfe/wfe.go:211 +0x976
```

I have no experience in Pebble's code base, but my guess is that authz should be locked like this:

```
diff --git a/wfe/wfe.go b/wfe/wfe.go
index 762153f..6dd9114 100644
--- a/wfe/wfe.go
+++ b/wfe/wfe.go
@@ -1556,6 +1556,8 @@ func (wfe *WebFrontEndImpl) Authz(
                response.WriteHeader(http.StatusNotFound)
                return
        }
+       authz.Lock()
+       defer authz.Unlock()
 
        // If the postData is not a POST-as-GET, treat this as case A) and update
        // the authorization based on the postData
```

Locking that seems to make the race test pass (at least it hasn't failed for ~30 runs).

[...]

PR #232, opened 2019-04-12
fix dataraces, add -race install, add load-gen.

cpu:
The `Authz()` method of the WFE was racey. First because it didn't lock the authorizations and orders it was working with. Second because the handling of displaying authorization challenges was working with `acme.Challenge` objects owned by `core.Challenge`'s that should have been locked for reading but were not. This mean the VA would datarace with the WFE when updating a validated challenge status.

To prevent future occurrences `travis.yml` is updated to install Pebble with the race detector enabled, and to run Pebble such that it will exit non-zero if a race is detected.

Since `Chisel2.py` is single threaded the Boulder `load-generator` is used for a short duration to drive concurrent request traffic. In practice before fixing the dataraces I found this would crash Pebble <30s.

Resolves https://github.com/letsencrypt/pebble/issues/230
Resolves https://github.com/letsencrypt/pebble/issues/228

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,15 +29,16 @@ before_install:
 install:
   # Install `golangci-lint` using their installer script
   - curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b $(go env GOPATH)/bin v1.15.0
-  # Install `cover` and `goveralls` without `GO111MODULE` enabled so that we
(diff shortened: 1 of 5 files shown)


cpu:
Going to close this to iterate. I think I overlooked a side effect of changing the challenge pointers.

cpu:
> Going to close this to iterate.

All better now :tada: CI is happy and I was able to run lots of concurrent load to a race detector build of Pebble without any races.

PR #234, opened 2019-05-12
allow to offer alternate chains (RFC8555 7.4.2)

felixfontein:
Allows to offer alternate chains for a certificate according to RFC8555, Section 7.4.2. For this, multiple root and intermediate certs are created (all with the same intermediate private key), and `alternative` relation links are offered during certificate download. The `/root` endpoint has also been extended to allow downloading the alternate roots (`?no=0`, `?no=1` etc.); it also provides `link` headers for its alternate forms.

This feature can be enabled by setting the environment variable `PEBBLE_ALTERNATE_ROOTS` to something larger than 0. For example, with `PEBBLE_ALTERNATE_ROOTS=2`:

```.sh
$ curl -i -k https://0.0.0.0:14000/root
HTTP/2 200 
content-type: application/pem-certificate-chain; charset=utf-8
link: <https://0.0.0.0:14000/dir>;rel="index"
link: <https://0.0.0.0:14000/root%3Fno=0>;rel="alternate"
link: <https://0.0.0.0:14000/root%3Fno=1>;rel="alternate"
content-length: 1107
date: Sun, 12 May 2019 15:05:37 GMT

-----BEGIN CERTIFICATE-----
...

$ curl -i -k https://0.0.0.0:14000/root?no=1
HTTP/2 200 
content-type: application/pem-certificate-chain; charset=utf-8
link: <https://0.0.0.0:14000/dir>;rel="index"
link: <https://0.0.0.0:14000/root>;rel="alternate"
link: <https://0.0.0.0:14000/root%3Fno=0>;rel="alternate"
content-length: 1107
date: Sun, 12 May 2019 15:06:07 GMT

-----BEGIN CERTIFICATE-----
...
```

(For context, see https://community.letsencrypt.org/t/please-update-the-certificate-compatibility-page-to-include-information-about-the-new-isrg-roots/91436/18)

### DIFF
--- a/ca/ca.go
+++ b/ca/ca.go
@@ -304,9 +304,8 @@ func (ca *CAImpl) GetNumberOfRootCerts() int {
 func (ca *CAImpl) getChain(no int) *chain {
 	if 0 <= no && no < len(ca.chains) {
 		return ca.chains[no]
-	} else {
-		return nil
 	}
+	return nil
 }
 
 func (ca *CAImpl) GetRootCert(no int) *core.Certificate {
--- a/ca/ca.go
+++ b/ca/ca.go
@@ -63,7 +63,7 @@ func makeKey() (*rsa.PrivateKey, error) {
 	return key, nil
 }
 
-func (chain *chain) makeRootCert(
+func makeRootCert(
 	ca *CAImpl,
 	subjectKey crypto.Signer,
 	subjCNPrefix string,
@@ -120,24 +120,23 @@ func (chain *chain) makeRootCert(
 	return newCert, nil
 }
 
-func (chain *chain) newRootIssuer(ca *CAImpl) error {
+func newRootIssuer(ca *CAImpl) (*issuer, error) {
 	// Make a root private key
 	rk, err := makeKey()
 	if err != nil {
-		return err
+		return nil, err
 	}
 	// Make a self-signed root certificate
-	rc, err := chain.makeRootCert(ca, rk, rootCAPrefix, nil)
+	rc, err := makeRootCert(ca, rk, rootCAPrefix, nil)
 	if err != nil {
-		return err
+		return nil, err
 	}
 
-	chain.root = &issuer{
+	ca.log.Printf("Generated new root issuer with serial %s\n", rc.ID)
+	return &issuer{
 		key:  rk,
 		cert: rc,
-	}
-	ca.log.Printf("Generated new root issuer with serial %s\n", rc.ID)
-	return nil
+	}, nil
 }
 
 func (chain *chain) newIntermediateIssuer(ca *CAImpl, ik crypto.Signer) error {
@@ -146,7 +145,7 @@ func (chain *chain) newIntermediateIssuer(ca *CAImpl, ik crypto.Signer) error {
 	}
 
 	// Make an intermediate certificate with the root issuer
-	ic, err := chain.makeRootCert(ca, ik, intermediateCAPrefix, chain.root)
+	ic, err := makeRootCert(ca, ik, intermediateCAPrefix, chain.root)
 	if err != nil {
 		return err
 	}
@@ -160,9 +159,11 @@ func (chain *chain) newIntermediateIssuer(ca *CAImpl, ik crypto.Signer) error {
 
 func newChain(ca *CAImpl, ik crypto.Signer) *chain {
 	chain := &chain{}
-	if err := chain.newRootIssuer(ca); err != nil {
+	root, err := newRootIssuer(ca)
+	if err != nil {
 		panic(fmt.Sprintf("Error creating new root issuer: %s", err.Error()))
 	}
+	chain.root = root
 	if err := chain.newIntermediateIssuer(ca, ik); err != nil {
 		panic(fmt.Sprintf("Error creating new intermediate issuer: %s", err.Error()))
 	}
--- a/ca/ca.go
+++ b/ca/ca.go
@@ -29,8 +29,7 @@ type CAImpl struct {
 	db               *db.MemoryStore
 	ocspResponderURL string
 
-	mainChain         *chain
-	alternativeChains []*chain
+	chains []*chain
 }
 
 type chain struct {
@@ -111,7 +110,8 @@ func makeRootCert(
 		DER:  der,
 	}
 	if signer != nil && signer.cert != nil {
-		newCert.Issuer = signer.cert
+		newCert.Issuers = make([]*core.Certificate, 1)
+		newCert.Issuers[0] = signer.cert
 	}
 	_, err = ca.db.AddCertificate(newCert)
 	if err != nil {
@@ -181,7 +181,7 @@ func (ca *CAImpl) newCertificate(domains []string, ips []net.IP, key crypto.Publ
 		return nil, fmt.Errorf("must specify at least one domain name or IP address")
 	}
 
-	issuer := ca.mainChain.intermediate
+	issuer := ca.chains[0].intermediate
 	if issuer == nil || issuer.cert == nil {
 		return nil, fmt.Errorf("cannot sign certificate - nil issuer")
 	}
@@ -216,19 +216,18 @@ func (ca *CAImpl) newCertificate(domains []string, ips []net.IP, key crypto.Publ
 		return nil, err
 	}
 
-	altIssuers := make([]*core.Certificate, len(ca.alternativeChains))
-	for i := 0; i < len(ca.alternativeChains); i++ {
-		altIssuers[i] = ca.alternativeChains[i].intermediate.cert
+	issuers := make([]*core.Certificate, len(ca.chains))
+	for i := 0; i < len(ca.chains); i++ {
+		issuers[i] = ca.chains[i].intermediate.cert
 	}
 
 	hexSerial := hex.EncodeToString(cert.SerialNumber.Bytes())
 	newCert := &core.Certificate{
-		ID:                 hexSerial,
-		AccountID:          accountID,
-		Cert:               cert,
-		DER:                der,
-		Issuer:             issuer.cert,
-		AlternativeIssuers: altIssuers,
+		ID:        hexSerial,
+		AccountID: accountID,
+		Cert:      cert,
+		DER:       der,
+		Issuers:   issuers,
 	}
 	_, err = ca.db.AddCertificate(newCert)
 	if err != nil {
@@ -252,12 +251,9 @@ func New(log *log.Logger, db *db.MemoryStore, ocspResponderURL string, alternate
 	if err != nil {
 		panic(fmt.Sprintf("Error creating new intermediate private key: %s", err.Error()))
 	}
-	ca.mainChain = newChain(ca, ik)
-
-	// Create alternative chains
-	ca.alternativeChains = make([]*chain, alternateRoots)
-	for i := 0; i < len(ca.alternativeChains); i++ {
-		ca.alternativeChains[i] = newChain(ca, ik)
+	ca.chains = make([]*chain, 1+alternateRoots)
+	for i := 0; i < len(ca.chains); i++ {
+		ca.chains[i] = newChain(ca, ik)
 	}
 	return ca
 }
@@ -301,15 +297,13 @@ func (ca *CAImpl) CompleteOrder(order *core.Order) {
 	order.Unlock()
 }
 
-func (ca *CAImpl) GetNumberOfAlternativeRootCerts() int {
-	return len(ca.alternativeChains)
+func (ca *CAImpl) GetNumberOfRootCerts() int {
+	return len(ca.chains)
 }
 
 func (ca *CAImpl) getChain(no int) *chain {
-	if no == 0 {
-		return ca.mainChain
-	} else if 0 < no && no <= len(ca.alternativeChains) {
-		return ca.alternativeChains[no-1]
+	if 0 <= no && no < len(ca.chains) {
+		return ca.chains[no]
 	} else {
 		return nil
 	}
--- a/ca/ca.go
+++ b/ca/ca.go
@@ -29,6 +29,11 @@ type CAImpl struct {
 	db               *db.MemoryStore
 	ocspResponderURL string
 
+	mainChain         *chain
+	alternativeChains []*chain
+}
+
+type chain struct {
 	root         *issuer
 	intermediate *issuer
 }
@@ -58,7 +63,8 @@ func makeKey() (*rsa.PrivateKey, error) {
 	return key, nil
 }
 
-func (ca *CAImpl) makeRootCert(
+func (chain *chain) makeRootCert(
+	ca *CAImpl,
 	subjectKey crypto.Signer,
 	subjCNPrefix string,
 	signer *issuer) (*core.Certificate, error) {
@@ -114,19 +120,19 @@ func (ca *CAImpl) makeRootCert(
 	return newCert, nil
 }
 
-func (ca *CAImpl) newRootIssuer() error {
+func (chain *chain) newRootIssuer(ca *CAImpl) error {
 	// Make a root private key
 	rk, err := makeKey()
 	if err != nil {
 		return err
 	}
 	// Make a self-signed root certificate
-	rc, err := ca.makeRootCert(rk, rootCAPrefix, nil)
+	rc, err := chain.makeRootCert(ca, rk, rootCAPrefix, nil)
 	if err != nil {
 		return err
 	}
 
-	ca.root = &issuer{
+	chain.root = &issuer{
 		key:  rk,
 		cert: rc,
 	}
@@ -134,28 +140,33 @@ func (ca *CAImpl) newRootIssuer() error {
 	return nil
 }
 
-func (ca *CAImpl) newIntermediateIssuer() error {
-	if ca.root == nil {
+func (chain *chain) newIntermediateIssuer(ca *CAImpl, ik crypto.Signer) error {
+	if chain.root == nil {
 		return fmt.Errorf("newIntermediateIssuer() called before newRootIssuer()")
 	}
 
-	// Make an intermediate private key
-	ik, err := makeKey()
-	if err != nil {
-		return err
-	}
-
 	// Make an intermediate certificate with the root issuer
-	ic, err := ca.makeRootCert(ik, intermediateCAPrefix, ca.root)
+	ic, err := chain.makeRootCert(ca, ik, intermediateCAPrefix, chain.root)
 	if err != nil {
 		return err
 	}
-	ca.intermediate = &issuer{
+	chain.intermediate = &issuer{
 		key:  ik,
 		cert: ic,
 	}
 	ca.log.Printf("Generated new intermediate issuer with serial %s\n", ic.ID)
 	return nil
+}
+
+func newChain(ca *CAImpl, ik crypto.Signer) *chain {
+	chain := &chain{}
+	if err := chain.newRootIssuer(ca); err != nil {
+		panic(fmt.Sprintf("Error creating new root issuer: %s", err.Error()))
+	}
+	if err := chain.newIntermediateIssuer(ca, ik); err != nil {
+		panic(fmt.Sprintf("Error creating new intermediate issuer: %s", err.Error()))
+	}
+	return chain
 }
 
 func (ca *CAImpl) newCertificate(domains []string, ips []net.IP, key crypto.PublicKey, accountID string) (*core.Certificate, error) {
@@ -168,7 +179,7 @@ func (ca *CAImpl) newCertificate(domains []string, ips []net.IP, key crypto.Publ
 		return nil, fmt.Errorf("must specify at least one domain name or IP address")
 	}
 
-	issuer := ca.intermediate
+	issuer := ca.mainChain.intermediate
 	if issuer == nil || issuer.cert == nil {
 		return nil, fmt.Errorf("cannot sign certificate - nil issuer")
 	}
@@ -203,13 +214,19 @@ func (ca *CAImpl) newCertificate(domains []string, ips []net.IP, key crypto.Publ
 		return nil, err
 	}
 
+	altIssuers := make([]*core.Certificate, len(ca.alternativeChains))
+	for i := 0; i < len(ca.alternativeChains); i++ {
+		altIssuers[i] = ca.alternativeChains[i].intermediate.cert
+	}
+
 	hexSerial := hex.EncodeToString(cert.SerialNumber.Bytes())
 	newCert := &core.Certificate{
-		ID:        hexSerial,
-		AccountID: accountID,
-		Cert:      cert,
-		DER:       der,
-		Issuer:    issuer.cert,
+		ID:                 hexSerial,
+		AccountID:          accountID,
+		Cert:               cert,
+		DER:                der,
+		Issuer:             issuer.cert,
+		AlternativeIssuers: altIssuers,
 	}
 	_, err = ca.db.AddCertificate(newCert)
 	if err != nil {
@@ -218,7 +235,7 @@ func (ca *CAImpl) newCertificate(domains []string, ips []net.IP, key crypto.Publ
 	return newCert, nil
 }
 
-func New(log *log.Logger, db *db.MemoryStore, ocspResponderURL string) *CAImpl {
+func New(log *log.Logger, db *db.MemoryStore, ocspResponderURL string, alternateRoots int) *CAImpl {
 	ca := &CAImpl{
 		log: log,
 		db:  db,
@@ -229,13 +246,16 @@ func New(log *log.Logger, db *db.MemoryStore, ocspResponderURL string) *CAImpl {
 		ca.log.Printf("Setting OCSP responder URL for issued certificates to %q", ca.ocspResponderURL)
 	}
 
-	err := ca.newRootIssuer()
+	ik, err := makeKey()
 	if err != nil {
-		panic(fmt.Sprintf("Error creating new root issuer: %s", err.Error()))
+		panic(fmt.Sprintf("Error creating new intermediate private key: %s", err.Error()))
 	}
-	err = ca.newIntermediateIssuer()
-	if err != nil {
-		panic(fmt.Sprintf("Error creating new intermediate issuer: %s", err.Error()))
+	ca.mainChain = newChain(ca, ik)
+
+	// Create alternative chains
+	ca.alternativeChains = make([]*chain, alternateRoots)
+	for i := 0; i < len(ca.alternativeChains); i++ {
+		ca.alternativeChains[i] = newChain(ca, ik)
 	}
 	return ca
 }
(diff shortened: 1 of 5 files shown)


ca/ca.go
```diff
@@ -112,50 +118,57 @@ func (ca *CAImpl) makeRootCert(
...
 		cert: ic,
 	}
 	ca.log.Printf("Generated new intermediate issuer with serial %s\n", ic.ID)
 	return nil
 }
 
+func newChain(ca *CAImpl, ik crypto.Signer) *chain {
+	chain := &chain{}
+	err := chain.newRootIssuer(ca)
+	if err != nil {
```
cpu: nit: you can collapse this line and the prev into a single assignment & check `if` expression like:
```
if err := chain.newRootIssuer(ca); err != nil { 
...
```

ca/ca.go
```diff
@@ -112,50 +118,57 @@ func (ca *CAImpl) makeRootCert(
...
+		panic(fmt.Sprintf("Error creating new root issuer: %s", err.Error()))
+	}
+	err = chain.newIntermediateIssuer(ca, ik)
+	if err != nil {
```
cpu: ditto here: re collapsing.

ca/ca.go
```diff
@@ -112,50 +118,57 @@ func (ca *CAImpl) makeRootCert(
...
 	}
 	ca.log.Printf("Generated new intermediate issuer with serial %s\n", ic.ID)
 	return nil
 }
 
+func newChain(ca *CAImpl, ik crypto.Signer) *chain {
+	chain := &chain{}
+	err := chain.newRootIssuer(ca)
+	if err != nil {
+		panic(fmt.Sprintf("Error creating new root issuer: %s", err.Error()))
```
cpu: lately I've preferred `%v` for errors instead of `%s` and `err.Error()`. I suspect this isn't consistent in Pebble. Do you have a pref?

ca/ca.go
```diff
@@ -263,9 +285,22 @@ func (ca *CAImpl) CompleteOrder(order *core.Order) {
 	order.Unlock()
 }
 
-func (ca *CAImpl) GetRootCert() *core.Certificate {
-	if ca.root == nil {
+func (ca *CAImpl) GetRootCert(no *int) *core.Certificate {
```
cpu: It seems strange to make the `int` parameter a pointer type. Was there a specific reason you did this I'm missing?

Otherwise I think it would be better as `GetRootCert(no uint)`. That will also let you remove the `if no == nil` check. Changing the `int` to `uint` will also let you drop the `if *no < 0` check.

cmd/pebble/main.go
```diff
@@ -56,8 +57,15 @@ func main() {
 		setupCustomDNSResolver(*resolverAddress)
 	}
 
+	alternateRoots := 0
+	alternateRootsEnv := "PEBBLE_ALTERNATE_ROOTS"
+	alternateRootsVal := os.Getenv(alternateRootsEnv)
```
cpu: since `alternateRootsEnv` isn't ref'd anywhere else I'd probably remove that var and just do:
```
alternateRootsVal := os.GetEnv("PEBBLE_ALTERNATE_ROOTS")
```

core/types.go
```diff
@@ -167,14 +168,17 @@ func (c Certificate) PEM() []byte {
 	return buf.Bytes()
 }
 
-func (c Certificate) Chain() []byte {
+func (c Certificate) Chain(no *int) []byte {
```
cpu: ditto here RE `*int`

ca/ca.go
felixfontein: Thanks! Fixed.

ca/ca.go
felixfontein: Fixed.

cmd/pebble/main.go
felixfontein: Fixed as well. I originally thought that this might become a constant somewhere else, but then I didn't found a place where to collect such constants for `main.go` and left it there...

ca/ca.go
felixfontein: So far I've always copied something close to the code I was editing ;-)

Besides that: grepping for `%v` in Pebble has one match (`cmd/pebble-challtestsrv/history.go`) and some more in `vendor/`. So the broad consensus in Pebble seems to be `%s` and `err.Error()`.

ca/ca.go
felixfontein: My idea was that `nil` stands for "default chain", and non-`nil` for Alternative #`no`. But it is probably easier to understand if `no == 0` is the "default chain", and `no == 1` up to `no == PEBBLE_ALTERNATE_ROOTS` the alternative chains.

core/types.go
```diff
@@ -167,14 +168,17 @@ func (c Certificate) PEM() []byte {
...
+func (c Certificate) Chain(no int) []byte {
 	chain := make([][]byte, 0)
 
 	// Add the leaf certificate
 	chain = append(chain, c.PEM())
 
 	// Add zero or more issuers
 	issuer := c.Issuer
+	if no != 0 && 0 < no && no <= len(c.AlternativeIssuers) {
+		issuer = c.AlternativeIssuers[no - 1]
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)

ca/ca.go
```diff
@@ -263,9 +283,19 @@ func (ca *CAImpl) CompleteOrder(order *core.Order) {
...
 }
 
-func (ca *CAImpl) GetRootCert() *core.Certificate {
-	if ca.root == nil {
+func (ca *CAImpl) GetRootCert(no int) *core.Certificate {
+	var chain *chain
+	if no == 0 {
+		chain = ca.mainChain
+	} else if 0 < no && no <= len(ca.alternativeChains) {
+		chain = ca.alternativeChains[no - 1]
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)

wfe/wfe.go
```diff
@@ -227,12 +227,38 @@ func (wfe *WebFrontEndImpl) RootCert(
...
 		response.WriteHeader(http.StatusServiceUnavailable)
 		return
 	}
 
+	// Add links to alternate roots
+	if no != 0 {
+		response.Header().Add("Link", link(wfe.relativeEndpoint(request, RootCertPath), "alternate"))
+	}
+	for i := 0; i < wfe.ca.GetNumberOfAlternativeRootCerts(); i++ {
+		if no == i + 1 {
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)

wfe/wfe.go
```diff
@@ -227,12 +227,38 @@ func (wfe *WebFrontEndImpl) RootCert(
 	response http.ResponseWriter,
 	request *http.Request) {
 
-	root := wfe.ca.GetRootCert()
+	// Check for parameter
+	var no int = 0
```
golangcibot: should drop = 0 from declaration of var no; it is the zero value (from `golint`)

wfe/wfe.go
```diff
@@ -1876,9 +1902,33 @@ func (wfe *WebFrontEndImpl) Certificate(
 		return
 	}
 
+	var no int = 0
```
golangcibot: should drop = 0 from declaration of var no; it is the zero value (from `golint`)

wfe/wfe.go
```diff
@@ -1876,9 +1908,32 @@ func (wfe *WebFrontEndImpl) Certificate(
...
 	}
 
+	var no int
+	if len(serialSplit) == 2 {
+		no, err := strconv.Atoi(serialSplit[1])
+		if err != nil || no < 0 || no >= len(cert.AlternativeIssuers) {
+			response.WriteHeader(http.StatusNotFound)
+			return
+		}
+		no = no + 1
```
golangcibot: ineffectual assignment to `no` (from `ineffassign`)

ca/ca.go
cpu: Thanks for grepping :-) I should have done that myself. Using `%s` and `err.Error()` here is OK with me given its the most consistent. Thanks!

ca/ca.go
cpu: I think `no == 0` as the default chain will be more idiomatic.

felixfontein: Good, then it should be fine now :)

felixfontein:
Rebased to remove conflicts introduced by #221.

cpu:
@felixfontein Thank you! Getting this and #236 reviewed/merged are top of my Pebble list this week.

cpu:
@felixfontein Can you resolve the conflicts on this branch when you have a chance? Thanks!

felixfontein:
I'll try to work on it today.

wfe/wfe.go
```diff
@@ -232,34 +232,95 @@ func (wfe *WebFrontEndImpl) sendError(prob *acme.ProblemDetails, response http.R
 	_, _ = response.Write(problemDoc)
 }
 
+type certGetter func(no int) *core.Certificate;
+type keyGetter func(no int) *rsa.PrivateKey;
+
 func (wfe *WebFrontEndImpl) handleCert(
-	cert *core.Certificate) func(
+	certGet certGetter,
+	numberOfAlternativeRootCerts int,
```
golangcibot: `(*WebFrontEndImpl).handleCert` - `numberOfAlternativeRootCerts` is unused (from `unparam`)

wfe/wfe.go
```diff
@@ -232,34 +232,95 @@ func (wfe *WebFrontEndImpl) sendError(prob *acme.ProblemDetails, response http.R
...
 		response.Header().Set("Content-Type", "application/pem-certificate-chain; charset=utf-8")
 		response.WriteHeader(http.StatusOK)
 		_, _ = response.Write(cert.PEM())
 	}
 }
 
 func (wfe *WebFrontEndImpl) handleKey(
-	key *rsa.PrivateKey) func(
+	keyGet keyGetter,
+	numberOfAlternativeRootCerts int,
```
golangcibot: `(*WebFrontEndImpl).handleKey` - `numberOfAlternativeRootCerts` is unused (from `unparam`)

wfe/wfe.go
```diff
@@ -232,34 +232,95 @@ func (wfe *WebFrontEndImpl) sendError(prob *acme.ProblemDetails, response http.R
 	_, _ = response.Write(problemDoc)
 }
 
+type certGetter func(no int) *core.Certificate;
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)

felixfontein:
It's good you didn't yet look at wfe/wfe.go, because I had to mostly rewrite it :) I squashed all commits to make the rebase bearable...

ca/ca.go
```diff
@@ -57,7 +62,8 @@ func makeKey() (*rsa.PrivateKey, error) {
 	return key, nil
 }
 
-func (ca *CAImpl) makeRootCert(
+func (chain *chain) makeRootCert(
```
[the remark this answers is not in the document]
felixfontein: Fixed.

ca/ca.go
```diff
@@ -113,50 +119,55 @@ func (ca *CAImpl) makeRootCert(
 	return newCert, nil
 }
 
-func (ca *CAImpl) newRootIssuer() error {
+func (chain *chain) newRootIssuer(ca *CAImpl) error {
```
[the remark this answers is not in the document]
felixfontein: Fixed.

ca/ca.go
```diff
@@ -113,50 +119,55 @@ func (ca *CAImpl) makeRootCert(
...
 		key:  rk,
 		cert: rc,
 	}
 	ca.log.Printf("Generated new root issuer with serial %s\n", rc.ID)
 	return nil
 }
 
-func (ca *CAImpl) newIntermediateIssuer() error {
-	if ca.root == nil {
+func (chain *chain) newIntermediateIssuer(ca *CAImpl, ik crypto.Signer) error {
```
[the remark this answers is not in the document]
felixfontein: This one too, now.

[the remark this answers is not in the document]
felixfontein: And now completely. Sorry, got sidetracked too much by a meeting...

wfe/wfe.go
```diff
@@ -232,34 +235,107 @@ func (wfe *WebFrontEndImpl) sendError(prob *acme.ProblemDetails, response http.R
...
+	urlSplit := strings.SplitN(url, "/alternate/", 2)
+	if len(urlSplit) == 0 {
+		return url, 0, 0
+	}
+	var urlTrunc string
+	var noStr string
+	if len(urlSplit) == 1 {
+		if strings.HasPrefix(url, "alternate/") {
+			urlTrunc = ""
+			noStr = url[len("alternate/"):len(url)]
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)

felixfontein:
I've uniformized alternate handling in 12782572d90f7203e062d669327ca937b9bc2b06 a bit. For that, I needed to modify `HandleFunc` slightly, so that for paths which don't end with a slash, it is possible to accept the slashed version as well. I hope that's fine.

ca/ca.go
```diff
@@ -113,48 +119,55 @@ func (ca *CAImpl) makeRootCert(
...
-		return fmt.Errorf("newIntermediateIssuer() called before newRootIssuer()")
-	}
-
-	// Make an intermediate private key
-	ik, err := makeKey()
-	if err != nil {
-		return err
+func newIntermediateIssuer(root *issuer, ca *CAImpl, ik crypto.Signer) (*issuer, error) {
+	if root == nil {
+		return nil, fmt.Errorf("newIntermediateIssuer() called before newRootIssuer()")
```
cpu: I think this error message needs to be updated now that the `*issuer` is a parameter and not something that would be populated by calling `newRootIssuer()` before `newIntermediateIssuer()`.

ca/ca.go
```diff
@@ -113,48 +119,55 @@ func (ca *CAImpl) makeRootCert(
...
-func (ca *CAImpl) newIntermediateIssuer() error {
-	if ca.root == nil {
-		return fmt.Errorf("newIntermediateIssuer() called before newRootIssuer()")
-	}
-
-	// Make an intermediate private key
-	ik, err := makeKey()
-	if err != nil {
-		return err
+func newIntermediateIssuer(root *issuer, ca *CAImpl, ik crypto.Signer) (*issuer, error) {
```
cpu: I think that `newIntermediateIssuer` should generate its `ik crypto.Signer` itself rather than take a parameter. Since its always randomly generated in `newChain`. That also matches what `newRootIssuer` is doing and what `newIntermediateIssuer` used to do.

Sorry I missed that in my first pass.

ca/ca.go
```diff
@@ -28,6 +28,11 @@ type CAImpl struct {
 	log *log.Logger
 	db  *db.MemoryStore
 
+	mainChain         *chain
```
cpu: I think that the logic in `getChain` could be simplified if there was just one `chains []*chain` field on this struct and we documented that the default chain (I think "default" is clearer than "main") is always index 0.

ca/ca.go
```diff
@@ -267,38 +289,56 @@ func (ca *CAImpl) CompleteOrder(order *core.Order) {
 	order.Unlock()
 }
 
-func (ca *CAImpl) GetRootCert() *core.Certificate {
-	if ca.root == nil {
+func (ca *CAImpl) GetNumberOfAlternativeRootCerts() int {
+	return len(ca.alternativeChains)
```
cpu: if you switch to one slice of chains remember to adjust this to be `len(ca.chains) - 1`

core/types.go
```diff
@@ -145,11 +145,12 @@ func (ch *Challenge) ExpectedKeyAuthorization(key *jose.JSONWebKey) string {
...
-	ID        string
-	Cert      *x509.Certificate
-	DER       []byte
-	Issuer    *Certificate
-	AccountID string
+	ID                 string
+	Cert               *x509.Certificate
+	DER                []byte
+	Issuer             *Certificate
+	AlternativeIssuers []*Certificate
```
cpu: Ditto my advice here about the chain fields. I think just one field of `[]*Certificate` would be simpler, with the default issuer certificate being index 0.

wfe/wfe.go
```diff
@@ -232,34 +235,107 @@ func (wfe *WebFrontEndImpl) sendError(prob *acme.ProblemDetails, response http.R
 	_, _ = response.Write(problemDoc)
 }
 
+func getAlternateNo(url string) (string, int, int) {
```
cpu: I think this function is a little bit sparse for comments. I'm not entirely sure I follow the URL parsing logic as-is or what the functions return values are.

wfe/wfe.go
```diff
@@ -232,34 +235,107 @@ func (wfe *WebFrontEndImpl) sendError(prob *acme.ProblemDetails, response http.R
...
+		noStr = urlSplit[1]
+	}
+	no, err := strconv.Atoi(noStr)
+	if err != nil || no < 0 {
+		return url, 0, http.StatusNotFound
+	}
+	return urlTrunc, no + 1, 0
+}
+
+func addAlternateLinks(response http.ResponseWriter, url string, no int, numberOfAlternates int) {
```
cpu: ditto this function for comments

ca/ca.go
felixfontein: Fixed.

ca/ca.go
felixfontein: Actually, all intermediate certs must use the same private key, otherwise they aren't all valid intermediates for the leaf certificate. Maybe that's why you missed it in your first pass :)

ca/ca.go
felixfontein: I simplified it here, and also in the `Issuer` / `AlternativeIssuers` of `Certificate`.

ca/ca.go
felixfontein: I renamed the function to `getNumberOfRootCerts` instead.

core/types.go
felixfontein: Ah, I overlooked this apparently when I wrote the comment above; this is also changed :)

wfe/wfe.go
felixfontein: I've added comments. I hope it's clearer now.

wfe/wfe.go
felixfontein: Same here.

ca/ca.go
cpu: :man_facepalming: Good point! That makes sense :-)

wfe/wfe.go
```diff
@@ -232,34 +235,127 @@ func (wfe *WebFrontEndImpl) sendError(prob *acme.ProblemDetails, response http.R
 	_, _ = response.Write(problemDoc)
 }
 
+func getAlternateNo(url string) (string, int, int) {
```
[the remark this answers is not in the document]
felixfontein: The difference is that there are URLs like `/certZ/0123456789/alternative/1` and `/root/alternative/1`. The handler gets `0123456789/alternative/1` resp. `alternative/1`. In the first case, we have to return everything before `/alternative/`, and in the second, there's no such string and we have to return `""`.

ca/ca.go
```diff
@@ -267,38 +285,54 @@ func (ca *CAImpl) CompleteOrder(order *core.Order) {
...
-func (ca *CAImpl) GetRootCert() *core.Certificate {
-	if ca.root == nil {
+func (ca *CAImpl) GetNumberOfRootCerts() int {
+	return len(ca.chains)
+}
+
+func (ca *CAImpl) getChain(no int) *chain {
+	if 0 <= no && no < len(ca.chains) {
+		return ca.chains[no]
+	} else {
```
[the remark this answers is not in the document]
felixfontein: Done.

wfe/wfe.go
```diff
@@ -232,34 +235,127 @@ func (wfe *WebFrontEndImpl) sendError(prob *acme.ProblemDetails, response http.R
...
+	// Parse the URL to extract alternate number (if available, default 0).
```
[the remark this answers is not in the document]
felixfontein: I moved it (and the same for `addAlternateLinks`).

wfe/wfe.go
```diff
@@ -232,34 +235,127 @@ func (wfe *WebFrontEndImpl) sendError(prob *acme.ProblemDetails, response http.R
...
+	// Returns the remaining URL, the number (0 or larger), and a HTTP
+	// error code (or 0 for success).
```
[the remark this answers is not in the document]

wfe/wfe.go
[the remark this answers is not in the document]
felixfontein: What we *could* do is call the function with `getAlternateNo("/" + request.URL.Path)` from `handleCert` / `handleKey`, then we can stick to the `/alternative/` case. Should I change it like this?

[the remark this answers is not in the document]
cpu: > What we could do is call the function with getAlternateNo("/" + request.URL.Path) from handleCert / handleKey, then we can stick to the /alternative/ case. Should I change it like this?

That seems reasonable to me :+1:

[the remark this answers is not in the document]
felixfontein: I've pushed the change. I can't test it right now, my laptop's HDD died and I haven't set up a development system again...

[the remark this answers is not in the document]
cpu: Eek! That's no good. Sorry to hear :-(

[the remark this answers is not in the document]
cpu: I'm working on testing now.

felixfontein:
Hmm, that looks really strange. Especially since the code looks OK, and the extra newline between the two chains looks very suspicious. I have absolutely no idea where it came from...

cpu:
@felixfontein :man_facepalming: It's my own tool. I forgot it was echoing just the chain, and also printing the HTTP response, so it appeared doubled :laughing: Sorry about the false alarm!

I'm almost done my testing.

felixfontein:
LOL ok! That explains it :D

cpu:
@jsha I think this branch is in a state where you could give it a second :mag: now.

wfe/wfe.go
```diff
@@ -280,25 +360,41 @@ func (wfe *WebFrontEndImpl) handleKey(
...
-	wfe.HandleFunc(m, RootCertPath, wfe.handleCert(wfe.ca.GetRootCert()), "GET")
-	wfe.HandleFunc(m, rootKeyPath, wfe.handleKey(wfe.ca.GetRootKey()), "GET")
-	wfe.HandleFunc(m, intermediateCertPath, wfe.handleCert(wfe.ca.GetIntermediateCert()), "GET")
-	wfe.HandleFunc(m, intermediateKeyPath, wfe.handleKey(wfe.ca.GetIntermediateKey()), "GET")
+	wfe.HandleFunc(m, noncePath, wfe.Nonce, false, "GET")
+	wfe.HandleFunc(m, RootCertPath, wfe.handleCert(
+		func(no int) *core.Certificate {
+			return wfe.ca.GetRootCert(no)
+		},
+		wfe.ca.GetNumberOfRootCerts(), RootCertPath), true, "GET")
```
jsha: Plumbing through the value of `GetNumberOfRootCerts` all the way to `handleCert` seems to add a lot of complexity. As far as I can tell it's only used to return 404 when a too-high number is requested. It would be simpler to rely on the fact that `GetRootCert` returns nil if the parameter is too high or too low. Then you don't need to define `GetNumberOfRootCerts` nor pass it as a parameter to `handleCert`.

wfe/wfe.go
```diff
@@ -280,25 +360,41 @@ func (wfe *WebFrontEndImpl) handleKey(
...
+	wfe.HandleFunc(m, DirectoryPath, wfe.Directory, false, "GET")
 	// Note for noncePath: "GET" also implies "HEAD"
-	wfe.HandleFunc(m, noncePath, wfe.Nonce, "GET")
-	wfe.HandleFunc(m, RootCertPath, wfe.handleCert(wfe.ca.GetRootCert()), "GET")
-	wfe.HandleFunc(m, rootKeyPath, wfe.handleKey(wfe.ca.GetRootKey()), "GET")
-	wfe.HandleFunc(m, intermediateCertPath, wfe.handleCert(wfe.ca.GetIntermediateCert()), "GET")
-	wfe.HandleFunc(m, intermediateKeyPath, wfe.handleKey(wfe.ca.GetIntermediateKey()), "GET")
+	wfe.HandleFunc(m, noncePath, wfe.Nonce, false, "GET")
+	wfe.HandleFunc(m, RootCertPath, wfe.handleCert(
+		func(no int) *core.Certificate {
```
jsha: You don't have to define a local function here, you can just write:

```
wfe.HandleFunc(m, RootCertPath, wfe.handleCert(wfe.ca.GetRootCert, ...
```

This is an unusual feature of Go called [method expressions](https://golang.org/ref/spec#Method_expressions).

I believe the same applies to the funcs you define in the other HandleFuncs below.

wfe/wfe.go
```diff
@@ -169,6 +169,7 @@ func (wfe *WebFrontEndImpl) HandleFunc(
 	mux *http.ServeMux,
 	pattern string,
 	handler wfeHandlerFunc,
+	includeTrailingSlash bool,
```
jsha: I'm generally wary of adding too many parameters to functions, especially boolean parameters (because it's hard to figure out when looking at the call site what the boolean does).

I think in this case, rather than adding `includeTrailingSlash`, you can just change the `pattern` parameter. So, for instance, you would change `RootCertPath` from `/root` to `/root/`. That should allow you to revert these changes to `HandleFunc` and simplify this PR.

wfe/wfe.go
ffdybuster: The reason I did this is because there are two distinct errors: 404 when a too-high (or invalid) number is requested, and 503 when the number is valid, but the certificate returned by the CA is `nil`. I assumed that this was because eventually, initialization of CA and WFE could be done in parallel, and then this would mean that the CA isn't fully initialized yet. If this isn't planned, returning always 404 and simplifying the code by not passing the value of `GetNumberOfRootCerts` around would be definitely easier.

wfe/wfe.go
ffdybuster: Ah, so Go behaves similarly to Java 8+ where this is also possible. Neat. Makes it a lot easier :)

wfe/wfe.go
ffdybuster: The disadvantage is that current users of Pebble which rely that the root can be get via `/root` will suddenly no longer work, since `/root` will yield 404 and `/root/` must be used. (Same for the new private key / intermediate certificate endpoints.) If that's not a concern, I don't mind changing this (as it indeed would simplify everything a lot).

ca/ca.go
```diff
@@ -57,7 +61,8 @@ func makeKey() (*rsa.PrivateKey, error) {
 	return key, nil
 }
 
-func (ca *CAImpl) makeRootCert(
+func makeRootCert(
```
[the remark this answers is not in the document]
ffdybuster: The current version is the result after two steps of refactoring, that's why it ended up as a function. But yes, it definitely makes sense to convert it back to be a method of `CAImpl`.

wfe/wfe.go
jsha: I can confirm we don't plan to initialize CA and WFE in parallel. Always returning 404 makes a lot of sense, thanks!

wfe/wfe.go
jsha: Good point about the existing clients that hardcode `/root`. How about this: make the new path `/roots/0` `/roots/1`, etc, then add a redirect from `/root` to `/roots/1`.

The addition of the `s` works around the fact that `net/http`'s ServeMux automatically redirects `foo` to `foo/` if `foo/` is registered as a subtree.

felixfontein:
Rebased to resolve conflicts with #242.

ca/ca.go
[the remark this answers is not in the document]
felixfontein: Fixed.

wfe/wfe.go

wfe/wfe.go
felixfontein: Just noticed, the value of `GetNumberOfRootCerts` is still needed for calling `addAlternateLinks`, since it needs to know how many alternatives there are.

wfe/wfe.go
jsha: Thanks for the fixes so far! Waiting on an update to this feedback before doing my next round of review.

jsha: Oops, I see you just pushed some more changes, including these! Ready for re-review?

felixfontein: I've implemented this almost, what I've left is that's there `/roots/`, `/roots/alternate/1`, etc. I could change this, but then `handleCert`/`handleKey` would work differently than `Certificate`, in particular adding the link headers.

felixfontein: Sorry, I'm still in the process of restoring my main dev machine from backups, so I'm a bit slow today :) Should be ready for re-review now!

jsha: No problem, I think your turnaround is very fast! I just wanted to communicate clearly so we didn't both wind up waiting on each other. :-)

felixfontein:
I think it is now working as expected (got my machine back running and actually was able to properly test it :) ), and it should also address your latest review comment.

felixfontein:
@cpu @jsha thanks a lot for all your feedback! :)

PR #236, opened 2019-05-15
Expose issuer root key, intermediate issuer cert and key through entrypoints

adferrand:
Since Pebble generates the issuer root and intermediate certificates on-the-fly at each startup, it can be useful for test purpose to get their current private key and certificate themselves.

The specific use case I have is a project of external OCSP mock server to run alongside with Pebble. 

The issuer root certificate is already exposed with `/root`, this PR adds the ad-hoc `/root-key`, `/intermediate` and `/intermediate-key` endpoints to dump them in PEM format.

### DIFF
--- a/README.md
+++ b/README.md
@@ -279,5 +279,5 @@ respectively.
 store that you use for ordinary browsing or that is used for non-testing
 purposes, since Pebble and its generated keys are not audited or held to the
 same standards as the Let's Encrypt production CA and their keys, are exposed
-by pebble and so are not safe to use for anything other than testing. Also,
+by Pebble and so are not safe to use for anything other than testing. Also,
 their private keys will be lost as soon as the Pebble process terminates.**
--- a/README.md
+++ b/README.md
@@ -271,7 +271,7 @@ and `https://localhost:14000/intermediate` respectively.
 You might need the root certificate to verify the complete trust chain of
 generated certificates, for example in end-to-end tests.
 
-The private keys of these certificates can also be retreived by a `GET` request
+The private keys of these certificates can also be retrieved by a `GET` request
 to `https://localhost:14000/root-key` and `https://localhost:14000/intermediate-key`
 respectively.
 
--- a/README.md
+++ b/README.md
@@ -278,6 +278,6 @@ respectively.
 **IMPORTANT: Do not add Pebble's root or intermediate certificate to a trust
 store that you use for ordinary browsing or that is used for non-testing
 purposes, since Pebble and its generated keys are not audited or held to the
-same standards as the Let's Encrypt production CA and their keys, are expose
+same standards as the Let's Encrypt production CA and their keys, are exposed
 by pebble and so are not safe to use for anything other than testing. Also,
 their private keys will be lost as soon as the Pebble process terminates.**
--- a/README.md
+++ b/README.md
@@ -278,6 +278,6 @@ respectively.
 **IMPORTANT: Do not add Pebble's root or intermediate certificate to a trust
 store that you use for ordinary browsing or that is used for non-testing
 purposes, since Pebble and its generated keys are not audited or held to the
-same standards as the Let's Encrypt production CA and their keys, are exposed
-by Pebble and so are not safe to use for anything other than testing. Also,
-their private keys will be lost as soon as the Pebble process terminates.**
+same standards as the Let's Encrypt production CA and their keys. Moreover
+these keys are exposed by Pebble and will be lost as soon as the process
+terminates: so they are not safe to use for anything other than testing.**
--- a/README.md
+++ b/README.md
@@ -262,17 +262,22 @@ store or to any production systems/codebases. The private key for this CA is
 intentionally made [publicly available in this
 repo](test/certs/pebble.minica.key.pem).**
 
-### CA Root Certificate
+### CA Root and Intermediate Certificates
 
-Note that the CA's root certificate is regenerated on every launch. It can be
-retrieved by a `GET` request to `https://localhost:14000/root`.
+Note that the CA's root and intermediate certificates are regenerated on every
+launch. It can be retrieved by a `GET` request to `https://localhost:14000/root`
+and `https://localhost:14000/intermediate` respectively.
 
 You might need the root certificate to verify the complete trust chain of
 generated certificates, for example in end-to-end tests.
 
+The private keys of these certificates can also be retreived by a `GET` request
+to `https://localhost:14000/root-key` and `https://localhost:14000/intermediate-key`
+respectively.
+
 **IMPORTANT: Do not add Pebble's root or intermediate certificate to a trust
 store that you use for ordinary browsing or that is used for non-testing
 purposes, since Pebble and its generated keys are not audited or held to the
-same standards as the Let's Encrypt production CA and their keys, and so are
-not safe to use for anything other than testing. Also, their private keys
-will be lost as soon as the Pebble process terminates.**
+same standards as the Let's Encrypt production CA and their keys, are expose
+by pebble and so are not safe to use for anything other than testing. Also,
+their private keys will be lost as soon as the Pebble process terminates.**
--- a/ca/ca.go
+++ b/ca/ca.go
@@ -73,7 +73,7 @@ func (ca *CAImpl) makeRootCert(
 		KeyUsage:              x509.KeyUsageDigitalSignature | x509.KeyUsageCertSign,
 		ExtKeyUsage:           []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth, x509.ExtKeyUsageClientAuth},
 		BasicConstraintsValid: true,
-		IsCA: true,
+		IsCA:                  true,
 	}
 
 	var signerKey crypto.Signer
@@ -182,7 +182,7 @@ func (ca *CAImpl) newCertificate(domains []string, key crypto.PublicKey, account
 		KeyUsage:              x509.KeyUsageDigitalSignature | x509.KeyUsageKeyEncipherment,
 		ExtKeyUsage:           []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth, x509.ExtKeyUsageClientAuth},
 		BasicConstraintsValid: true,
-		IsCA: false,
+		IsCA:                  false,
 	}
 	der, err := x509.CreateCertificate(rand.Reader, template, issuer.cert.Cert, key, issuer.key)
 	if err != nil {
--- a/ca/ca.go
+++ b/ca/ca.go
@@ -74,7 +74,7 @@ func (ca *CAImpl) makeRootCert(
 		KeyUsage:              x509.KeyUsageDigitalSignature | x509.KeyUsageCertSign,
 		ExtKeyUsage:           []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth, x509.ExtKeyUsageClientAuth},
 		BasicConstraintsValid: true,
-		IsCA: true,
+		IsCA:                  true,
 	}
 
 	var signerKey crypto.Signer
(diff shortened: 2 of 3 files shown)


ca/ca.go
```diff
@@ -73,7 +74,7 @@ func (ca *CAImpl) makeRootCert(
 		KeyUsage:              x509.KeyUsageDigitalSignature | x509.KeyUsageCertSign,
 		ExtKeyUsage:           []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth, x509.ExtKeyUsageClientAuth},
 		BasicConstraintsValid: true,
-		IsCA:                  true,
+		IsCA: true,
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)

ca/ca.go
```diff
@@ -182,7 +185,7 @@ func (ca *CAImpl) newCertificate(domains []string, key crypto.PublicKey, account
 		KeyUsage:              x509.KeyUsageDigitalSignature | x509.KeyUsageKeyEncipherment,
 		ExtKeyUsage:           []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth, x509.ExtKeyUsageClientAuth},
 		BasicConstraintsValid: true,
-		IsCA:                  false,
+		IsCA: false,
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)

README.md
```diff
@@ -262,17 +262,22 @@ store or to any production systems/codebases. The private key for this CA is
...
+to `https://localhost:14000/root-key` and `https://localhost:14000/intermediate-key`
+respectively.
+
 **IMPORTANT: Do not add Pebble's root or intermediate certificate to a trust
 store that you use for ordinary browsing or that is used for non-testing
 purposes, since Pebble and its generated keys are not audited or held to the
-same standards as the Let's Encrypt production CA and their keys, and so are
-not safe to use for anything other than testing. Also, their private keys
-will be lost as soon as the Pebble process terminates.**
+same standards as the Let's Encrypt production CA and their keys, are expose
```
cpu:
```suggestion
same standards as the Let's Encrypt production CA and their keys, are exposed
```

README.md
```diff
@@ -262,17 +262,22 @@ store or to any production systems/codebases. The private key for this CA is
...
+by pebble and so are not safe to use for anything other than testing. Also,
```
cpu:
```suggestion
by Pebble and so are not safe to use for anything other than testing. Also,
```

ca/ca.go
```diff
@@ -32,8 +32,9 @@ type CAImpl struct {
 }
 
 type issuer struct {
-	key  crypto.Signer
-	cert *core.Certificate
+	key    crypto.Signer
+	keyRaw *rsa.PrivateKey
```
cpu: I think you can get away without having this `keyRaw` field by making a [type switch](https://github.com/golang/go/wiki/Switch#type-switch) on the `key` field in the location you'd be calling `GetIntermediateKey`, looking for a `rsa.PrivateKey` type (presently the only kind of key that Pebble will generate).

ca/ca.go
```diff
@@ -269,3 +272,26 @@ func (ca *CAImpl) GetRootCert() *core.Certificate {
 	}
 	return ca.root.cert
 }
+
+func (ca *CAImpl) GetRootKey() *rsa.PrivateKey {
```
cpu: I would change this function and the corresponding `GetIntermediateKey` to return the interface type (`crypto.Signer`) and then use the type switch I mentioned earlier in the caller. That will leave the CA more general and make it easier to support different key types in the future.

wfe/wfe.go
```diff
@@ -49,6 +51,12 @@ const (
 	revokeCertPath    = "/revoke-cert"
 	keyRolloverPath   = "/rollover-account-key"
 
+	// Theses entrypoint is obviously not a part of the standard ACME endpoints,
```
cpu:
```suggestion
	// Theses entrypoints are not a part of the standard ACME endpoints,
```

wfe/wfe.go
```diff
@@ -49,6 +51,12 @@ const (
...
+	// and are expose by Pebble as an integration tests tool.
```
cpu:
```suggestion
	// and are exposed by Pebble as an integration test tool.
```

wfe/wfe.go
```diff
@@ -238,13 +246,86 @@ func (wfe *WebFrontEndImpl) RootCert(
...
+	}
+
+	var buf bytes.Buffer
+
+	err := pem.Encode(&buf, &pem.Block{
+		Type:  "RSA PRIVATE KEY",
+		Bytes: x509.MarshalPKCS1PrivateKey(key),
+	})
+	if err != nil {
+		panic(fmt.Sprintf("Unable to encode private key to PEM: %s",
```
cpu: I think it might be better to replace the `panic` with `wfe.sendError(acme.InternalErrorProblem(msg)); return`. That's the way we've handled other "shouldn't happen" conditions (e.g. see `Directory` when `relativeDirectory` returns an err).

cpu:
One other thing: It might be sensible to try and rebase this on top of https://github.com/letsencrypt/pebble/pull/234 early. I suspect there's some overlap between your work and @felixfontein's and it will need to be squared away.

ca/ca.go
adferrand: In fact I tried the type checking, but the compiler refuse it because PrivateKey is not implementing Signer. The doc says that this interface is designed to be opaque, so I do not know how to go through that.

ca/ca.go
adferrand: See above.

ca/ca.go
adferrand: Ok nevermind. I was trying to cast to the interface crypto.PrivateKey instead of the actual implementation rsa.PrivateKey. The type switch is working now.

wfe/wfe.go
adferrand: Done.

adferrand:
@cpu for the comment about #234, I think you should merge first my PR, since it is simpler and it seems that we are almost done with the review.

Then, on #234, we should:
1) Abstracting out the logic to choose the correct chain as a method in `CAImpl`
2) Adding the three methods introduced by my PR `GetRootKey`, `GetIntermediateCert` and `GetIntermediateKey`
3) For these three methods plus `GetRootCert`, use the logic in 1) to do something like:
```
chain := ca.selectChain(0)
return chain.root.cert // or chain.intermediate.cert
```
or
```
chain := ca.selectChain(0)
switch key := chain.ca.root.key.(type) {
case *rsa.PrivateKey:
	return key
}
```
... and so on for the intermediate certificate.

adferrand:
Can't help but hate to repeat myself. So I abstracted the logic to avoid to write 4 times the exact same method.

README.md
```diff
@@ -262,17 +262,22 @@ store or to any production systems/codebases. The private key for this CA is
...
+respectively.
+
 **IMPORTANT: Do not add Pebble's root or intermediate certificate to a trust
 store that you use for ordinary browsing or that is used for non-testing
 purposes, since Pebble and its generated keys are not audited or held to the
-same standards as the Let's Encrypt production CA and their keys, and so are
-not safe to use for anything other than testing. Also, their private keys
-will be lost as soon as the Pebble process terminates.**
+same standards as the Let's Encrypt production CA and their keys, are exposed
+by Pebble and so are not safe to use for anything other than testing. Also,
```
felixfontein: I think this somewhat breaks the whole sentence, which already was pretty long before. I think it would be better to stop the sentence after "CA and their keys" and start a new one. Or keep the old sentence and just modify the next one (or add another one).

README.md
```diff
@@ -262,17 +262,22 @@ store or to any production systems/codebases. The private key for this CA is
...
-Note that the CA's root certificate is regenerated on every launch. It can be
-retrieved by a `GET` request to `https://localhost:14000/root`.
+Note that the CA's root and intermediate certificates are regenerated on every
+launch. It can be retrieved by a `GET` request to `https://localhost:14000/root`
+and `https://localhost:14000/intermediate` respectively.
 
 You might need the root certificate to verify the complete trust chain of
 generated certificates, for example in end-to-end tests.
 
+The private keys of these certificates can also be retreived by a `GET` request
```
felixfontein:
```suggestion
The private keys of these certificates can also be retrieved by a `GET` request
```

wfe/wfe.go
```diff
@@ -244,7 +283,10 @@ func (wfe *WebFrontEndImpl) Handler() http.Handler {
 	wfe.HandleFunc(m, DirectoryPath, wfe.Directory, "GET")
 	// Note for noncePath: "GET" also implies "HEAD"
 	wfe.HandleFunc(m, noncePath, wfe.Nonce, "GET")
-	wfe.HandleFunc(m, RootCertPath, wfe.RootCert, "GET")
+	wfe.HandleFunc(m, rootCertPath, wfe.handleCert(wfe.ca.GetRootCert()), "GET")
```
golangcibot: undeclared name: `rootCertPath` (from `typecheck`)

wfe/wfe.go
```diff
@@ -244,7 +283,10 @@ func (wfe *WebFrontEndImpl) Handler() http.Handler {
...
+	wfe.HandleFunc(m, RootKeyPath, wfe.handleKey(wfe.ca.GetRootKey()), "GET")
```
golangcibot: undeclared name: `RootKeyPath` (from `typecheck`)

wfe/wfe.go
```diff
@@ -49,6 +51,14 @@ const (
 	revokeCertPath    = "/revoke-cert"
 	keyRolloverPath   = "/rollover-account-key"
 
+	// Theses entrypoints are not a part of the standard ACME endpoints,
+	// and are exposed by Pebble as an integration test tool.
+	intermediateKeyPath  = "/intermediate-key"
+	intermediateCertPath = "/intermediate"
+	rootKeyPath          = "/root-key"
+        // We export RootCertPath so that the pebble binary can reference it
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)

README.md
adferrand: Indeed, it is useless to export them. And I grouped the RootCertPath with them, as it is also not an acme entrypoint.

wfe/wfe.go
```diff
@@ -49,6 +52,12 @@ const (
 	revokeCertPath    = "/revoke-cert"
 	keyRolloverPath   = "/rollover-account-key"
 
+	// Theses entrypoints are not a part of the standard ACME endpoints,
+	// and are exposed by Pebble as an integration test tool.
+	IntermediateCertPath = "/intermediate"
+	IntermediateKeyPath  = "/intermediate-key"
+	RootKeyPath          = "/root-key"
```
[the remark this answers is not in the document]
adferrand: Indeed, it is useless to export them. And I grouped the RootCertPath with them, as it is also not an acme entrypoint.

wfe/wfe.go
```diff
@@ -49,6 +51,14 @@ const (
 	revokeCertPath    = "/revoke-cert"
 	keyRolloverPath   = "/rollover-account-key"
 
+	// Theses entrypoints are not a part of the standard ACME endpoints,
+	// and are exposed by Pebble as an integration test tool. We export
+        // RootCertPath so that the pebble binary can reference it.
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)

README.md
adferrand: Alright. I rephrased this part of the README.

cpu:
> @cpu for the comment about #234, I think you should merge first my PR, since it is simpler and it seems that we are almost done with the review.

@adferrand That sounds reasonable to me :+1:

PR #237, opened 2019-06-03
Bugfix for IP address support: uniqueIPs() result contains nil as first element(s)

felixfontein:
If `uniqueIPs()` is called with a non-empty list (of `n` IP addresses), the resulting list will be preceeded by `n` `nil` values. This leads to an error result: `{'type': 'urn:ietf:params:acme:error:malformed', 'detail': 'Order included malformed IP type identifier value: \"<nil>\"\\n', 'status': 400}`


orangepizza:
Opps. I shouldn't patch around when I don't have right client for test it
.. and I should've added the appveyor.yml for ip cert test.

felixfontein:
Thanks for reviewing and merging!

PR #238, opened 2019-06-04
Fix dns-01 validation, improve DNS error logging

felixfontein:
Fixes another bug in #221 which makes dns-01 validation no longer work. The problem is that to create `_acme-challenge.xxx`, `task.Identifier` was used for `xxx` instead of `task.Identifier.Value` (`task.Identifier` used to be a string, now it is a `Identifier` object).

This also improves DNS validation error logging a bit by including the DNS error message. This helped me finding this...

### DIFF
--- a/va/va.go
+++ b/va/va.go
@@ -304,7 +304,7 @@ func (va VAImpl) validateDNS01(task *vaTask) *core.ValidationRecord {
 
 	txts, err := net.DefaultResolver.LookupTXT(ctx, challengeSubdomain)
 	if err != nil {
-		result.Error = acme.UnauthorizedProblem("Error retrieving TXT records for DNS challenge")
+		result.Error = acme.UnauthorizedProblem(fmt.Sprintf("Error retrieving TXT records for DNS challenge (%q)", err))
 		return result
 	}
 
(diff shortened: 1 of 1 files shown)


felixfontein:
With this (and thus also with #239), all my validation tests run through (HTTP, DNS and TLS-ALPN validation for DNS names and IPs where it makes sense).

orangepizza:
I think your PR is better, as mine is just reflex fix for the bug itself. but I wonder if the error message returned to client better to be in sync with boulder, as clients may use the error messages to react when the see the error? but I don't think it's hard rule.

felixfontein:
Actually, I think it's better if Pebble's error `detail` messages are different from Boulder's messages, to avoid people relying on the `detail` string (which is implementation specific and should not be relied on at all).

@cpu @jsha any opinions on this?

cpu:
> Actually, I think it's better if Pebble's error detail messages are different from Boulder's messages, to avoid people relying on the detail string (which is implementation specific and should not be relied on at all).

@felixfontein Agreed. I don't want to encourage matching detail messages since it's almost certain that as other CAs implement RFC 8555 they won't match Boulder's exact messages. It's better if Pebble differs to make that point clear as soon as possible.

felixfontein:
Thanks everyone!

PR #239, opened 2019-06-04
Fix dns challenge to use updated Identifier struct

orangepizza:
Fix DNS challenge to match with updated Identifier
task.Identifier >> task.Identifier.Value


felixfontein:
Looks like we both found the same fix ;) (see #238)

I'd prefer my PR since it improves the error message. Or should that error message better be printed to the log? If you prefer that, feel free to add that to your PR, then @cpu and @jsha can decide what they prefer...

orangepizza:
I think your PR is better, as mine is just reflex fix for the bug itself. but I wonder if the error message returned to client needs to sync with boulder, as clients may use the error messages to react when the see it?

PR #240, opened 2019-06-11
Publish release assets in GitHub

adferrand:
You can see in several GitHub projects (in particular for Go-base ones) that the releases contain several executables for each target plateform, ready to be download and used as is. See for instance: https://github.com/containous/traefik/releases/tag/v1.7.12

These assets are even used in the standard workflow of a production-grade installation, as for golintci-lint for instance: https://install.goreleaser.com/github.com/golangci/golangci-lint.sh.

This PR proposes to setup automated deployment of executable assets to Pebble GitHub releases.

Indeed, Pebble is also a Go project, that produces two executables (pebble and challtestsrv) that can be used out-of-the-box on the target platform, thanks to the Go embedding techniques when compiling executables.

To do so, this PR configure Travis CI and AppVeyor to take advantage of their GitHub release deploy provider. It allows to setup upon a tag build the proper creation of a GitHub release draft (if it does not exist yet), then create/update the relevant pebble/challtestsrv executables to this release.

Note that Travis handles the Linux executables, and AppVeyor the Windows ones. In theory, thanks to the capability of Go to cross-compile, one CI system could do all the job. However here, I ensure that the platform-specific assets are not pushed to the release before they have been fully tested in their native platform. I think this is desirable and so I setup the workflow accordingly.

Once done, upon a tag push to GitHub (like v2.0.3), the two CI will trigger the tag build, and deploy the artifacts at the end. You can see here an example of the workflow result on my fork of Pebble.

image

Note again that the release is set as a Draft, and so do not become public before a manual action is done, like editing the release body to push a changelog. Note also that the two CIs works independently from each other, and reuse the existing draft release if exists, without trying to recreate the release (that would drop the assets pushed by the other CI).

And finally, after proper editions of the release body, and its publication: https://github.com/adferrand/pebble/releases/tag/v2.0.3

To make this work, Travis and AppVeyor requires push capabilities to GitHub. So, using a machine account that has write access to `letsencrypt/pebble` GitHub project:
* on GitHub, generate a new personal token, with the `public_repo` scope
* on Travis CI, create a secured environment variable `GITHUB_AUTH_TOKEN` in the interface with the GitHub token as the content: https://travis-ci.com/letsencrypt/pebble/settings
* on AppVeyor, do the same than on Travis CI: https://ci.appveyor.com/project/letsencrypt/pebble/settings/environment

Then the setup is ready.


cpu:
In the past we've wanted to avoid supporting binary releases when the Docker images & self-compilation seem to meet most user requirements.

@jsha Is that still your preference?

adferrand:
I would like to elaborate a little more about why I think adding binaries to releases would be really helpful.

So Boulder deployment is built on top of Docker, and it makes a lot of sense. Indeed, there is several components to orchestrate together, and as a deployment tool, making a docker-compose is a quick win.

However, Pebble is a more simple product, with only one component. So basically, thanks to Go embedding capabilities, one can start Pebble with only one command. In this specific context, even if Docker offers containerization, and build from Go allows to start from sources, I think it simplifies a lot the deployment of Pebble as an integration tool.

With this, you do not need either to install Docker, or to setup a Go dev environment. You just grab the executable and start it.

Now, more specifically on Certbot, we have the integration tests, that are now fully packaged in a tox target using `tox -e integration`. As it is indicated as a standard test workflow in https://certbot.eff.org/docs/contributing.html#integration-testing-with-the-pebble-ca, we tried to make executing integration tests locally with Pebble a very suitable part of the tests a developer could do, like the unit tests.

However currently, this still requires Docker and docker-compose to be installed. Usually, this needs a not so obvious step to install it, following instructions here https://docs.docker.com/install/. So a not so obvious step.

Removing this dependency to Docker would need to build then start Pebble from source. So it needs a Go environment setup. Here, it becomes unclear for the developer: why would he need to setup Go to execute integration tests on a Python project ?

But all of this is still doable. And it is possible to setup the CI correctly, at least.

Now, comes Windows. We want to execute integration tests against this platform. And using Docker becomes a blocker. I tried several ways to use the Docker approach here, but the network capabilities from Windows are just not sufficient: https://docs.microsoft.com/en-us/virtualization/windowscontainers/container-networking/network-drivers-topologies. The more "versatile" driver (nat) would require a very complex setup on route tables to makes all things work using Docker. Where starting the components locally would be straight-forward (for more details, do not hesitate to ask me directly).

So from the general usecase, then the specific certbot one, then the windows specific certbot one, I think it would be really usefull.

Note also that in term of complexity, once this PR is merged and environment variables are setup, all things will be done automatically, the documentation for both CIs is complete, so I think the overhead is low.

adferrand:
Oh, forgot to say that installing Docker on Windows is really heavy, and but a very big constraint on the runtime: it needs HyperV to be enabled, that forbid the usage of any other virtualization method (like Virtualbox).

jsha:
I think @adferrand does a pretty good job of motivating why binary releases solve a significant problem that Docker and/or self-builds don't. @adferrand are you willing to help maintain the code that provides these binary releases? If so, I think it's a reasonable thing to do.

adferrand:
Yes @jsha, I am willing to support this logic. In general it is also the case for the parts of the CI workflow I modified, in particular the Windows part with AppVeyor.

jsha:
Thanks for offering to support this! I noticed one other thing:

> on GitHub, generate a new personal token, with the public_repo scope
> on Travis CI, create a secured environment variable GITHUB_AUTH_TOKEN in the interface with the GitHub token as the content: https://travis-ci.com/letsencrypt/pebble/settings

We'll need to create a new GitHub user with only access to the Pebble repo and get a personal auth token for that user. Personal auth tokens have full privilege, including pushing to repositories (like Boulder), so I want to be very cautious with them.

adferrand:
Indeed! This is also recommended by Travis in its documentation https://docs.travis-ci.com/user/deployment/releases/#authenticating-with-an-oauth-token, and can be applied to AppVeyor as well.

.travis.yml
```diff
@@ -56,10 +56,25 @@ script:
   # step is performed just to shake out data races with concurrent requests.
   - load-generator -config ./test/config/load-generator-config.json > /dev/null
 
+before_deploy:
+  - mkdir -p deploy
+  - cp $(go env GOPATH)/bin/pebble deploy/pebble_linux-amd64
+  - cp $(go env GOPATH)/bin/pebble-challtestsrv deploy/challtestsrv_linux-amd64
```
cpu: @adferrand I'd prefer to keep the `pebble-` prefix on the `challtestsrv` artifact. Did you have an explicit reason for dropping it that I should consider?

appveyor.yml
```diff
@@ -24,6 +31,10 @@ build_script:
 
 after_build:
   - ps: $PebbleProcess = Start-Process pebble -PassThru
+  - mkdir deploy
+  # Two following lines are copying built executable to the proper artifacts folder
+  - copy %USERPROFILE%\go\bin\pebble.exe deploy\pebble_windows-amd64.exe
+  - copy %USERPROFILE%\go\bin\pebble-challtestsrv.exe deploy\challtestsrv_windows-amd64.exe
```
cpu: ditto here: re `pebble-challtestsrv` (and I suspect the `artifacts` section earlier in this file will need a tweak too)

.travis.yml
adferrand: No reason in particular, I thought it was a shorter name. I changed for `pebble-challtestsrv`.

appveyor.yml
adferrand: Done here also.

cpu:
> This looks good to me. We'll hold off merging until we've got the special GitHub user created.

I'll get started on that today.

>  Or do you want to check in a Certbot-specific Pebble config for the Certbot tests?

I don't think it's a good idea to put client specific configs in-tree. I'd rather find another solution.

adferrand:
Concerning the configuration file, you can check what I am doing at https://github.com/adferrand/certbot/blob/699c1540fb16221cae998d1bcd841dda7b85bc80/certbot-ci/certbot_integration_tests/utils/pebble_artifacts.py

I simply build my own configuration file for the certbot integration tests. I think the documentation and example file in github source code are sufficient, no need to put it as a release asset.

cpu:
>  We'll hold off merging until we've got the special GitHub user created.

:ballot_box_with_check: - :wave: @pebblebot

I created the required env vars in Travis and AppVeyor.

cpu:
I'd like to try and get https://github.com/letsencrypt/pebble/pull/234 landed before I cut a release tag to test this automation end-to-end.

jsha:
> I don't think it's a good idea to put client specific configs in-tree. I'd rather find another solution.

What I meant here was checking in a config file in the Certbot repo's test directory, for use with the released Pebble.

Issue #74, opened 2017-12-13
HTTP Management Interface


cpu:
Discussion on https://github.com/letsencrypt/pebble/pull/39 indicated that there are users who would find it valuable to be able to inject test data into Pebble to help support integration testing with an ACME client.

Similarly, in https://github.com/letsencrypt/boulder/issues/2697 there was a desire for a way to signal "Reject the next `n` requests as if they had a badNonce". 

Pebble should offer a basic HTTP "admin"/"test" interface on a different path/port than the ACME API that can be used to do management tasks like the two listed above.

adferrand:
As discussed in #242, I would like to propose some basic spec around this feature.

I propose to expose every admin/test related stuff endpoint at the root context `/`, listening to a port defined in a new `adminPort` field in the Pebble JSON configuration file. The one proposed in the GIT repo will have the default value of `4000`. Not specifying this field a given configuration will disable the interface. The protocol will be HTTPS, using the same certificate than on port `14000` for the ACME protocol, on so consuming the existing configuration fields on that matter. 

What do you think of that @cpu @jsha?

cpu:
@adferrand That sounds great :+1: I like the idea of putting it on a separate port and disabling the administrative interface outright if it isn't configured. Thanks for volunteering to spec this out/implement it!

> The one proposed in the GIT repo will have the default value of 4000

This is totally a bike-shed comment but I think I'd prefer 15000 as the default port. We use `:4000` for HTTP access to the legacy ACME v1 API in Boulder and I'd prefer to avoid confusing someone into thinking any part of this management interface relates to ACME v1.

cpu:
I'm going to re-open this issue since while there is a separate HTTP(s) management interface thanks to @adferrand (:tada:) it doesn't yet let you inject test data, query expiration data, or otherwise affect validation. There is probably an argument to be made for splitting those into separate issues but for now we can use this one.

felixfontein:
I guess it is a good idea to list all PRs here which add functionality to the new management interface.

#252 (recently merged) allows to retrieve certificate revocation status (and also the certificate itself) by the certificate's serial number.

jvanasco:
I would appreciate the ability to trigger/untrigger pebble's responses so that:

* the next order(s) will be considered:
  [x]  pending (current behavior)
  []  invalid
  []  ready
  []  processing
  [] valid

* the next authorizations(s) will be considered:
  [x]  pending (current behavior)
  [] valid
  []  invalid
[...]

PR #242, opened 2019-06-19
Create a new parameter in configuration file to set the OCSP Responder URL in certificates

adferrand:
Pebble is not designed to be an OCSP responder of the certificates it issues, since it is not its purpose. However, it is easy to setup a proper OCSP responder alongside it, thanks to the ability to get the intermediate CA certificate with `/intermediate` and `/intermediate-key` endpoints.

To make this interaction complete, and allow a client (like Certbot) to get a proper OCSP response status during tests, the certificates generated by Pebble need to contain an arbitrary OCSP Responder URL.

This PR allows that, by setting a `ocspResponderURL` configuration field. If set, then Pebble will add to the certificate template the given OCSP Responder URL, to be included in the certificates when they are generated. If not set (empty string), the template does not contain the OCSP field.

Typical example is:
`"ocspResponderURL": "http://127.0.0.1:4002 pebble"`

Instruction in README are updated with this new config variable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -281,3 +281,17 @@ purposes, since Pebble and its generated keys are not audited or held to the
 same standards as the Let's Encrypt production CA and their keys. Moreover
 these keys are exposed by Pebble and will be lost as soon as the process
 terminates: so they are not safe to use for anything other than testing.**
+
+### OCSP Responder URL
+
+Pebble does not support the OCSP protocol as a responder and so does not set
+the OCSP Responder URL in the issued certificates. However, if you setup a
+proper OCSP Responder piece of software side by side with Pebble, you may want
+to set this URL. This is possible by setting the environment variable
+`PEBBLE_CA_OCSP_RESPONDER_URL`: if this variable is set to a non-empty string,
+its value will be used in the appropriate field of all issued certificates.
+
+For instance, to instruct a client to check the URL `http://127.0.0.1:4002` and
+retrieve the OCSP status of a certificate, run:
+
+`PEBBLE_CA_OCSP_RESPONDER_URL=http://127.0.0.1:4002 pebble`
--- a/README.md
+++ b/README.md
@@ -287,11 +287,13 @@ terminates: so they are not safe to use for anything other than testing.**
 Pebble does not support the OCSP protocol as a responder and so does not set
 the OCSP Responder URL in the issued certificates. However, if you setup a
 proper OCSP Responder run side by side with Pebble, you may want to set this URL.
-This is possible by setting the environment variable `PEBBLE_CA_OCSP_RESPONDER_URL`:
-if this variable is set to a non-empty string, its value will be used in the
-appropriate field of all issued certificates.
+This is possible by setting the field `ocspResponderURL` of the `pebble-config.json`
+consummed by Pebble to a non empty string: in this case, this string will be use
+in the appropriate field of all issued certificates.
 
-For instance, to have Pebble issue certificates that instruct a client to check the URL `http://127.0.0.1:4002` to
-retrieve the OCSP status of a certificate, run:
+For instance, to have Pebble issue certificates that instruct a client to check the URL `http://127.0.0.1:4002`
+to retrieve the OCSP status of a certificate, run Pebble with a `pebble-config.json` that includes:
 
-`PEBBLE_CA_OCSP_RESPONDER_URL=http://127.0.0.1:4002 pebble`
+```
+  "ocspResponderURL": "http://127.0.0.1:4002",
+```
--- a/README.md
+++ b/README.md
@@ -286,10 +286,10 @@ terminates: so they are not safe to use for anything other than testing.**
 
 Pebble does not support the OCSP protocol as a responder and so does not set
 the OCSP Responder URL in the issued certificates. However, if you setup a
-proper OCSP Responder piece of software side by side with Pebble, you may want
-to set this URL. This is possible by setting the environment variable
-`PEBBLE_CA_OCSP_RESPONDER_URL`: if this variable is set to a non-empty string,
-its value will be used in the appropriate field of all issued certificates.
+proper OCSP Responder run side by side with Pebble, you may want to set this URL.
+This is possible by setting the environment variable `PEBBLE_CA_OCSP_RESPONDER_URL`:
+if this variable is set to a non-empty string, its value will be used in the
+appropriate field of all issued certificates.
 
 For instance, to instruct a client to check the URL `http://127.0.0.1:4002` and
 retrieve the OCSP status of a certificate, run:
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -281,3 +281,17 @@ purposes, since Pebble and its generated keys are not audited or held to the
 same standards as the Let's Encrypt production CA and their keys. Moreover
 these keys are exposed by Pebble and will be lost as soon as the process
 terminates: so they are not safe to use for anything other than testing.**
+
+### OCSP Responder URL
+
+Pebble does not support the OCSP protocol as a responder and so does not set
+the OCSP Responder URL in the issued certificates. However, if you setup a
+proper OCSP Responder run side by side with Pebble, you may want to set this URL.
+This is possible by setting the environment variable `PEBBLE_CA_OCSP_RESPONDER_URL`:
```
cpu: This feels like a configuration that would be at home in the `pebble-config.json` file instead of an env var. The other env vars are related to runtime behaviour changes and feel like things someone could want to toggle quickly at the command line. The OCSP responder URL is likely to be set once, per-project and not changed. In that respect I think it fits better in the config file.

@adferrand What do you think?

ca/ca.go
```diff
@@ -22,11 +23,18 @@ import (
 const (
 	rootCAPrefix         = "Pebble Root CA "
 	intermediateCAPrefix = "Pebble Intermediate CA "
+
+	// ocspResponderURL defines the URL to set in the 1.3.6.1.5.5.7.48.1 field
```
cpu: I don't think you need to use the OID for this extension here. It's not likely to be useful for anyone reading the code without having to Google it :-) I'd stick with the Go `x509.Certificate` field name when describing this.

```suggestion
	// ocspResponderURL defines the URL to set in the OCSPServer field
```

ca/ca.go
```diff
@@ -217,6 +230,13 @@ func New(log *log.Logger, db *db.MemoryStore) *CAImpl {
 		log: log,
 		db:  db,
 	}
+
+	oscpRespURL := os.Getenv(ocspResponderURL)
+	if oscpRespURL != "" {
+		ca.ocspRespURL = oscpRespURL
+		ca.log.Printf("Setting OCSP responder URL for issued certificates to %s", ca.ocspRespURL)
```
cpu: tiny nit: I think using `%q` instead of `%s` will make this output a little bit nicer.

```suggestion
		ca.log.Printf("Setting OCSP responder URL for issued certificates to %q", ca.ocspRespURL)
```

README.md
```diff
@@ -281,3 +281,17 @@ purposes, since Pebble and its generated keys are not audited or held to the
...
+if this variable is set to a non-empty string, its value will be used in the
+appropriate field of all issued certificates.
+
+For instance, to instruct a client to check the URL `http://127.0.0.1:4002` and
```
cpu: nit: I think changing "and" to "to" makes this a little bit more readable, and I'd maybe make it a little bit clearer how the client learns the OCSP server URL (via a field in the certificate, not via ACME).

```suggestion
For instance, to have Pebble issue certificates that instruct a client to check the URL `http://127.0.0.1:4002` to
```

README.md
adferrand: Initially I implemented this as a flag to pass to the command line. Then I thought it was not worth a fully-fledged flag changing pebble's command API for something which is nonetheless a corner case. So I moved to a environment variable.

But you are right, it is too much ephemeral like this. Your proposition to pass it in the configuration file seems a good compromise. I will do that.

ca/ca.go
adferrand: Oh, I was not aware of %q to quote something!

README.md
adferrand: Implemented.

test/config/pebble-config.json
```diff
@@ -4,6 +4,7 @@
     "certificate": "test/certs/localhost/cert.pem",
     "privateKey": "test/certs/localhost/key.pem",
     "httpPort": 5002,
-    "tlsPort": 5001
+    "tlsPort": 5001,
+    "ocspResponderURL": "test"
```
cpu: Can you make this value `""` in the example config instead? I think it makes the configuration field discoverable by those that need it, but prevents the default configuration from including a non-sense URL in the issued certificates.
```suggestion
    "ocspResponderURL": ""
```

adferrand: Ah yes, I was about to set this as empty, it was a mistake during the push.

cpu:
I think the scope of this PR is small enough that it can be merged without blocking on a second review. I'll pull the trigger when CI is satisfied.

felixfontein:
Out of curiosity: how do you allow the OCSP responder to interact with pebble, to look up a certificate / its revocation status via its serial? (Also, out of curiosity: which OCSP responder are you using for testing?)

adferrand:
In fact I do make communicate the OCSP responder with Pebble, since there is no visible communication of Pebble to the outside world (like files or webhooks). The idea is to have a mock OCSP responder, that will blindly return an OCSP response, for a certificate and a OCSP status decided at the creation of the mock.

One tricky part are to make the client (here Certbot) to contact the mock OCSP server. This is done by setting the `ocspResponderURL` that my PR added here. The other tricky part is to return a correct (in term of structure and signature) OCSP response. This is done thanks to the builders available in `cryptography` Python project, that consumes the key/cert issuer from the endpoints I also exposed from Pebble in https://github.com/letsencrypt/pebble/pull/236.

You can see the work in progress of this here: https://github.com/adferrand/certbot/pull/53

felixfontein:
Hmm, ok, if you don't care whether the OCSP answer is correct (in terms of revoked / not revoked state), then that's fine.

I think it would be a good idea to add an endpoint to Pebble which returns for a certificate serial whether it is unknown, known and valid, or known and revoked. Pebble already keeps track of revoked certificates (so that the correct answer can be given if the certificate is revoked twice with ACME), so implementing that should be pretty simple.

@cpu @jsha would such an endpoint be OK for you? I can create a PR for it if you want it.

adferrand:
If you provide this endpoint, I would definitely build my mocked ocsp server on top of it, and make it a "real" one.

cpu:
> @cpu @jsha would such an endpoint be OK for you? I can create a PR for it if you want it.

Yes provided its clear that this feature is part of a separate Pebble-specific management interface (like described in #74) and not a general ACME/RFC 8555 method. I [mentioned](https://github.com/letsencrypt/pebble/issues/74#issuecomment-441655468) supporting a light-weight way to query revocation status as part of that issue and still think it would be valuable. Hearing two separate client implementers express interest also helps :-)

@jsha Do you have any additional thoughts?

adferrand:
On this case, I think we should also move the endpoints `/root`, `/intermediate` and stuff to this dedicated interface, as they are not part also of the RFC 8555 spec.

cpu:
@adferrand Agreed, I think that would be a sensible accompanying change.

Issue #245, opened 2019-06-21
strict: reject JWS bodies that include "resource" field.


cpu:
Some legacy ACME v1 clients may end up sending a `"resource"` field with the request URL in ACME v2 / RFC 8555 settings (e.g. [Cerbot](https://github.com/certbot/certbot/issues/5708)). 

Pebble should reject these requests in `-strict` mode. Yes: the "resource" field isn't required/specified in RFC 8555 JWS bodies because the full request URL is in the JWS protected headers.

tammersaleh:
I'm attempting to make use of the `acme` python library (part of Certbot), using the v2 ACME specification.  When running `pebble -strict`, it's triggering this error.  Just to clarify, does that mean that my application won't be v2 compliant (as v1 is being deprecated), or is this normal behavior for v2 clients?  I'm trying to determine if I have a bug in my code.

tammersaleh:
(also, thank you for your help on pebble - it's exactly what we needed for local testing!)

tammersaleh:
Sorry, I just saw that [`acme` now supports this correctly](https://github.com/certbot/certbot/pull/7176).

jsha:
Yep, you spotted it! The latest version of `acme` does the right thing.

Issue #244, opened 2019-06-20
strict: Reject non-empty JSON body in challenge POST


cpu:
[RFC 8555 Section 7.5.1](https://tools.ietf.org/html/rfc8555#section-7.5.1) says unambiguously that challenges are initiated by POSTing `{}` in the JWS body:

>  The client indicates to the server that it is ready for the challenge
>  validation by sending an empty JSON body ("{}") carried in a POST
>  request to the challenge URL (not the authorization URL).

Presently Pebble rejects challenge POST's that contain the legacy `keyAuthorization` field by default:
https://github.com/letsencrypt/pebble/blob/afbe2db32cf03eb8c240a696e21fc52f7b0be6ab/wfe/wfe.go#L1855-L1876

Pebble in `-strict` should go one step further and reject anything other than the trivial empty JSON payload `{}`. This will help push ACME clients that are POSTing non-standard JSON bodies (e.g. [`Certbot`](https://community.letsencrypt.org/t/acme-spec-asks-for-empty-json-blob-in-challenge-validation-request/96252)) to move towards spec compliance.

PR #246, opened 2019-06-21
WFE: reject legacy JWS requests in -strict

cpu:
## wfe: reject out-of-spec chall POST in strict.

Challenge POST JWS bodies should be exactly `{}` per [RFC 8555 Section 7.5.1](https://tools.ietf.org/html/rfc8555#section-7.5.1).

Running `pebble -strict` and using the `chisel2.py` client based on the Certbot `acme` module now results in the following Problem being returned on the first challenge initiation POST:
```
{
   "type": "urn:ietf:params:acme:error:malformed",
   "detail": "challenge initiation POST JWS body was not \"{}\"",
   "status": 400
}
```
See https://github.com/certbot/certbot/issues/7171

Resolves https://github.com/letsencrypt/pebble/issues/244

## wfe: reject JWS JSON with non-empty "resource".

The "resource" field of a JSON JWS body is a legacy artifact of ACME v1 and should not be sent in RFC 8555/ACME v2 requests.

Running `pebble -strict` and using the `chisel2.py` client based on the Certbot `acme` module now results in the following Problem being returned on the first POST request to `newAccount`:
```
{
   "type": "urn:ietf:params:acme:error:malformed",
   "detail": "JWS body included JSON with a deprecated ACME v1 \"resource\" field (\"new-reg\")",
   "status": 400
}
```
See https://github.com/certbot/certbot/issues/5708

Resolves https://github.com/letsencrypt/pebble/issues/245

### DIFF
--- a/wfe/wfe.go
+++ b/wfe/wfe.go
@@ -1870,7 +1870,7 @@ func (wfe *WebFrontEndImpl) updateChallenge(
 
 	// In strict mode we reject any challenge POST with a body other than `{}`.
 	// This matches RFC 8555 Section 7.5.1.
-	if wfe.strict && bytes.Compare(postData.body, []byte("{}")) != 0 {
+	if wfe.strict && !bytes.Equal(postData.body, []byte("{}")) {
 		wfe.sendError(
 			acme.MalformedProblem(`challenge initiation POST JWS body was not "{}"`), response)
 		return
--- a/wfe/wfe.go
+++ b/wfe/wfe.go
@@ -663,6 +663,22 @@ func (wfe *WebFrontEndImpl) verifyJWS(
 			expectedURL.String(), headerURL))
 	}
 
+	// In -strict mode, verify that any JWS body that is valid JSON doesn't
+	// include a non-empty "resource" field. This is a legacy artifiact from ACME
+	// v1. This won't catch an empty "resource" field but that would have been
+	// broken in ACMEv1 anyway and is hopefully less likely to occur in code that
+	// is updated for ACMEv2.
+	if wfe.strict {
+		var bodyObj struct {
+			Resource string
+		}
+		if err := json.Unmarshal(payload, &bodyObj); err == nil && bodyObj.Resource != "" {
+			return nil, acme.MalformedProblem(fmt.Sprintf(
+				`JWS body included JSON with a deprecated ACME v1 "resource" field (%q)`,
+				bodyObj.Resource))
+		}
+	}
+
 	return &authenticatedPOST{
 		postAsGet: string(payload) == "",
 		body:      payload,
--- a/wfe/wfe.go
+++ b/wfe/wfe.go
@@ -1869,7 +1869,11 @@ func (wfe *WebFrontEndImpl) updateChallenge(
 	}
 
 	// In strict mode we reject any challenge POST with a body other than `{}`.
-	// This matches RFC 8555 Section 7.5.1.
+	// This matches RFC 8555 Section 7.5.1 and the ACME challenge types that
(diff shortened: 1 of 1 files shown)


wfe/wfe.go
```diff
@@ -1852,27 +1868,37 @@ func (wfe *WebFrontEndImpl) updateChallenge(
...
 	}
 
-	var chalResp struct {
-		KeyAuthorization *string
-	}
-	err := json.Unmarshal(postData.body, &chalResp)
-	if err != nil {
+	// In strict mode we reject any challenge POST with a body other than `{}`.
+	// This matches RFC 8555 Section 7.5.1.
+	if wfe.strict && bytes.Compare(postData.body, []byte("{}")) != 0 {
```
golangcibot: S1004: should use !bytes.Equal(postData.body, []byte("{}")) instead (from `gosimple`)

cpu: Resolved in 248c6b3

wfe/wfe.go
```diff
@@ -1852,27 +1868,37 @@ func (wfe *WebFrontEndImpl) updateChallenge(
 		return
 	}
 
-	var chalResp struct {
-		KeyAuthorization *string
-	}
-	err := json.Unmarshal(postData.body, &chalResp)
-	if err != nil {
+	// In strict mode we reject any challenge POST with a body other than `{}`.
+	// This matches RFC 8555 Section 7.5.1.
```
jsha: This is true for all challenges Pebble implements, though [per this erratum](https://www.rfc-editor.org/errata/eid5729) it might not be true for future challenges. Since some people will probably turn to Pebble to understand the spec better, could you describe that situation and link to the erratum?

wfe/wfe.go
```diff
@@ -1852,27 +1868,37 @@ func (wfe *WebFrontEndImpl) updateChallenge(
...
-	// sent (and that's what Boulder will do) but for Pebble we'd like to offer
-	// a way to be more aggressive about pushing clients implementations in the
-	// right direction, so we treat this as a malformed request.
-	if chalResp.KeyAuthorization != nil {
-		wfe.sendError(
-			acme.MalformedProblem(
-				"Challenge response body contained legacy KeyAuthorization field, "+
-					"POST body should be `{}`"), response)
-		return
+		// Historically challenges were updated by POSTing a KeyAuthorization. This is
```
jsha: It looks like we already reject challenge POSTs with "KeyAuthorization", even in non-strict mode. So we don't need to guard the new code with an `if strict`. We can just make it more aggressive if we want. Alternately we can consider the existing KeyAuthorization rejection code as good enough, when combined with the new `resource` rejection, since there were no other fields in common use for challenge posts.

wfe/wfe.go
cpu: Sure

wfe/wfe.go
cpu: I think you might have read the diff wrong. The key auth reject is in the `else` of a check for strict. E.g. it only happens when strict is disabled. I could have left it outside of the `else` entirely but it felt silly to check for a key auth field in the case where we've already verified the body is `{}` for strict.

cpu: We do need to guard the new code with `-strict` because otherwise integration tests will fail.

wfe/wfe.go
cpu: Resolved in 0580ec4

wfe/wfe.go
jsha: Oh yeah, I get it now. I was thinking of `resource` as already handled by the auth code, but it also comes into play here, and is still send by `acme`.

PR #248, opened 2019-06-30
Expose a management interface in Pebble

adferrand:
Fixes https://github.com/letsencrypt/pebble/issues/74

This PR adds a management interface, that listens to a specific address and port as defined by the field `managementListenAddress` in the Pebble JSON configuration file, and serves endpoints unrelated to the ACME protocol, but useful to interface Pebble in a testing environment. This interface is reached through HTTPS and is secured using the same certificate than the ACME interface.

I moved on this interface the endpoints related to dump the internal certificates/keys used by the CA in Pebble. I disabled the redirections done from `/intermediate` to `/intermediates/0` (and equivalent for the other endpoints) because I think it adds unnessarily some complexity, and most of these endpoints have been added in the latest release by me, and I do not consumes them yet. However if you disagree with this approach for `/root`, since it was there before, I can make a specific redirection from `/root` on the ACME port to `/roots/0` on the management port. 

Not setting `managementListenAddress` disables the management interface.

`README.md` and docker compose files are updated accordingly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -275,7 +275,7 @@ This interface is configured by the `managementListenAddress` field in
 interface will listen on. Set `managementListenAddress` to an empty string or `null`
 to disable it.
 
-The proposed configuration for this management interface as defined in
+The default configuration for this management interface as defined in
 `test/config/pebble-config.yml` is to listen on any address on port 15000:
 
 ```
--- a/README.md
+++ b/README.md
@@ -268,7 +268,7 @@ repo](test/certs/pebble.minica.key.pem).**
 
 In order to ease the interaction of Pebble with testing systems, a specific HTTP
 management interface is exposed on a different port than the ACME protocol,
-and exposes several useful endpoints.
+and offers several useful testing endpoints.
 
 This interface is configured by the `managementListenAddress` field in
 `pebble-config.json` that defines the address and the port on which the management
--- a/README.md
+++ b/README.md
@@ -285,7 +285,7 @@ The proposed configuration for this management interface as defined in
 #### CA Root and Intermediate Certificates
 
 Note that the CA's root and intermediate certificates are regenerated on every
-launch. It can be retrieved by a `GET` request to `https://localhost:15000/roots/0`
+launch. They can be retrieved by a `GET` request to `https://localhost:15000/roots/0`
 and `https://localhost:15000/intermediates/0` respectively.
 
 You might need the root certificate to verify the complete trust chain of
--- a/README.md
+++ b/README.md
@@ -270,7 +270,7 @@ In order to ease the interaction of Pebble with testing systems, a specific HTTP
 management interface is exposed on a different port than the ACME protocol,
 and offers several useful testing endpoints.
 
-These endpoints are specific to Pebble and its internal behavior, but are not part
+These endpoints are specific to Pebble and its internal behavior, and are not part
 of the RFC 8555 that defines the ACME protocol.
 
 The management interface is configured by the `managementListenAddress` field in
--- a/README.md
+++ b/README.md
@@ -97,7 +97,8 @@ docker-compose up
 ```
 
 Afterwards you can access the ACME API from your host machine at
-`https://localhost:14000/dir` and the `pebble-challtestsrv`'s management
+`https://localhost:14000/dir`, the ̀`pebble`'s management interface
+at `https://localhost:15000` and the `pebble-challtestsrv`'s management
 interface at `http://localhost:8055`.
 
 To get started you may want to update the `pebble-challtestsrv` mock DNS data
@@ -127,7 +128,8 @@ services:
   image: letsencrypt/pebble
   command: pebble -config /test/my-pebble-config.json
   ports:
-    - 14000:14000
+    - 14000:14000  # ACME port
+    - 15000:15000  # Management port
   environment:
     - PEBBLE_VA_NOSLEEP=1
   volumes:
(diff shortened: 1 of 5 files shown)


felixfontein:
I think there definitely should be a redirect for `/root`, if this is released in just another minor Pebble version increase. For `/intermediate`, `/root-key` and `/intermediate-key`, I'm kind of divided, I guess @jsha / @cpu have to decide this. On the other hand, the alternate-friendly URLs (`/roots`, `/intermdiates`, etc.) have been available as well in a released version (2.1.0), so they are potentially consumed by other people. So one could argue that these should be redirected as well. Sounds like a big mess :)

Maybe it is a good idea to cut off backwards compatibility (even for `/root`) before we have a huge amount of redirects? This should probably be accompanied by a major version bump to say 3.0.0. I think I would prefer this option.

cmd/pebble/main.go
```diff
@@ -71,11 +72,26 @@ func main() {
 	wfeImpl := wfe.New(logger, db, va, ca, *strictMode)
 	muxHandler := wfeImpl.Handler()
 
+	if len(c.Pebble.ManagementListenAddress) > 0 {
+		go func() {
```
felixfontein: I don't know Go that well. Is there a reason you're using an anonymous function here which you directly call, instead of just running this code directly? I've only seen this construct in JavaScript before, but I don't think Go is as horribly broken as JS to justify this construct :D

adferrand:
About https://github.com/letsencrypt/pebble/pull/248#issuecomment-507032612

I would agree with you for a piece software designed for production. But this is for testing, and Pebble is already designed to break systems. For instance, it breaks regularly Certbot CI on strict mode, when new controls regarding ACME protocol are added.

So I think it is not a big deal to break the retro-compatibility if a proper version bump is done, and help to keep the code as short and as simple as possible.

Indeed we will see the opinion of @cpu and @jsha regarding this issue :)

cmd/pebble/main.go
adferrand: Well it is not for the same purpose. Autocall anonymous functions in Javascript are used to isolate the context (variables, mainly) altered by the called function, to not leak them in the outer scope.

Here it is to create a goroutine to execute several actions, then the blocking call to `ListenAndServeTLS`. This way you can listen on several interfaces on parallel.

felixfontein: Ah, now I notice the small little `go` in front of `func()`. I've never used/seen goroutines before, so sorry for missing that :)

cpu:
> So I think it is not a big deal to break the retro-compatibility if a proper version bump is done, and help to keep the code as short and as simple as possible.

I'm :+1: on bumping the major version and not making life too complicated :-)

cmd/pebble/main.go
adferrand: No problem!

cmd/pebble/main.go
```diff
@@ -71,11 +72,26 @@ func main() {
...
+				c.Pebble.Certificate,
+				c.Pebble.PrivateKey,
+				adminHandler)
+			cmd.FailOnError(err, "Calling ListenAndServeTLS() for admin interface")
+		}()
+		logger.Printf("Management interface listening on: %s\n", c.Pebble.ManagementListenAddress)
+		logger.Printf("Root CA certificate(s) available at: https://%s%s",
+			c.Pebble.ManagementListenAddress, wfe.RootCertPath)
+	} else {
+		logger.Printf("Management interface is disabled")
```
cpu: Nit: Since this isn't using a format string you can switch to `logger.Print`
```suggestion
		logger.Print("Management interface is disabled")
```

wfe/wfe.go
```diff
@@ -364,6 +344,17 @@ func (wfe *WebFrontEndImpl) Handler() http.Handler {
...
 }
 
+func (wfe *WebFrontEndImpl) ManagementHandler() http.Handler {
+	m := http.NewServeMux()
+	// GET only handlers
+	wfe.HandleFunc(m, RootCertPath, wfe.handleCert(wfe.ca.GetRootCert, RootCertPath), "GET")
+	wfe.HandleFunc(m, rootKeyPath, wfe.handleKey(wfe.ca.GetRootKey, rootKeyPath), "GET")
+	wfe.HandleFunc(m, intermediateCertPath, wfe.handleCert(wfe.ca.GetIntermediateCert, intermediateCertPath), "GET")
+	wfe.HandleFunc(m, intermediateKeyPath, wfe.handleKey(wfe.ca.GetIntermediateKey, intermediateKeyPath), "GET")
+
```
cpu: nit: I think you can lose this whitespace without making the function harder to read.
```suggestion
```

wfe/wfe.go
```diff
@@ -364,6 +344,17 @@ func (wfe *WebFrontEndImpl) Handler() http.Handler {
 	return m
 }
 
+func (wfe *WebFrontEndImpl) ManagementHandler() http.Handler {
```
cpu: Can you add a brief godoc comment to this function that explains its purpose?

cmd/pebble/main.go
```diff
@@ -71,11 +72,26 @@ func main() {
 	wfeImpl := wfe.New(logger, db, va, ca, *strictMode)
 	muxHandler := wfeImpl.Handler()
 
+	if len(c.Pebble.ManagementListenAddress) > 0 {
```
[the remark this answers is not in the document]
adferrand: Yes I think so also. I was just following this approach because it is used on another if branch above, but it is certainly better to stick to idiomatic solutions nonetheless.

README.md
```diff
@@ -262,17 +264,35 @@ store or to any production systems/codebases. The private key for this CA is
 intentionally made [publicly available in this
 repo](test/certs/pebble.minica.key.pem).**
 
-### CA Root and Intermediate Certificates
+### Management interface
```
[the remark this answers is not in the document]
adferrand: I did that.

wfe/wfe.go
adferrand: I added the godoc comment.

cmd/pebble/main.go
```diff
@@ -71,11 +72,26 @@ func main() {
...
+			adminHandler := wfeImpl.ManagementHandler()
+			err = http.ListenAndServeTLS(
+				c.Pebble.ManagementListenAddress,
+				c.Pebble.Certificate,
+				c.Pebble.PrivateKey,
+				adminHandler)
+			cmd.FailOnError(err, "Calling ListenAndServeTLS() for admin interface")
+		}()
+		logger.Printf("Management interface listening on: %s\n", c.Pebble.ManagementListenAddress)
+		logger.Printf("Root CA certificate(s) available at: https://%s%s",
```
felixfontein: I would either write
```suggestion
		logger.Printf("Root CA certificate available at: https://%s%s0",
```
or something like
```suggestion
		logger.Printf("Root CA certificate(s) available at: https://%s%s<no>, where <no> is 0, 1, ...",
```
(or even include `alternateRoots` in here)

adferrand: I implemented your last proposition: if several alternate roots are generated, each of them has a dedicated log to display its dedicated `root` endpoint.

cmd/pebble/main.go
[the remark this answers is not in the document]
cpu: Ahh! Apologies for missing that other bit in earlier reviews :-) Thanks for fixing this instance :+1:

README.md
```diff
@@ -262,17 +264,38 @@ store or to any production systems/codebases. The private key for this CA is
...
+
+In order to ease the interaction of Pebble with testing systems, a specific HTTP
+management interface is exposed on a different port than the ACME protocol,
+and offers several useful testing endpoints.
+
+These endpoints are specific to Pebble and its internal behavior, but are not part
```
cpu: minor grammatical nit:
```suggestion
These endpoints are specific to Pebble and its internal behavior, and are not part
```

cmd/pebble/main.go
cpu: Thanks for the :mag: @felixfontein !

cpu:
@jsha Would you like to give this a second :mag: before its merged?

cpu:
> @jsha Would you like to give this a second mag before its merged?

I'm going to go ahead and merge this. It's largely a shuffling of existing functionality.

felixfontein:
Thanks for implementing this! I'll take a look at adding a management endpoint for revocation info probably this weekend.

PR #252, opened 2019-07-12
Management interface: add /cert-status-by-serial/ endpoint

felixfontein:
This adds a `/cert-status-by-serial/` endpoint to the management interface. The certificate is identified by its serial number (in hexadecimal). (It can be extracted with OpenSSL via `openssl x509 -in cert.pem -noout -text | sed -En 's/.*Serial Number.*\(0x([0-9a-f]+)\)/\1/p'`.) The endpoint returns the certificate itself (in PEM format), the serial (hexadecimal) and the revocation status (`Valid` or `Revoked`) as JSON.

Example usage:
```
$ curl -ki https://127.0.0.1:15000/cert-status-by-serial/5ab954d39632fd78
HTTP/2 200 
content-type: application/json; charset=utf-8
link: <https://127.0.0.1:15000/dir>;rel="index"
content-length: 1671
date: Fri, 12 Jul 2019 21:39:24 GMT

{
   "Certificate": "-----BEGIN CERTIFICATE-----\nMIIEVz...tcw=\n-----END CERTIFICATE-----\n",
   "Serial": "5ab954d39632fd78",
   "Status": "Valid"
}

$ curl -ki https://127.0.0.1:15000/cert-status-by-serial/7811e0b42b98cc1c
HTTP/2 200 
content-type: application/json; charset=utf-8
link: <https://127.0.0.1:15000/dir>;rel="index"
content-length: 1728
date: Fri, 12 Jul 2019 22:14:17 GMT

{
   "Certificate": "-----BEGIN CERTIFICATE-----\nMIIEVz...tcw=\n-----END CERTIFICATE-----\n",
   "RevokedAt": "2019-07-13T00:14:13.719654003+02:00",
   "Serial": "7811e0b42b98cc1c",
   "Status": "Revoked"
}

$ curl -ki https://127.0.0.1:15000/cert-status-by-serial/66317d2e02f5d3d6
HTTP/2 200 
content-type: application/json; charset=utf-8
link: <https://127.0.0.1:15000/dir>;rel="index"
content-length: 1740
date: Fri, 12 Jul 2019 22:14:21 GMT

{
   "Certificate": "-----BEGIN CERTIFICATE-----\nMIIEVz...tcw=\n-----END CERTIFICATE-----\n",
   "Reason": 4,
   "RevokedAt": "2019-07-13T00:13:20.418489956+02:00",
   "Serial": "66317d2e02f5d3d6",
   "Status": "Revoked"
}
```

CC @adferrand

### DIFF
--- a/README.md
+++ b/README.md
@@ -316,7 +316,7 @@ intermediate certificates and keys.
 
 The certificate (in PEM format) and its revocation status can be queried by doing
 a `GET` request to `https://localhost:15000/cert-status-by-serial/<serial>`, where
-`<serial>` is the hexadecimal representation of the certificate's serial number.
+`<serial>` is the hexadecimal representation of the certificate's serial number (no `0x` prefix).
 It can be obtained via:
 
     openssl x509 -in cert.pem -noout -serial | cut -d= -f2
--- a/README.md
+++ b/README.md
@@ -312,6 +312,17 @@ request to `https://localhost:15000/roots/0`, `https://localhost:15000/root-keys
 etc. These endpoints also send `Link` HTTP headers for all alternative root and
 intermediate certificates and keys.
 
+#### Certificate Status
+
+The certificate (in PEM format) and its revocation status can be queried by doing
+a `GET` request to `https://localhost:15000/cert-status-by-serial/<serial>`, where
+`<serial>` is the hexadecimal representation of the certificate's serial number.
+It can be obtained via:
+
+    openssl x509 -in cert.pem -noout -text | sed -En 's/.*Serial Number.*\(0x([0-9a-f]+)\)/\1/p'
+
+The endpoint returns the information as a JSON.
+
 ### OCSP Responder URL
 
 Pebble does not support the OCSP protocol as a responder and so does not set
--- a/README.md
+++ b/README.md
@@ -314,7 +314,7 @@ intermediate certificates and keys.
 
 #### Certificate Status
 
-The certificate (in PEM format) and its revocation status can be queried by doing
+The certificate (in PEM format) and its revocation status can be queried by sending
 a `GET` request to `https://localhost:15000/cert-status-by-serial/<serial>`, where
 `<serial>` is the hexadecimal representation of the certificate's serial number (no `0x` prefix).
 It can be obtained via:
--- a/README.md
+++ b/README.md
@@ -319,9 +319,26 @@ a `GET` request to `https://localhost:15000/cert-status-by-serial/<serial>`, whe
 `<serial>` is the hexadecimal representation of the certificate's serial number.
 It can be obtained via:
 
-    openssl x509 -in cert.pem -noout -text | sed -En 's/.*Serial Number.*\(0x([0-9a-f]+)\)/\1/p'
+    openssl x509 -in cert.pem -noout -serial | cut -d= -f2
 
-The endpoint returns the information as a JSON.
+The endpoint returns the information as a JSON:
+
+    $ curl -ki https://127.0.0.1:15000/cert-status-by-serial/66317d2e02f5d3d6
+    HTTP/2 200
+    cache-control: public, max-age=0, no-cache
+    content-type: application/json; charset=utf-8
+    link: <https://127.0.0.1:15000/dir>;rel="index"
+    content-length: 1740
+    date: Fri, 12 Jul 2019 22:14:21 GMT
+
+    {
+       "Certificate": "-----BEGIN CERTIFICATE-----\nMIIEVz...tcw=\n-----END CERTIFICATE-----\n",
+       "Reason": 4,
+       "RevokedAt": "2019-07-13T00:13:20.418489956+02:00",
+       "Serial": "66317d2e02f5d3d6",
+       "Status": "Revoked"
+    }
+
 
 ### OCSP Responder URL
 
--- a/README.md
+++ b/README.md
@@ -321,7 +321,7 @@ It can be obtained via:
 
     openssl x509 -in cert.pem -noout -serial | cut -d= -f2
 
-The endpoint returns the information as a JSON:
+The endpoint returns the information as a JSON object:
 
     $ curl -ki https://127.0.0.1:15000/cert-status-by-serial/66317d2e02f5d3d6
     HTTP/2 200
--- a/core/types.go
+++ b/core/types.go
@@ -192,6 +192,7 @@ func (c Certificate) Chain(no int) []byte {
 	return bytes.Join(chain, nil)
 }
 
+// RevokedCertificate is a certificate together with information about its revocation.
 type RevokedCertificate struct {
 	Certificate *Certificate
 	RevokedAt   time.Time
--- a/core/types.go
+++ b/core/types.go
@@ -192,6 +192,12 @@ func (c Certificate) Chain(no int) []byte {
 	return bytes.Join(chain, nil)
 }
 
+type RevokedCertificate struct {
+	Certificate *Certificate
+	RevokedAt   time.Time
+	Reason      *uint
+}
+
 type ValidationRecord struct {
 	URL         string
 	Error       *acme.ProblemDetails
--- a/db/memorystore.go
+++ b/db/memorystore.go
@@ -6,6 +6,7 @@ import (
 	"crypto/x509"
 	"encoding/hex"
 	"fmt"
+	"math/big"
 	"reflect"
 	"strconv"
 	"sync"
@@ -321,4 +322,33 @@ func keyToID(key crypto.PublicKey) (string, error) {
 		spkiDigest := sha256.Sum256(keyDER)
 		return hex.EncodeToString(spkiDigest[:]), nil
 	}
+}
+
+// GetCertificateBySerial loops over all certificates to find the one that matches the provided
+// serial number. This method is linear and it's not optimized to give you a quick response.
+func (m *MemoryStore) GetCertificateBySerial(serialNumber *big.Int) *core.Certificate {
+	m.RLock()
+	defer m.RUnlock()
+	for _, c := range m.certificatesByID {
+		if serialNumber.Cmp(c.Cert.SerialNumber) == 0 {
+			return c
+		}
+	}
+
+	return nil
+}
+
+// GetCertificateBySerial loops over all revoked certificates to find the one that matches the
+// provided serial number. This method is linear and it's not optimized to give you a quick
+// response.
+func (m *MemoryStore) GetRevokedCertificateBySerial(serialNumber *big.Int) *core.Certificate {
+	m.RLock()
+	defer m.RUnlock()
+	for _, c := range m.revokedCertificatesByID {
+		if serialNumber.Cmp(c.Cert.SerialNumber) == 0 {
+			return c
+		}
+	}
+
+	return nil
 }
--- a/db/memorystore.go
+++ b/db/memorystore.go
@@ -47,7 +47,7 @@ type MemoryStore struct {
 	challengesByID map[string]*core.Challenge
 
 	certificatesByID        map[string]*core.Certificate
-	revokedCertificatesByID map[string]*core.Certificate
+	revokedCertificatesByID map[string]*core.RevokedCertificate
 }
 
 func NewMemoryStore() *MemoryStore {
@@ -59,7 +59,7 @@ func NewMemoryStore() *MemoryStore {
 		authorizationsByID:      make(map[string]*core.Authorization),
 		challengesByID:          make(map[string]*core.Challenge),
 		certificatesByID:        make(map[string]*core.Certificate),
-		revokedCertificatesByID: make(map[string]*core.Certificate),
+		revokedCertificatesByID: make(map[string]*core.RevokedCertificate),
 	}
 }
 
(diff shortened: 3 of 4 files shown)


felixfontein:
I forgot: if Pebble doesn't know the certificate, or can't parse the serial as hex, it will return a 404 with no content.

adferrand:
Looks promising!

One interesting thing to get also (to build a complete OCSP responder for instance) would be the date since when the certificate has been revoked on Pebble side. It could be retrieved as a fourth parameter in the JSON response, if the status is `Revoked`.

I think it is doable to store this date in the `revokedCertificatesById` map from `ca.db`.

What do you think of that @felixfontein?

felixfontein:
Yes, that should be doable. I would include a full timestamp and not just a date, though. And also the revocation code used. I'm trying it now...

felixfontein:
Now revocation timestamp and reason are stored with the revoked certificate, and returned by the endpoint (the reason only when available):
```
$ curl -ki https://127.0.0.1:15000/cert-status-by-serial/7811e0b42b98cc1c
HTTP/2 200 
content-type: application/json; charset=utf-8
link: <https://127.0.0.1:15000/dir>;rel="index"
content-length: 1728
date: Fri, 12 Jul 2019 22:14:17 GMT

{
   "Certificate": "-----BEGIN CERTIFICATE-----\nMIIEVz...tcw=\n-----END CERTIFICATE-----\n",
   "RevokedAt": "2019-07-13T00:14:13.719654003+02:00",
   "Serial": "7811e0b42b98cc1c",
   "Status": "Revoked"
}

$ curl -ki https://127.0.0.1:15000/cert-status-by-serial/66317d2e02f5d3d6
HTTP/2 200 
content-type: application/json; charset=utf-8
link: <https://127.0.0.1:15000/dir>;rel="index"
content-length: 1740
date: Fri, 12 Jul 2019 22:14:21 GMT

{
   "Certificate": "-----BEGIN CERTIFICATE-----\nMIIEVz...tcw=\n-----END CERTIFICATE-----\n",
   "Reason": 4,
   "RevokedAt": "2019-07-13T00:13:20.418489956+02:00",
   "Serial": "66317d2e02f5d3d6",
   "Status": "Revoked"
}
```

felixfontein:
I changed it a little so that the timestamp is returned in UTC:
```
    "RevokedAt": "2019-07-12T22:21:05.392675552Z",
```

felixfontein:
BTW (and unrelated): all requests to the management interface have a `Link` header for the directory included, but with the wrong port (that of the management interface).

cpu:
> BTW (and unrelated): all requests to the management interface have a Link header for the directory 

@felixfontein Good catch! https://github.com/letsencrypt/pebble/issues/253

I'll try to review this PR in the next couple of days.

felixfontein:
Rebased to remove conflicts with #254.

README.md
```diff
@@ -312,6 +312,17 @@ request to `https://localhost:15000/roots/0`, `https://localhost:15000/root-keys
...
 intermediate certificates and keys.
 
+#### Certificate Status
+
+The certificate (in PEM format) and its revocation status can be queried by doing
+a `GET` request to `https://localhost:15000/cert-status-by-serial/<serial>`, where
+`<serial>` is the hexadecimal representation of the certificate's serial number.
+It can be obtained via:
+
+    openssl x509 -in cert.pem -noout -text | sed -En 's/.*Serial Number.*\(0x([0-9a-f]+)\)/\1/p'
```
cpu: nit: I think it's slightly easier to get the serial number with OpenSSL this way, no regex required.
`openssl x509 -in cert.pem -noout -serial | cut -d= -f2`

README.md
```diff
@@ -312,6 +312,17 @@ request to `https://localhost:15000/roots/0`, `https://localhost:15000/root-keys
...
+
+The endpoint returns the information as a JSON.
```
cpu: I think it would be helpful to add an example response here like you did in the PR description.

core/types.go
```diff
@@ -192,6 +192,12 @@ func (c Certificate) Chain(no int) []byte {
 	return bytes.Join(chain, nil)
 }
 
+type RevokedCertificate struct {
```
cpu: Could you add a short godoc comment of the form "// RevokedCertificate is ...." to this type?

db/memorystore.go
```diff
@@ -322,3 +323,32 @@ func keyToID(key crypto.PublicKey) (string, error) {
...
+	for _, c := range m.certificatesByID {
+		if serialNumber.Cmp(c.Cert.SerialNumber) == 0 {
+			return c
+		}
+	}
+
+	return nil
+}
+
+// GetCertificateBySerial loops over all revoked certificates to find the one that matches the
```
cpu: typo: the function name in your Godoc comment here is the wrong one :-)
```suggestion
// GetRevokedCertificateBySerial loops over all revoked certificates to find the one that matches the
```

wfe/wfe.go
```diff
@@ -329,6 +331,54 @@ func (wfe *WebFrontEndImpl) handleKey(
...
 
+func (wfe *WebFrontEndImpl) handleCertStatusBySerial(
+	ctx context.Context,
+	response http.ResponseWriter,
+	request *http.Request) {
+
+	serialStr := strings.TrimPrefix(request.URL.Path, certStatusBySerial)
+	serial := big.NewInt(0)
+	if _, ok := serial.SetString(serialStr, 16); !ok {
+		response.WriteHeader(http.StatusNotFound)
```
cpu: nit: in this case I think it makes more sense to return `http.StatusBadRequest` instead of `http.StatusNotFound` since it indicates the serial wasn't a valid hex encoded bigint.
```suggestion
		response.WriteHeader(http.StatusBadRequest)
```

wfe/wfe.go
```diff
@@ -329,6 +331,54 @@ func (wfe *WebFrontEndImpl) handleKey(
...
+		return
+	}
+
+	var status string
+	var cert *core.Certificate
+	var rcert *core.RevokedCertificate
+	if cert = wfe.db.GetCertificateBySerial(serial); cert != nil {
+		status = "Valid"
+	}
+	if rcert = wfe.db.GetRevokedCertificateBySerial(serial); rcert != nil {
```
cpu: How do you feel about:

1. Flipping the order of these checks so that `wfe.db.GetRevokedCertificateBySerial` is checked first
2. Making the other clause an `else if`.

My reasoning for the first change is that I suspect more often Pebble will be asked about cert's that are expected to be revoked and so we can save some of the linear search time by doing that first. There's also likely to be less certificates to scan linearly in the revoked slice than the overall issued slice.

wfe/wfe.go
```diff
@@ -329,6 +331,54 @@ func (wfe *WebFrontEndImpl) handleKey(
...
+		status = "Revoked"
+		cert = rcert.Certificate
+	}
+
+	if status == "" {
```
cpu: I think it would be prudent to also check if `cert` is `nil`:
```suggestion
	if status == "" || cert == nil {
```

wfe/wfe.go
```diff
@@ -329,6 +331,54 @@ func (wfe *WebFrontEndImpl) handleKey(
...
+		response.WriteHeader(http.StatusNotFound)
+		return
+	}
+	result := make(map[string]interface{})
```
cpu: In this scenario I'd generally create an anonymous type instead of a map of `interface{}`. It gives you slightly better type safety and I think is a little bit more idiomatic for Go.

Something like:
```
result := struct{
  Status string
  Serial string
  Certificate sttring
} {
  Status: status,
  Serial: serial.Text(16),
  Certificate: string(cert.PEM()),
}
....
```

README.md
cpu: One other note: In your PR desc's example output it doesn't include the `Reason` or `RevokedAt` fields in the revoked certificate response output.

README.md
```diff
@@ -312,6 +312,17 @@ request to `https://localhost:15000/roots/0`, `https://localhost:15000/root-keys
 etc. These endpoints also send `Link` HTTP headers for all alternative root and
 intermediate certificates and keys.
 
+#### Certificate Status
+
+The certificate (in PEM format) and its revocation status can be queried by doing
+a `GET` request to `https://localhost:15000/cert-status-by-serial/<serial>`, where
+`<serial>` is the hexadecimal representation of the certificate's serial number.
```
cpu:
```suggestion
`<serial>` is the hexadecimal representation of the certificate's serial number (no `0x` prefix).
```

README.md
felixfontein: That's much better! Thanks!

README.md
felixfontein: I haven't updated the PR description with the later examples yet. Will do that.

core/types.go
felixfontein: Done.

db/memorystore.go
felixfontein: Thanks! Updated.

README.md
felixfontein: Updated!

wfe/wfe.go
felixfontein: Good point!

wfe/wfe.go
felixfontein: Also, usually the set of revoked certs is much smaller than the set of non-revoked certs, so this makes even more sense.

wfe/wfe.go

wfe/wfe.go
felixfontein: Done! Thanks for this as well.

Adding `Reason` and `RevokedAt` was more tricky; it turned out that `json:"omitempty"` doesn't output anything ever, and I needed to use `json:",omitempty"`. The comma turned out to be very important :)

felixfontein:
@cpu thanks for your feedback! I hope I've adjusted everything correctly. At least it still works in my tests :)

wfe/wfe.go
cpu: Doh! I'm 90% sure I've fallen into that trap before myself. I conveniently left out those fields in my suggested code :laughing: I'm glad you figured it out!

README.md
```diff
@@ -312,6 +312,34 @@ request to `https://localhost:15000/roots/0`, `https://localhost:15000/root-keys
...
+#### Certificate Status
+
+The certificate (in PEM format) and its revocation status can be queried by doing
+a `GET` request to `https://localhost:15000/cert-status-by-serial/<serial>`, where
+`<serial>` is the hexadecimal representation of the certificate's serial number (no `0x` prefix).
+It can be obtained via:
+
+    openssl x509 -in cert.pem -noout -serial | cut -d= -f2
+
+The endpoint returns the information as a JSON:
```
cpu: minor wording nit
```suggestion
The endpoint returns the information as a JSON object:
```

README.md
```diff
@@ -312,6 +312,34 @@ request to `https://localhost:15000/roots/0`, `https://localhost:15000/root-keys
 etc. These endpoints also send `Link` HTTP headers for all alternative root and
 intermediate certificates and keys.
 
+#### Certificate Status
+
+The certificate (in PEM format) and its revocation status can be queried by doing
```
cpu: minor wording nit
```suggestion
The certificate (in PEM format) and its revocation status can be queried by sending
```

wfe/wfe.go
```diff
@@ -329,6 +331,60 @@ func (wfe *WebFrontEndImpl) handleKey(
...
+		}
+		result.RevokedAt = rcert.RevokedAt.UTC().String()
+	}
+
+	resultJSON, err := marshalIndent(result)
+	if err != nil {
+		response.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	response.Header().Set("Content-Type", "application/json; charset=utf-8")
```
cpu: I think this line and the following two can be replaced with one call to the `webFrontEndImpl.writeJSONResponse` function. In this case I think it would be fair to throw away the potential error return, there isn't much to do with it.

wfe/wfe.go
```diff
@@ -329,6 +331,60 @@ func (wfe *WebFrontEndImpl) handleKey(
...
+		Certificate: string(cert.PEM()),
+	}
+	if rcert != nil {
+		if rcert.Reason != nil {
+			result.Reason = rcert.Reason
+		}
+		result.RevokedAt = rcert.RevokedAt.UTC().String()
+	}
+
+	resultJSON, err := marshalIndent(result)
```
cpu: I think this line and most of the logic that follows can be replaced with one call to the `webFrontEndImpl.writeJSONResponse` function, like:

```
_ = wfe.writeJSONResponse(response, http.StatusOK, result)
```

In this case I think it would be fair to throw away the potential error return, there isn't much to do with it.

felixfontein: Changed. I've decided to keep the `http.StatusInternalServerError` return in case serialization fails. It's better to indicate some kind of error to the client, that makes handling errors for callers much easier. A empty 200/204 response (or whatever Go will send by default) makes debugging and client-side error handling more complex.

cpu: Good points, that works for me :+1:

cpu:
Thanks again @felixfontein. My goal is to cut a new Pebble release that will have this commit and some of the other new work next week (hopefully Monday).

felixfontein:
@cpu thanks again for reviewing and merging! :)

cpu:
> My goal is to cut a new Pebble release that will have this commit and some of the other new work next week (hopefully Monday).

https://github.com/letsencrypt/pebble/releases/tag/v2.2.0 :tada:

PR #254, opened 2019-07-16
Remove `Link` headers for management interface

### LINKED ISSUE #253: Remove `Link` headers for management interface.
As noticed by @felixfontein :
> all requests to the management interface have a Link header for the directory included, but with the wrong port (that of the management interface).

There should be no `Link` header in responses to requests to the management interface.

felixfontein:
Fixes #253.

I'm not sure that this is the best implementation. Another way would be to add a boolean argument (I don't think this is good, though), or to have a more generic function which is used by `HandleFunc` and `HandleManagementFunc`.

Any opinions on this?

### DIFF
--- a/wfe/wfe.go
+++ b/wfe/wfe.go
@@ -224,42 +224,8 @@ func (wfe *WebFrontEndImpl) HandleFunc(
 func (wfe *WebFrontEndImpl) HandleManagementFunc(
 	mux *http.ServeMux,
 	pattern string,
-	handler wfeHandlerFunc,
-	methods ...string) {
-
-	methodsMap := make(map[string]bool)
-	for _, m := range methods {
-		methodsMap[m] = true
-	}
-
-	if methodsMap["GET"] && !methodsMap["HEAD"] {
-		// Allow HEAD for any resource that allows GET
-		methods = append(methods, "HEAD")
-		methodsMap["HEAD"] = true
-	}
-
-	methodsStr := strings.Join(methods, ", ")
-	defaultHandler := http.StripPrefix(pattern,
-		&topHandler{
-			wfe: wfeHandlerFunc(func(ctx context.Context, response http.ResponseWriter, request *http.Request) {
-				addNoCacheHeader(response)
-
-				if !methodsMap[request.Method] {
-					response.Header().Set("Allow", methodsStr)
-					wfe.sendError(acme.MethodNotAllowed(), response)
-					return
-				}
-
-				wfe.log.Printf("%s %s (management) -> calling handler()\n", request.Method, pattern)
-
-				// TODO(@cpu): Configurable request timeout
-				timeout := 1 * time.Minute
-				ctx, cancel := context.WithTimeout(ctx, timeout)
-				handler(ctx, response, request)
-				cancel()
-			},
-			)})
-	mux.Handle(pattern, defaultHandler)
+	handler wfeHandlerFunc) {
+	mux.Handle(pattern, http.StripPrefix(pattern, handler))
 }
 
 func (wfe *WebFrontEndImpl) sendError(prob *acme.ProblemDetails, response http.ResponseWriter) {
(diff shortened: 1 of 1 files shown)


wfe/wfe.go
```diff
@@ -349,10 +390,10 @@ func (wfe *WebFrontEndImpl) Handler() http.Handler {
 func (wfe *WebFrontEndImpl) ManagementHandler() http.Handler {
 	m := http.NewServeMux()
 	// GET only handlers
-	wfe.HandleFunc(m, RootCertPath, wfe.handleCert(wfe.ca.GetRootCert, RootCertPath), "GET")
-	wfe.HandleFunc(m, rootKeyPath, wfe.handleKey(wfe.ca.GetRootKey, rootKeyPath), "GET")
-	wfe.HandleFunc(m, intermediateCertPath, wfe.handleCert(wfe.ca.GetIntermediateCert, intermediateCertPath), "GET")
-	wfe.HandleFunc(m, intermediateKeyPath, wfe.handleKey(wfe.ca.GetIntermediateKey, intermediateKeyPath), "GET")
+	wfe.HandleManagementFunc(m, RootCertPath, wfe.handleCert(wfe.ca.GetRootCert, RootCertPath), "GET")
```
jsha: I think you could probably remove `HandleManagementFunc` and just call m.Handle directly:

```suggestion
	m.Handle(RootCertPath, wfe.handleCert(wfe.ca.GetRootCert, RootCertPath))
```

Most of the stuff that's in HandleFunc (and your copy of it in HandleManagementFunc) is specific to quirks of ACME, and we don't really need it in Pebble. For that matter, we probably don't even need most of it for ACME - for instance, in RFC8555 everything is a POST except for new-nonce and directory; we could clean up the code that deals with tying things to specific HTTP methods.

felixfontein: It would have to be
```suggestion
	wfe.HandleManagementFunc(m, RootCertPath, wfe.handleCert(wfe.ca.GetRootCert, RootCertPath), "GET")
	m.Handle(RootCertPath, http.StripPrefix(RootCertPath, wfeHandlerFunc(wfe.handleCert(wfe.ca.GetRootCert, RootCertPath))))
```
`StripPrefix` is needed since otherwise `RootCertPath` needs to cut off the path prefix itself, and the `wfeHandlerFunc()` cast is needed since otherwise there's a compile error. While this works, it is pretty explicit (and that has to be repeated for every management endpoint). Having a small `wfe.HandleManagementFunc` like
```.go
func (wfe *WebFrontEndImpl) HandleManagementFunc(
	mux *http.ServeMux,
	pattern string,
	handler wfeHandlerFunc) {
	mux.Handle(pattern, http.StripPrefix(pattern, handler))
}
```
makes this already much nicer.

One disadvantage this approach has is that the method checking is gone, i.e. the handlers reply to all HTTP methods independently of HTTP semantics. Also, there's no longer a no-cache header automatically added.

jsha: Your simplified wrapper function looks nicer. I'm not too worried about losing the method checking and no-cache header.

wfe/wfe.go
```diff
@@ -221,6 +221,13 @@ func (wfe *WebFrontEndImpl) HandleFunc(
 	mux.Handle(pattern, defaultHandler)
 }
 
+func (wfe *WebFrontEndImpl) HandleManagementFunc(
+	mux *http.ServeMux,
+	pattern string,
+	handler wfeHandlerFunc) {
```
golangcibot: `handler` can be `net/http.Handler` (from `interfacer`)

wfe/wfe.go
felixfontein: I've updated the code with the simplified wrapper.

wfe/wfe.go
felixfontein: When I do that, I get
```
wfe/wfe.go:359:58: cannot use wfe.handleCert(wfe.ca.GetRootCert, RootCertPath) (type func(context.Context, http.ResponseWriter, *http.Request)) as type http.Handler in argument to wfe.HandleManagementFunc:
	func(context.Context, http.ResponseWriter, *http.Request) does not implement http.Handler (missing ServeHTTP method)
wfe/wfe.go:360:56: cannot use wfe.handleKey(wfe.ca.GetRootKey, rootKeyPath) (type func(context.Context, http.ResponseWriter, *http.Request)) as type http.Handler in argument to wfe.HandleManagementFunc:
	func(context.Context, http.ResponseWriter, *http.Request) does not implement http.Handler (missing ServeHTTP method)
wfe/wfe.go:361:66: cannot use wfe.handleCert(wfe.ca.GetIntermediateCert, intermediateCertPath) (type func(context.Context, http.ResponseWriter, *http.Request)) as type http.Handler in argument to wfe.HandleManagementFunc:
	func(context.Context, http.ResponseWriter, *http.Request) does not implement http.Handler (missing ServeHTTP method)
wfe/wfe.go:362:64: cannot use wfe.handleKey(wfe.ca.GetIntermediateKey, intermediateKeyPath) (type func(context.Context, http.ResponseWriter, *http.Request)) as type http.Handler in argument to wfe.HandleManagementFunc:
	func(context.Context, http.ResponseWriter, *http.Request) does not implement http.Handler (missing ServeHTTP method)
```

jsha: Yeah I *think* golangcibot may be confused here and can probably be ignored.

That said, Travis jobs are failing, so let's see whether golangcibot still reports this once those tests are fixed.

felixfontein: Both Travis and AppVeyor fail because of
```
$ golangci-lint run
wfe/wfe.go:227:2: `handler` can be `net/http.Handler` (interfacer)
	handler wfeHandlerFunc) {
	^
```
So exactly the same "problem". Every other test passed.

jsha: Can you add `interfacer` to the `disable` list in .golangci.yml?

felixfontein: That sounds too general for me. I've first tried to disable this particular test for this line (according to https://github.com/golangci/golangci-lint#nolint). I'm not sure whether it's the correct line, but since GolangCI seems already to be happy, I guess it was the correct one :)

felixfontein:
@jsha thanks for handling this!

PR #255, opened 2019-07-22
Fix custom resolver implementation for Windows

adferrand:
Pebble accepts a `--dnsserver` flag, that allows to select a custom DNS resolver to use when validating the ACME challenges. This option is critical to make the certbot integration tests work in particular.

Current implementation is to set the `net.DefaultResolver` on a new `net.Resolver` instance with a custom `Dial` property to consume the value from `--dnsserver`. This allows to transparently resolve any host using this custom DNS resolver from a pure-Go implementation of the DNS resolution API.

Sadly this approach does not work on Windows, and the `--dnsserver` has no effect on how resolution is done, and it is always the OS mechanism that is used.

One can reveal this behavior by running the following piece of code on Linux and on Windows:
```go
// test.go
package main

import (
	"context"
	"fmt"
	"net"
)

func main() {
	resolver := &net.Resolver{
		PreferGo: true,
		Dial: func(ctx context.Context, _, _ string) (net.Conn, error) {
			d := net.Dialer{}
			return d.DialContext(ctx, "udp", "4.3.2.1:404")
		},
	}
	net.DefaultResolver = resolver
	ctx, cancelfunc := context.WithTimeout(context.Background(), 30)
	defer cancelfunc()
	_, err := resolver.LookupTXT(ctx, "stupid.entry.net")

	fmt.Printf("Error is: %s\n", err)
}
```

This piece of code tries to resolve a non-existent domain on a non-existent DNS server as IP `4.3.2.1:404`.

On Linux, you will get the following error:
```bash
Error is: lookup stupid.entry.net on 127.0.0.53:53: dial udp 4.3.2.1:404: i/o timeout
```

That indicates that the system tried to reach the non-existent DNS server, and get back a timeout exception on it.

However on Windows you will get:
```bash
Error is: lookup stupid.entry.net: dnsquery: DNS name does not exist.
```

This indicates that the system ignored the dummy DNS server address, contacted the OS DNS resolver, that responded that the DNS name does not exist.

One can see also the reason for this behavior on Windows on the `net` godoc, https://godoc.org/net, in particular this line in the module introduction:
```
On Windows, the resolver always uses C library functions, such as GetAddrInfo and DnsQuery.
```

In fact, the pure Go DNS resolver is not applicable on Windows, the OS DNS resolver will be used, ignoring any customization.

Several relevant discussions, in particular a proposal (not developed yet) to make the pure Go DNS resolver available on Windows:
* https://github.com/golang/go/issues/29621
* https://github.com/golang/go/issues/33097
* https://github.com/golang/go/issues/33086

To fix this, this PR makes Pebble switch to a different logic:
* if `-dnsserver` is not set, use the default API to resolve the names
* if `-dnsserver` is set, use a dedicated DNS client, to avoid to use the OS one both on Linux and Windows

The DNS client is https://github.com/miekg/dns, a highly used and supported DNS library.

With these modifications, integrations tests on Certbot are running correctly both on Linux and Windows.

### DIFF
--- a/va/va.go
+++ b/va/va.go
@@ -21,6 +21,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/miekg/dns"
+
 	"github.com/letsencrypt/challtestsrv"
 	"github.com/letsencrypt/pebble/acme"
 	"github.com/letsencrypt/pebble/core"
@@ -93,29 +95,39 @@ type vaTask struct {
 }
 
 type VAImpl struct {
-	log         *log.Logger
-	httpPort    int
-	tlsPort     int
-	tasks       chan *vaTask
-	sleep       bool
-	sleepTime   int
-	alwaysValid bool
-	strict      bool
+	log                *log.Logger
+	httpPort           int
+	tlsPort            int
+	tasks              chan *vaTask
+	sleep              bool
+	sleepTime          int
+	alwaysValid        bool
+	strict             bool
+	customResolverAddr string
 }
 
 func New(
 	log *log.Logger,
 	httpPort, tlsPort int,
-	strict bool) *VAImpl {
+	strict bool, customResolverAddr string) *VAImpl {
 	va := &VAImpl{
-		log:       log,
-		httpPort:  httpPort,
-		tlsPort:   tlsPort,
-		tasks:     make(chan *vaTask, taskQueueSize),
-		sleep:     true,
-		sleepTime: defaultSleepTime,
-		strict:    strict,
+		log:                log,
+		httpPort:           httpPort,
+		tlsPort:            tlsPort,
+		tasks:              make(chan *vaTask, taskQueueSize),
+		sleep:              true,
+		sleepTime:          defaultSleepTime,
+		strict:             strict,
+		customResolverAddr: customResolverAddr,
 	}
+
+	if customResolverAddr != "" {
+		va.log.Printf("Using custom DNS resolver for ACME challenges: %s", customResolverAddr)
+	} else {
+		va.log.Print("Using system DNS resolver for ACME challenges")
+	}
+
+	va.log.Print(va.resolveIP("8.8.8.8"))
 
 	// Read the PEBBLE_VA_NOSLEEP environment variable string
 	noSleep := os.Getenv(noSleepEnvVar)
@@ -299,10 +311,7 @@ func (va VAImpl) validateDNS01(task *vaTask) *core.ValidationRecord {
 		ValidatedAt: time.Now(),
 	}
 
-	ctx, cancelfunc := context.WithTimeout(context.Background(), validationTimeout)
-	defer cancelfunc()
-
-	txts, err := net.DefaultResolver.LookupTXT(ctx, challengeSubdomain)
+	txts, err := va.getTXTEntry(challengeSubdomain)
 	if err != nil {
 		result.Error = acme.UnauthorizedProblem(fmt.Sprintf("Error retrieving TXT records for DNS challenge (%q)", err))
 		return result
@@ -490,6 +499,13 @@ func (va VAImpl) fetchHTTP(identifier string, token string) ([]byte, string, *ac
 	httpRequest.Header.Set("User-Agent", userAgent())
 	httpRequest.Header.Set("Accept", "*/*")
 
+	addrs, err := va.resolveIP(identifier)
+
+	if err != nil || len(addrs) == 0 {
+		return nil, url.String(), acme.MalformedProblem(
+			fmt.Sprintf("Could not resolve URL %q", url.String()))
+	}
+
 	transport := &http.Transport{
 		// We don't expect to make multiple requests to a client, so close
 		// connection immediately.
@@ -500,6 +516,12 @@ func (va VAImpl) fetchHTTP(identifier string, token string) ([]byte, string, *ac
 		// to an HTTPS host.
 		TLSClientConfig: &tls.Config{
 			InsecureSkipVerify: true,
+		},
+
+		// Control specifically which IP will be used for this request
+		DialContext: func(ctx context.Context, network, addr string) (net.Conn, error) {
+			dialer := &net.Dialer{}
+			return dialer.DialContext(ctx, network, net.JoinHostPort(addrs[0], portString))
 		},
 	}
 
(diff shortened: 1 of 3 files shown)


cpu:
@adferrand Thanks for the detailed summary of the problem and a proposed PR :+1: It's a shame that the language differs in behaviour across platforms in this way.

> The DNS client is https://github.com/miekg/dns, a highly used and supported DNS library.

In the past I resisted adding this dependency to Pebble. Not because of any concerns about quality (we use this dependency in both Boulder and the `challtestsrv`) but with the hopes we could keep the project as small and self-contained as possible. 

That said, it's already there transitively by way of `github.com/letsencrypt/challtestsrv` and it's clearly required for Windows support. I'm happy to change positions on this and accept the new dependency for those reasons. :+1:

va/va.go
```diff
@@ -93,30 +95,40 @@ type vaTask struct {
...
+		customResolverAddr: customResolverAddr,
+	}
+
+	if customResolverAddr != "" {
+		va.log.Printf("Using custom DNS resolver for ACME challenges: %s", customResolverAddr)
+	} else {
+		va.log.Print("Using system DNS resolver for ACME challenges")
 	}
 
+	va.log.Print(va.resolveIP("8.8.8.8"))
```
cpu: Is this by chance some leftover debugging cruft?

va/va.go
```diff
@@ -534,6 +556,80 @@ func (va VAImpl) fetchHTTP(identifier string, token string) ([]byte, string, *ac
...
+func (va VAImpl) getTXTEntry(name string) ([]string, error) {
+	ctx, cancelfunc := context.WithTimeout(context.Background(), validationTimeout)
+	defer cancelfunc()
+
+	if va.customResolverAddr == "" {
+		return net.DefaultResolver.LookupTXT(ctx, name)
+	}
+
+	txts := []string{}
+	client := new(dns.Client)
```
cpu: I think it would make sense to hoist the `*dns.Client` to a field on the `VAImpl` that may be `nil` (in the case where `va.customResolverAddr` isn't in use). That way it could be constructed once instead of per-lookup.

va/va.go
```diff
@@ -534,6 +556,80 @@ func (va VAImpl) fetchHTTP(identifier string, token string) ([]byte, string, *ac
 	return body, url.String(), nil
 }
 
+func (va VAImpl) getTXTEntry(name string) ([]string, error) {
```
cpu: Can you add a brief godoc comment to `getTXTEntry` that describes what it does? Something like:

```
// getTXTEntry fetches TXT entries for the given domain name using the recursive resolver located at `va.customResolverAddr`, or the default system resolver if no custom resolver addr is specified
```

va/va.go
```diff
@@ -534,6 +556,80 @@ func (va VAImpl) fetchHTTP(identifier string, token string) ([]byte, string, *ac
...
+	for _, record := range in.Answer {
+		if t, ok := record.(*dns.TXT); ok {
+			txts = append(txts, t.Txt...)
+		}
+	}
+
+	return txts, nil
+}
+
+func (va VAImpl) resolveIP(name string) ([]string, error) {
```
cpu: ditto for this function and a brief godoc comment

va/va.go
```diff
@@ -534,6 +556,80 @@ func (va VAImpl) fetchHTTP(identifier string, token string) ([]byte, string, *ac
...
+
+	addrs := []string{}
+	parsed := net.ParseIP(name)
+
+	if parsed != nil {
+		addrs = append(addrs, name)
+		return addrs, nil
+	}
+
+	client := new(dns.Client)
```
cpu: ditto for using a per-VA client here instead of constructing it per-query.

va/va.go
```diff
@@ -534,6 +556,80 @@ func (va VAImpl) fetchHTTP(identifier string, token string) ([]byte, string, *ac
...
+func (va VAImpl) resolveIP(name string) ([]string, error) {
+	ctx, cancelfunc := context.WithTimeout(context.Background(), validationTimeout)
+	defer cancelfunc()
+
+	if va.customResolverAddr == "" {
+		return net.DefaultResolver.LookupHost(ctx, name)
+	}
+
+	addrs := []string{}
+	parsed := net.ParseIP(name)
```
cpu: Is this to handle the case where `resolveIP` is called with `name` as an IP address encoded as a string? I'm not sure I understand this parsing check.

va/va.go
```diff
@@ -534,6 +556,80 @@ func (va VAImpl) fetchHTTP(identifier string, token string) ([]byte, string, *ac
...
+		return addrs, errAAAA
+	}
+
+	for _, record := range inA.Answer {
+		if t, ok := record.(*dns.A); ok {
+			addrs = append(addrs, t.A.String())
+		}
+	}
+
+	for _, record := range inAAAA.Answer {
```
cpu: Are you open to changing the order of these two `ranges`? I think the AAAA records should appear first in the `addrs` slice to better match typical "happy eyeballs" behaviour.

The godoc comment for this function should also mention the order of results w.r.t. dual-homed names.

adferrand:
Also note that as soon as a new version of Go allows to use the native pure Go resolver, I am willing to make a new PR to revert this one.

va/va.go
adferrand: Yes it is. Remove it.

va/va.go
adferrand: Done.

va/va.go
adferrand: Done.

va/va.go
adferrand: Done.

va/va.go
adferrand: Done.

va/va.go
adferrand: It is for this case indeed. I add an inline comment.

va/va.go
```diff
@@ -93,28 +95,38 @@ type vaTask struct {
...
+	log                *log.Logger
+	httpPort           int
+	tlsPort            int
+	tasks              chan *vaTask
+	sleep              bool
+	sleepTime          int
+	alwaysValid        bool
+	strict             bool
+	customResolverAddr string
+	dnsClient 		   *dns.Client
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
	dnsClient          *dns.Client
```

va/va.go
adferrand: Indeed, I think it would be beneficial for the sake of the future of IPv6. However to work properly, maybe it is needed to make some conditional appends here. I mean that on a non-IPv6 capable system, this function should not return any IPv6 addresses (and similarly for IPv4) right ?

If so, @cpu do you have an idea of how I could test the availability of each network stack on the current system, to conditionally trigger one, or both, of the two loops ?

cpu: I think I would advocate an approach where we ignore the availability of the network stack. If a Pebble environment is configured with a DNS server that returns AAAA records for a name and the Pebble environment doesn't have IPv6 connectivity I think the solution is to remove the AAAA records from the mock data.

adferrand: That works for me. I switch the order of `AAAA` and `A` records here.

cpu: Great, thanks! I'll give this branch another pass shortly.

va/va.go
```diff
@@ -490,6 +499,13 @@ func (va VAImpl) fetchHTTP(identifier string, token string) ([]byte, string, *ac
 	httpRequest.Header.Set("User-Agent", userAgent())
 	httpRequest.Header.Set("Accept", "*/*")
 
+	addrs, err := va.resolveIP(identifier)
+
+	if err != nil || len(addrs) == 0 {
```
jsha: Let's break these out into two separate cases, so we can pass through the error from `resolveIP`.

va/va.go
```diff
@@ -534,6 +556,82 @@ func (va VAImpl) fetchHTTP(identifier string, token string) ([]byte, string, *ac
...
+	if va.customResolverAddr == "" {
+		return net.DefaultResolver.LookupTXT(ctx, name)
+	}
+
+	txts := []string{}
+	message := new(dns.Msg)
+	message.SetQuestion(dns.Fqdn(name), dns.TypeTXT)
+	in, _, err := va.dnsClient.ExchangeContext(ctx, message, va.customResolverAddr)
+
+	if err != nil || in.Rcode != dns.RcodeSuccess {
```
jsha: In the `in.Rcode != dns.RcodeSuccess` case, this will return a nil err. I think you need to split out the two error cases.

va/va.go
```diff
@@ -534,6 +556,82 @@ func (va VAImpl) fetchHTTP(identifier string, token string) ([]byte, string, *ac
...
+	for _, record := range in.Answer {
+		if t, ok := record.(*dns.TXT); ok {
+			txts = append(txts, t.Txt...)
+		}
+	}
+
+	return txts, nil
+}
+
+// resolveIP find all associated IPs to the given domain name using the recursive resolver located at
```
jsha:
```suggestion
// resolveIP find all IPs for the given domain name using the recursive resolver located at
```

va/va.go
jsha: Let's put this check at the top of the function then, since it short-circuits the rest.

Also, it would be clearer to make the return statement be `return []net.IP{parsed}`.

va/va.go
jsha: Could you change the return value to `[]net.IP, error`? That makes it a little clearer what is being returned.

va/va.go
```diff
@@ -534,6 +556,82 @@ func (va VAImpl) fetchHTTP(identifier string, token string) ([]byte, string, *ac
...
+
+	messageAAAA := new(dns.Msg)
+	messageAAAA.SetQuestion(dns.Fqdn(name), dns.TypeAAAA)
+	inAAAA, _, errAAAA := va.dnsClient.ExchangeContext(ctx, messageAAAA, va.customResolverAddr)
+	messageA := new(dns.Msg)
+	messageA.SetQuestion(dns.Fqdn(name), dns.TypeA)
+	inA, _, errA := va.dnsClient.ExchangeContext(ctx, messageA, va.customResolverAddr)
+
+	if errAAAA != nil {
+		return addrs, errAAAA
```
jsha: For the error cases, it's more idiomatic to say `return nil, errAAAA`.

Also, in this case since we call each ExchangeContext serially, you don't need to define separate `errAAAA` and `errA`; you can just assign to `err` and check it directly after the call.

va/va.go
adferrand: Done.

va/va.go
adferrand: Done.

va/va.go
adferrand: Done.

va/va.go
```diff
@@ -490,6 +499,18 @@ func (va VAImpl) fetchHTTP(identifier string, token string) ([]byte, string, *ac
 	httpRequest.Header.Set("User-Agent", userAgent())
 	httpRequest.Header.Set("Accept", "*/*")
 
+	addrs, err := va.resolveIP(identifier)
+
+	if err != nil {
+		return nil, url.String(), acme.MalformedProblem(
+			fmt.Sprintf("Error occured while resolving URL %q: %q", url.String(), err))
```
golangcibot: `occured` is a misspelling of `occurred` (from `misspell`)

adferrand:
Ok @jsha, I improved the error handling thanks to your comments. It is true that for a newcomer in Go, the error handling in this language is quite disturbing at the beginning ^^

jsha:
Thanks for the PR, and for the excellent and thorough description of the problem and the solution!

PR #259, opened 2019-07-26
Implement reuse of valid authorizations.

alexzorin:
I am not sure whether the probabilistic aspect of this furthers the goals of Pebble or not. I am more than happy to remove it if that's the feedback. 

---

RFC8555 mentions (https://tools.ietf.org/html/rfc8555#section-7.1.3)
that clients should check the "status" field of an order to determine
whether they need to take any action.

This commit implements authorization reuse, such that any valid and
unexpired authz completed by the ACME account for that identifier has a
chance to be reused in new orders. The chance is controlled by the env
variable PEBBLE_WFE_AUTHZREUSE and it defaults to 50 (percent). The
chance is evaluated regardless of whether there is an eligible authz to
reuse or not.

### DIFF
--- a/db/memorystore.go
+++ b/db/memorystore.go
@@ -215,8 +215,8 @@ func (m *MemoryStore) GetAuthorizationByID(id string) *core.Authorization {
 	return m.authorizationsByID[id]
 }
 
-// FindValidAuthorization locates an existing valid authorization for an identifier
-// from the ACME account matching accountID.
+// FindValidAuthorization fetches the first, if any, valid and unexpired authorization for the
+// provided identifier, from the ACME account matching accountID.
 func (m *MemoryStore) FindValidAuthorization(accountID string, identifier acme.Identifier) *core.Authorization {
 	m.RLock()
 	defer m.RUnlock()
--- a/db/memorystore.go
+++ b/db/memorystore.go
@@ -215,13 +215,15 @@ func (m *MemoryStore) GetAuthorizationByID(id string) *core.Authorization {
 	return m.authorizationsByID[id]
 }
 
+// FindValidAuthorization locates an existing valid authorization for an identifier
+// from the ACME account matching accountID.
 func (m *MemoryStore) FindValidAuthorization(accountID string, identifier acme.Identifier) *core.Authorization {
 	m.RLock()
 	defer m.RUnlock()
 	for _, authz := range m.authorizationsByID {
-		if authz.Identifier.Type == identifier.Type && authz.Identifier.Value == identifier.Value &&
-			authz.Order.AccountID == accountID && authz.ExpiresDate.After(time.Now()) &&
-			authz.Status == acme.StatusValid {
+		if authz.Status == acme.StatusValid && identifier.Equals(authz.Identifier) &&
+			authz.Order != nil && authz.Order.AccountID == accountID &&
+			authz.ExpiresDate.After(time.Now()) {
 			return authz
 		}
 	}
--- a/db/memorystore.go
+++ b/db/memorystore.go
@@ -9,9 +9,11 @@ import (
 	"reflect"
 	"strconv"
 	"sync"
+	"time"
 
 	"gopkg.in/square/go-jose.v2"
 
+	"github.com/letsencrypt/pebble/acme"
 	"github.com/letsencrypt/pebble/core"
 )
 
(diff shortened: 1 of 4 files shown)


cpu:
> I am not sure whether the probabilistic aspect of this furthers the goals of Pebble or not. I am more than happy to remove it if that's the feedback.

I think the probabilistic approach is the perfect solution here :+1: 

Historically I was hesitant about the idea of building authz reuse into Pebble because I was wary of clients coming to depend on it and breaking with other RFC 8555 implementations that don't bother with this performance optimization. I thought about an on/off toggle that would default to off but that precludes easy testing of authz reuse and might result in the opposite behaviour: code that works without authz reuse but breaks with it :dizzy_face: It seems to me this approach strikes a good balance.

db/memorystore.go
```diff
@@ -213,6 +215,19 @@ func (m *MemoryStore) GetAuthorizationByID(id string) *core.Authorization {
 	return m.authorizationsByID[id]
 }
 
+func (m *MemoryStore) FindValidAuthorization(accountID string, identifier acme.Identifier) *core.Authorization {
+	m.RLock()
+	defer m.RUnlock()
+	for _, authz := range m.authorizationsByID {
+		if authz.Identifier.Type == identifier.Type && authz.Identifier.Value == identifier.Value &&
```
cpu: How do you feel about implementing `func (id acme.Identifier) Equals(other acme.Identifier)` in the `acme `package and using that here?

cpu: But to be clear I think the `authz.ExpiresDate.After` should remain outside of the `Equals` impl.

db/memorystore.go
```diff
@@ -213,6 +215,19 @@ func (m *MemoryStore) GetAuthorizationByID(id string) *core.Authorization {
 	return m.authorizationsByID[id]
 }
 
+func (m *MemoryStore) FindValidAuthorization(accountID string, identifier acme.Identifier) *core.Authorization {
```
cpu: Could you add a brief Godoc comment to this function?

wfe/wfe.go
```diff
@@ -92,6 +92,16 @@ const (
...
 	aACompromiseRevocationReason = 10
+
+	// authzReuseEnvVar defines an environment variable name used to provide a
+	// percentage value for how often Pebble should try to reuse valid authorizations
+	// for each identifier in an order. The percentage is independent of whether a
+	// valid authorization exists or not for each identifier in an order.
+	authzReuseEnvVar = "PEBBLE_WFE_AUTHZREUSE"
+
+	// The default value when PEBBLE_WFE_AUTHZREUSE is not set, how often to try
+	// and reuse valid authorizations?
```
cpu: small nit: the trailing `?` reads to me as uncertainty :-)
```suggestion
	// and reuse valid authorizations.
```

wfe/wfe.go
```diff
@@ -154,14 +165,24 @@ func New(
 	}
 	log.Printf("Configured to reject %d%% of good nonces", nonceErrPercent)
 
+	// Get authz reuse percent from the environment
+	authzReusePercent := defaultAuthzReuse
+	if val, err := strconv.ParseInt(os.Getenv(authzReuseEnvVar), 10, 0); err == nil &&
+		val >= 0 && val <= 100 {
+		authzReusePercent = int(val)
+	}
+	log.Printf("Configured to attempt authz reuse for each identifiers %d%% of the time",
```
cpu: I think this is a small typo
```suggestion
	log.Printf("Configured to attempt authz reuse for each identifier %d%% of the time",
```

alexzorin:
Should `PEBBLE_WFE_AUTHZREUSE` potentially be called `PEBBLE_RA_AUTHZREUSE`? Or since there is not really any RA, just `PEBBLE_AUTHZREUSE`?

jsha:
Hm, it does look like we've followed a pattern of naming the component, e.g. `PEBBLE_VA_NOSLEEP` and `PEBBLE_WFE_NONCEREJECT`. But on thinking about it some more I think it's better to not name the component, since that's not necessarily part of the mental model of the user. So let's go with `PEBBLE_AUTHZREUSE`.

alexzorin:
Don't think I can modify the env var in the original commit message without a force push, but otherwise I think I'm done for now. Not sure what the build failure is about, I didn't touch Go modules.

jsha:
I'm guessing that AppVeyor failure was a transient failure to fetch one of the dependencies. Hopefully the next run will pass.

PR #260, opened 2019-07-27
Implementing list orders endpoint (and adding `orders` to account object)

### LINKED ISSUE #84: Required 'orders' key missing in account objects

felixfontein:
Fixes #84.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,10 +54,9 @@ clients are not hardcoding URLs.)
 ## Limitations
 
 Pebble is missing some ACME features (PRs are welcome!). It does not presently
-support the "orders" field of account objects, subproblems, pre-authorization or
-external account binding. Pebble does not support revoking a certificate issued
-by a different ACME account by proving authorization of all of the certificate's
-domains.
+support subproblems, pre-authorization or external account binding. Pebble does
+not support revoking a certificate issued by a different ACME account by proving
+authorization of all of the certificate's domains.
 
 Pebble does not perform all of the same input validation as Boulder. Some domain
 names that would be rejected by Boulder/Let's Encrypt may work with Pebble.
--- a/README.md
+++ b/README.md
@@ -365,3 +365,14 @@ to retrieve the OCSP status of a certificate, run Pebble with a `pebble-config.j
 ```
   "ocspResponderURL": "http://127.0.0.1:4002",
 ```
+
+### Listing orders
+
+[As opposed to Boulder](https://github.com/letsencrypt/boulder/issues/3335), Pebble has support
+for enumerating all orders for an ACME account object according to
+[RFC 8555, Section 7.1.2](https://tools.ietf.org/html/rfc8555#section-7.1.2.1). By default, three
+orders are returned per page, to make it easy to test pagination. This number can be modified by
+setting the `PEBBLE_WFE_ORDERS_PER_PAGE` environment variable to a positive integer. For example,
+to have 15 orders per page, run
+
+`PEBBLE_WFE_ORDERS_PER_PAGE=15 pebble`
--- a/README.md
+++ b/README.md
@@ -367,8 +367,7 @@ to retrieve the OCSP status of a certificate, run Pebble with a `pebble-config.j
 
 ### Listing orders
 
-[As opposed to Boulder](https://github.com/letsencrypt/boulder/issues/3335), Pebble has support
-for enumerating all orders for an ACME account object according to
+Pebble has support for enumerating all orders for an ACME account object according to
 [RFC 8555, Section 7.1.2](https://tools.ietf.org/html/rfc8555#section-7.1.2.1). By default, three
 orders are returned per page, to make it easy to test pagination. This number can be modified by
 setting the `PEBBLE_WFE_ORDERS_PER_PAGE` environment variable to a positive integer. For example,
--- a/db/memorystore.go
+++ b/db/memorystore.go
@@ -42,7 +42,8 @@ type MemoryStore struct {
 	// key bytes.
 	accountsByKeyID map[string]*core.Account
 
-	ordersByID map[string]*core.Order
+	ordersByID        map[string]*core.Order
+	ordersByAccountID map[string][]*core.Order
 
 	authorizationsByID map[string]*core.Authorization
 
@@ -58,6 +59,7 @@ func NewMemoryStore() *MemoryStore {
 		accountsByID:            make(map[string]*core.Account),
 		accountsByKeyID:         make(map[string]*core.Account),
 		ordersByID:              make(map[string]*core.Order),
+		ordersByAccountID:       make(map[string][]*core.Order),
 		authorizationsByID:      make(map[string]*core.Authorization),
 		challengesByID:          make(map[string]*core.Challenge),
 		certificatesByID:        make(map[string]*core.Certificate),
@@ -170,6 +172,13 @@ func (m *MemoryStore) AddOrder(order *core.Order) (int, error) {
 		return 0, fmt.Errorf("order %q already exists", orderID)
 	}
 
+	var ordersByAccountID []*core.Order
+	var present bool
+	if ordersByAccountID, present = m.ordersByAccountID[order.AccountID]; !present {
+		ordersByAccountID = make([]*core.Order, 0)
+	}
+	m.ordersByAccountID[order.AccountID] = append(ordersByAccountID, order)
+
 	m.ordersByID[orderID] = order
 	return len(m.ordersByID), nil
 }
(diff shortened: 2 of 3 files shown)


wfe/wfe.go
```diff
@@ -950,6 +977,76 @@ func (wfe *WebFrontEndImpl) UpdateAccount(
...
+	}
+
+	if end < len(filteredOrders) {
+		nextPageURL := wfe.relativeEndpoint(request, fmt.Sprintf("%s%s/page/%d", ordersPath, existingAcct.ID, page+2))
+		response.Header().Add("Link", link(nextPageURL, "next"))
+	}
+
+	err := wfe.writeJSONResponse(response, http.StatusOK, result)
+	if err != nil {
+		wfe.sendError(acme.InternalErrorProblem("Error marshalling orders list"), response)
```
golangcibot: `marshalling` is a misspelling of `marshaling` (from `misspell`)

felixfontein: Both spellings are valid (https://en.wikipedia.org/wiki/Marshalling_(computer_science)). Also, the rest of the code already uses the other spelling. On the other hand, the Golang team decided to stick to marshaling with one l (see https://github.com/golang/go/commit/d8264de8683dac99ffbbbc1f46415e627b73c9ed). Any opinions?

jsha: In matters of style, I would say "obey the bot." :-) It's like Rob Pike says: gofmt's style is nobody's favorite, but gofmt is everybody's favorite.

A quick grep of the repo shows we use both spellings. I'd say: let's adopt the bot-friendly spelling in this PR, and then as a (very low-priority) followup edit the existing usages to the bot-friendly spelling.

felixfontein: I've fixed it for this PR. I'll create another PR fixing all other occurrences.

felixfontein:
Looks like AppVeyor has some transient problems here as well... (`Failed to connect to github.com port 443: Timed out`)

felixfontein:
I tested this with ansible/ansible#59697, which implements fetching orders for Ansible's `acme_account_info` module.

felixfontein:
Rebased to resolve conflicts with #259.

felixfontein:
@alexzorin good point! I've removed that part from the section.

README.md
```diff
@@ -365,3 +364,14 @@ to retrieve the OCSP status of a certificate, run Pebble with a `pebble-config.j
 ```
   "ocspResponderURL": "http://127.0.0.1:4002",
 ```
+
+### Listing orders
+
+[As opposed to Boulder](https://github.com/letsencrypt/boulder/issues/3335), Pebble has support
```
cpu: I think its probably not necessary to discuss Boulder here. It will be one more place to remember to update if/when we resolve Boulder issue 3335.

README.md
```diff
@@ -365,3 +364,14 @@ to retrieve the OCSP status of a certificate, run Pebble with a `pebble-config.j
...
+for enumerating all orders for an ACME account object according to
+[RFC 8555, Section 7.1.2](https://tools.ietf.org/html/rfc8555#section-7.1.2.1). By default, three
+orders are returned per page, to make it easy to test pagination. This number can be modified by
+setting the `PEBBLE_WFE_ORDERS_PER_PAGE` environment variable to a positive integer. For example,
```
cpu: Do you think there's an advantage to making this configurable? At first glance it feels unnecessary but I might be overlooking something that makes this desirable from a client developer perspective.

db/memorystore.go
```diff
@@ -170,6 +172,13 @@ func (m *MemoryStore) AddOrder(order *core.Order) (int, error) {
 		return 0, fmt.Errorf("order %q already exists", orderID)
 	}
 
+	var ordersByAccountID []*core.Order
```
cpu: I think you can get rid of these two explicit `var` declarations and use:

```
if ordersByAccountID, present := m.ordersByAccountID[order.AccountID]; present {
....
}
```

db/memorystore.go
```diff
@@ -170,6 +172,13 @@ func (m *MemoryStore) AddOrder(order *core.Order) (int, error) {
...
+	var present bool
+	if ordersByAccountID, present = m.ordersByAccountID[order.AccountID]; !present {
+		ordersByAccountID = make([]*core.Order, 0)
+	}
+	m.ordersByAccountID[order.AccountID] = append(ordersByAccountID, order)
```
cpu: I think this is appending to the wrong slice:

```suggestion
	m.ordersByAccountID[order.AccountID] = append(m.ordersByAccountID, order)
```

db/memorystore.go
```diff
@@ -170,6 +172,13 @@ func (m *MemoryStore) AddOrder(order *core.Order) (int, error) {
 		return 0, fmt.Errorf("order %q already exists", orderID)
 	}
 
+	var ordersByAccountID []*core.Order
+	var present bool
+	if ordersByAccountID, present = m.ordersByAccountID[order.AccountID]; !present {
```
cpu: I think you need to capture this `order.AccountID` at the same place the `orderID` is assigned in order to do it while the `order` is `RLock()`'d

db/memorystore.go
```diff
@@ -191,6 +200,25 @@ func (m *MemoryStore) GetOrderByID(id string) *core.Order {
...
+			if err != nil {
+				panic(err)
+			}
+			order.Lock()
+			defer order.Unlock()
+			order.Status = orderStatus
+		}
+		return orders
+	}
+	return make([]*core.Order, 0)
```
cpu: Is there a reason you preferred returning an empty slice instead of `nil` here?

README.md
felixfontein: I mainly wanted to make sure that people remember that just because Pebble can do this, it doesn't mean that Boulder will also return `orders` :) But I can remove it.

README.md
felixfontein: I've thought about this a bit; I think it would make sense to make it configurable, but I would probably allow all such options to be configurable (via the management interface). I thought it's probably best to come up with a uniform interface for updating (or also querying) configuration values, but I don't think it should happen in this PR.

db/memorystore.go
felixfontein: I tried that first, but then `ordersByAccountID` won't be available after the `if` block.

db/memorystore.go
felixfontein: Either it should be `m.ordersByAccountID[order.AccountID]`, or `ordersByAccountID`. I used the latter because the former might not be defined yet.

db/memorystore.go
felixfontein: Good point!

db/memorystore.go
felixfontein: Yes, so that the code calling `db.GetOrdersByAccountID` has a non-`nil` slice and doesn't have to special case for the case it is `nil`.

(It could be of course that this isn't necessary in Go. I'm not that familiar with all the details of Go :) )

db/memorystore.go
felixfontein: I've also moved the lock release before the `if`, so that the lock is released in any case.

cpu:
@felixfontein Just wanted to say I haven't forgotten about this PR, sorry for the delays getting back to it. I promise I won't let it linger too far into next week. Thanks again!

README.md
cpu: Sounds good. I don't feel strongly about this so I'm happy to leave it as-is in this PR.

db/memorystore.go
cpu: I think the calling code can be adjusted so that it would handle the `nil` without extra casing. I'll leave comments there.

wfe/wfe.go
```diff
@@ -971,6 +998,76 @@ func (wfe *WebFrontEndImpl) UpdateAccount(
...
+	if len(accountPageSplit) == 2 {
+		no, err := strconv.Atoi(accountPageSplit[1])
+		if err != nil || no < 2 {
+			wfe.sendError(acme.NotFoundProblem("Invalid page number"), response)
+			return
+		}
+		page = no - 1
+	}
+	orders := wfe.db.GetOrdersByAccountID(existingAcct.ID)
+	filteredOrders := make([]*core.Order, 0, len(orders))
```
cpu: I think if you switch this to:
```
var filteredOrders []*core.Order
for _, order := range orders {
  ...
  filteredOrders = append(filteredOrders, order)
}
```
then a `nil` result from `wfe.db.GetOrdersByAccountID` would be handled fine. You would end up with an empty `filteredOrders` slice and could check if `len == 0` and exit early.

db/memorystore.go
felixfontein: Changed it to return `nil`.

wfe/wfe.go
felixfontein: I think there's no code change needed to accomodate for `nil`. Both `len()` and `range` seem to work fine when `orders` is `nil`.

felixfontein: Just tested it, works fine.

cpu: Good call! Thanks @felixfontein

felixfontein:
Thanks a lot for reviewing this @cpu. I know that reviews can take some time, that's not different here than at other places ;)

PR #267, opened 2019-08-05
wfe: change UpdateAccount to allow blank (ie, removing) emails

eggsampler:
POST'ing to this endpoint with missing or empty contacts resulted in an error: "Use POST-as-GET to retrieve account data instead of doing an empty update" - however, no contacts is a perfectly valid update to an account.

This change rearranges and simplifies some of the logic in the wfe UpdateAccount function to allow a no contact update, and is a little more explicit about the POST-AS-GET to the endpoint.

### DIFF
--- a/wfe/wfe.go
+++ b/wfe/wfe.go
@@ -887,44 +887,30 @@ func (wfe *WebFrontEndImpl) UpdateAccount(
 		return
 	}
 
+	if postData.postAsGet {
+		wfe.sendError(
+			acme.MalformedProblem("Use the newAccount endpoint with onlyReturnExisting to retrieve an existing account."), response)
+		return
+	}
+
+	// Attempt to fetch existing account
+	// If the account is already deactivated, this will return a problem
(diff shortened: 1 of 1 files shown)


felixfontein:
Last time I checked, removing contacts worked fine by sending `{"contacts":[]}`. That does not result in any error message. Sending `{contacts:null}` or `{}` should not affect contacts at all (the former because `null` is treated the same as "`contacts` not specified").

Also, IIRC, disallowing an update of `{}` (which I think is OK by RFC8555, it does not mandate that all account updates have to be accepted) is because old ACME clients used an update of `{}` to query the current account information -- and for that, they should now use POST-as-GET.

eggsampler:
In review I need to make a few more changes to the logic and better state the problems with replication examples. I'll close this for now and submit another later.

felixfontein:
How about creating an issue where you describe what you are sending and what you expect to happen (and what actually happens)?

PR #269, opened 2019-08-18
Add SubjectKeyId to certificates

felixfontein:
According to https://tools.ietf.org/html/rfc5280.html#section-4.2.1.2 all certificates including the Basic Constraints extension should have a `SubjectKeyId`. This PR makes sure the certificates generated by Pebble have one: it adds random `SubjectKeyId`s to all certificates. According to https://golang.org/pkg/crypto/x509/#CreateCertificate, the intermediate and leaf certs should get `AuthorityKeyId` automatically added.

### DIFF
--- a/ca/ca.go
+++ b/ca/ca.go
@@ -4,8 +4,10 @@ import (
 	"crypto"
 	"crypto/rand"
 	"crypto/rsa"
+	"crypto/sha1"
 	"crypto/x509"
 	"crypto/x509/pkix"
+	"encoding/asn1"
 	"encoding/hex"
 	"fmt"
 	"log"
@@ -50,21 +52,49 @@ func makeSerial() *big.Int {
 	return serial
 }
 
-// makeKey and makeRootCert are adapted from MiniCA:
-// https://github.com/jsha/minica/blob/3a621c05b61fa1c24bcb42fbde4b261db504a74f/main.go
-
-// makeKey creates a new 2048 bit RSA private key
-func makeKey() (*rsa.PrivateKey, error) {
-	key, err := rsa.GenerateKey(rand.Reader, 2048)
+// Taken from https://github.com/cloudflare/cfssl/blob/b94e044bb51ec8f5a7232c71b1ed05dbe4da96ce/signer/signer.go#L221-L244
+func makeSubjectKeyID(key crypto.PublicKey) ([]byte, error) {
+	// Marshal the public key as ASN.1
+	pubAsDER, err := x509.MarshalPKIXPublicKey(key)
 	if err != nil {
 		return nil, err
 	}
-	return key, nil
+
+	// Unmarshal it again so we can extract the key bitstring bytes
+	var pubInfo struct {
+		Algorithm        pkix.AlgorithmIdentifier
+		SubjectPublicKey asn1.BitString
+	}
+	_, err = asn1.Unmarshal(pubAsDER, &pubInfo)
+	if err != nil {
+		return nil, err
+	}
+
+	// Hash it according to https://tools.ietf.org/html/rfc5280#section-4.2.1.2 Method #1:
+	ski := sha1.Sum(pubInfo.SubjectPublicKey.Bytes)
+	return ski[:], nil
+}
+
+// makeKey and makeRootCert are adapted from MiniCA:
+// https://github.com/jsha/minica/blob/3a621c05b61fa1c24bcb42fbde4b261db504a74f/main.go
+
+// makeKey creates a new 2048 bit RSA private key and a Subject Key Identifier
+func makeKey() (*rsa.PrivateKey, []byte, error) {
+	key, err := rsa.GenerateKey(rand.Reader, 2048)
+	if err != nil {
+		return nil, nil, err
+	}
+	ski, err := makeSubjectKeyID(key.Public())
+	if err != nil {
+		return nil, nil, err
+	}
+	return key, ski, nil
 }
 
 func (ca *CAImpl) makeRootCert(
 	subjectKey crypto.Signer,
 	subject pkix.Name,
+	subjectKeyID []byte,
 	signer *issuer) (*core.Certificate, error) {
 
 	serial := makeSerial()
@@ -76,6 +106,7 @@ func (ca *CAImpl) makeRootCert(
 
 		KeyUsage:              x509.KeyUsageDigitalSignature | x509.KeyUsageCertSign,
 		ExtKeyUsage:           []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth, x509.ExtKeyUsageClientAuth},
+		SubjectKeyId:          subjectKeyID,
 		BasicConstraintsValid: true,
 		IsCA:                  true,
 	}
@@ -119,7 +150,7 @@ func (ca *CAImpl) makeRootCert(
 
 func (ca *CAImpl) newRootIssuer() (*issuer, error) {
 	// Make a root private key
-	rk, err := makeKey()
+	rk, subjectKeyID, err := makeKey()
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


ca/ca.go
```diff
@@ -50,6 +50,14 @@ func makeSerial() *big.Int {
 	return serial
 }
 
+// Create random Subject Key Identifier
+func makeSubjectKeyId() []byte {
+	subjectKeyId := make([]byte, 20)
+    rand.Read(subjectKeyId)
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
	rand.Read(subjectKeyId)
```

ca/ca.go
```diff
@@ -50,6 +50,14 @@ func makeSerial() *big.Int {
 	return serial
 }
 
+// Create random Subject Key Identifier
+func makeSubjectKeyId() []byte {
```
golangcibot: func makeSubjectKeyId should be makeSubjectKeyID (from `golint`)

ca/ca.go
```diff
@@ -50,6 +50,14 @@ func makeSerial() *big.Int {
...
+	subjectKeyId := make([]byte, 20)
```
golangcibot: var `subjectKeyId` should be `subjectKeyID` (from `golint`)

ca/ca.go
```diff
@@ -68,6 +76,7 @@ func (ca *CAImpl) makeRootCert(
 	signer *issuer) (*core.Certificate, error) {
 
 	serial := makeSerial()
+	subjectKeyId := makeSubjectKeyId()
```
golangcibot: var `subjectKeyId` should be `subjectKeyID` (from `golint`)

ca/ca.go
```diff
@@ -186,6 +196,7 @@ func (ca *CAImpl) newCertificate(domains []string, ips []net.IP, key crypto.Publ
 	}
 
 	serial := makeSerial()
+	subjectKeyId := makeSubjectKeyId()
```
golangcibot: var `subjectKeyId` should be `subjectKeyID` (from `golint`)

ca/ca.go
```diff
@@ -50,6 +50,13 @@ func makeSerial() *big.Int {
 	return serial
 }
 
+// Create random Subject Key Identifier
+func makeSubjectKeyId() []byte {
+	subjectKeyId := make([]byte, 20)
+	rand.Read(subjectKeyId)
```
golangcibot: Error return value of `rand.Read` is not checked (from `errcheck`)

ca/ca.go
```diff
@@ -50,6 +50,13 @@ func makeSerial() *big.Int {
 	return serial
 }
 
+// Create random Subject Key Identifier
+func makeSubjectKeyID() []byte {
+	subjectKeyID := make([]byte, 20)
+	rand.Read(subjectKeyID)
```
golangcibot: Error return value of `rand.Read` is not checked (from `errcheck`)

felixfontein:
There's a potential problem: this causes the leaf certificate to get a `AuthorizedKeyIdentifier` extension, which points to the main intermediate. But this certificate is also delivered as an alternative leaf certificate with an alternative chain, which would then be wrong. If I understand this correctly, there need to be one leaf certificate per alternate chain.

alexzorin:
For alternate chains in ACME, it's necessarily the case that `issuer.ski == leaf.aki`, because it's a requirement of RFC5280. Anything else is an invalid configuration.

felixfontein:
@alexzorin that's correct, though it might encourage ACME clients to assume that this is always the case with alternate chains returned by ACME servers. I've extended Pebble so that it creates a different leaf certificate per alternate chain, so that each alternate chain is a proper chain according to RFC5280.

alexzorin:
I don't think issuing two leaf certificates is what is intended by ACME.

>The server MAY provide one or more link relation header fields
   [RFC8288] with relation "alternate".  Each such field SHOULD express
   an alternative certificate chain **starting with the same end-entity
   certificate**

The SKI is supposed to uniquely and consistently identify the public key of the certificate (https://tools.ietf.org/html/rfc5280#section-4.2.1.2):

>For CA certificates, subject key identifiers SHOULD be derived from
   the public key or a method that generates unique values
 
Pebble already correctly re-uses the same private key for every intermediate that it uses to actually sign certificates:

https://github.com/letsencrypt/pebble/blob/bb4787328a963c67e223f3f37741908aff2a6e54/ca/ca.go#L249-L256

What follows is that every intermediate's SKI is identical (according to RFC5280, but not according to this PR so far).

This leads to a very simple solution: set the SKI of the intermediate according to its public key (you can use [this](https://play.golang.org/p/58cpHs-N7lV)), and everything else falls into place (no SKI/AKI mismatches between the leaf and any of the alternate chains).

felixfontein:
> I don't think issuing two leaf certificates is what is intended by ACME.
> 
> > The server MAY provide one or more link relation header fields
> > [RFC8288] with relation "alternate".  Each such field SHOULD express
> > an alternative certificate chain **starting with the same end-entity
> > certificate**

Ah. I thought I'd read that somewhere, but didn't see it when looking again yesterday.

> The SKI is supposed to uniquely and consistently identify the public key of the certificate (https://tools.ietf.org/html/rfc5280#section-4.2.1.2):
> 
> > For CA certificates, subject key identifiers SHOULD be derived from
> > the public key or a method that generates unique values

So a random value is fine as well :) But all intermediates should better have the same one.

> This leads to a very simple solution: set the SKI of the intermediate according to its public key (you can use [this](https://play.golang.org/p/sJ_zLczoP20) to calculate it), and everything else falls into place (no SKI/AKI mismatches between the leaf and any of the alternate chains).

Where does that code come from? (Is it yours?) What license does it have? Can we use it in Pebble?

felixfontein:
I've implemented this now. I've exchanged the random SKI with the code provided by @alexzorin in the last commit; in case it turns out we can't use this code, we can simply remove that commit (or replace it by something else).

alexzorin:
The SKI code [is taken from cfssl](https://github.com/cloudflare/cfssl/blob/b94e044bb51ec8f5a7232c71b1ed05dbe4da96ce/signer/signer.go#L221-L244), which is what Boulder uses to implement its CA component. @cpu has previously noted that cfssl is too big a dependency to include in Pebble, but hopefully what is effectively a 3-liner can be copied under fair use.

felixfontein:
I guess that's up to @jsha and @cpu to decide. I've added error handling to the snippet (similar to the original), now it's a bit more than three lines :) I've also added the source in a comment.

cpu:
@felixfontein Can you help me explain what motivated this change? I agree the certificates being issued by Pebble now don't match the SubjectKeyId requirement from RFC 5280 but without understanding what complications that is causing its hard to evaluate the need for the fix. 

It seems like a sensible change but if there is a larger perspective in mind to have Pebble issue certificates that fully adhere to 5280 and the baseline requirements vs a one-off fix there are probably going to be other spec considerations that aren't being adhered to.

cpu:
> hopefully what is effectively a 3-liner can be copied under fair use.

Agreed, I don't think there is a sensible alternative way to do what this code is doing with Go without effectively copying the function.

felixfontein:
> @felixfontein Can you help me understand what motivated this change? I agree the certificates being issued by Pebble now don't match the SubjectKeyId requirement from RFC 5280 but without understanding what complications that is causing its hard to evaluate the need for the fix.
> 
> It seems like a sensible change but if there is a larger perspective in mind to have Pebble issue certificates that fully adhere to 5280 and the baseline requirements vs a one-off fix there are probably going to be other spec considerations that aren't being adhered to.

For me the main motivation is that I'm implementing alternate certificate chain selection in my client, and selecting a chain by `authorityKeyID` is one of the options I want to offer. As far as I understand it, using `authorityKeyID` (which needs `subjectKeyID` to work) is the best way to do this.

Since Pebble isn't including `subjectKeyID` (and thus also no `authorityKeyID` down the chain), I currently have no way to test this with integration tests.

felixfontein:
(Rebasing so it's easier for me to run all tests at once.)

felixfontein:
I've rebased to remove the conflict (with #276 merged). During that process I squashed all commits, to avoid having to do the same conflict resolution several times.

felixfontein:
@alexzorin @cpu @jsha thanks a lot for your feedback, reviewing and merging! :)

Issue #268, opened 2019-08-13
CI data race


cpu:
```
==================
WARNING: DATA RACE
Write at 0x00c0001b80e8 by goroutine 156:
  github.com/letsencrypt/pebble/va.VAImpl.setAuthzInvalid()
      /home/travis/gopath/src/github.com/letsencrypt/pebble/va/va.go:222 +0x88
  github.com/letsencrypt/pebble/va.VAImpl.process()
      /home/travis/gopath/src/github.com/letsencrypt/pebble/va/va.go:256 +0x478
Previous read at 0x00c0001b80e8 by goroutine 153:
  github.com/letsencrypt/pebble/db.(*MemoryStore).FindValidAuthorization()
      /home/travis/gopath/src/github.com/letsencrypt/pebble/db/memorystore.go:225 +0x169
  github.com/letsencrypt/pebble/wfe.(*WebFrontEndImpl).makeAuthorizations()
      /home/travis/gopath/src/github.com/letsencrypt/pebble/wfe/wfe.go:1296 +0x2d5
  github.com/letsencrypt/pebble/wfe.(*WebFrontEndImpl).NewOrder()
      /home/travis/gopath/src/github.com/letsencrypt/pebble/wfe/wfe.go:1474 +0xc61
  github.com/letsencrypt/pebble/wfe.(*WebFrontEndImpl).NewOrder-fm()
      /home/travis/gopath/src/github.com/letsencrypt/pebble/wfe/wfe.go:1402 +0x73
  github.com/letsencrypt/pebble/wfe.(*WebFrontEndImpl).HandleFunc.func1()
      /home/travis/gopath/src/github.com/letsencrypt/pebble/wfe/wfe.go:240 +0x976
  github.com/letsencrypt/pebble/wfe.wfeHandlerFunc.ServeHTTP()
      /home/travis/gopath/src/github.com/letsencrypt/pebble/wfe/wfe.go:113 +0x7a
  github.com/letsencrypt/pebble/wfe.(*topHandler).ServeHTTP()
      /home/travis/gopath/src/github.com/letsencrypt/pebble/wfe/wfe.go:125 +0x6e
  net/http.StripPrefix.func1()
      /home/travis/.gimme/versions/go1.12.8.linux.amd64/src/net/http/server.go:2034 +0x362
  net/http.HandlerFunc.ServeHTTP()
      /home/travis/.gimme/versions/go1.12.8.linux.amd64/src/net/http/server.go:1995 +0x51
[...]

PR #270, opened 2019-08-18
db: missing authz rlock in FindValidAuthorization

alexzorin:
Fixes the race reported in #268

### DIFF
--- a/db/memorystore.go
+++ b/db/memorystore.go
@@ -222,11 +222,14 @@ func (m *MemoryStore) FindValidAuthorization(accountID string, identifier acme.I
 	m.RLock()
 	defer m.RUnlock()
 	for _, authz := range m.authorizationsByID {
+		authz.RLock()
 		if authz.Status == acme.StatusValid && identifier.Equals(authz.Identifier) &&
 			authz.Order != nil && authz.Order.AccountID == accountID &&
 			authz.ExpiresDate.After(time.Now()) {
+			authz.RUnlock()
 			return authz
 		}
+		authz.RUnlock()
 	}
 	return nil
 }


alexzorin:
On reflection, I don't think adding more granular locks is the right fix. After all, this function already held a read lock on the MemoryStore.

Seems like the actual problem is that MemoryStore is handing out mutable pointers to its underlying data and then some code (like `va.setAuthzInvalid`) is abusing that fact.

Uh oh!

cpu:
> Seems like the actual problem is that MemoryStore is handing out mutable pointers to its underlying data and then some code (like va.setAuthzInvalid) is abusing that fact.

:sweat: I think the slap dash parts of Pebble we threw together at the start to bootstrap the Boulder ACME v2 work are starting to show their age/limited design thought.

> On reflection, I don't think adding more granular locks is the right fix. After all, this function already held a read lock on the MemoryStore.

Agreed. I think more subtantial rework is probably required. I'm hoping to have a chance to pivot to more substantial Pebble work in ~Sept. On the top of my list was more comprehensive tests. I think addressing the memorystore hack is another priority. 

I think I'd be willing to swallow the cost of a new dependency to gain a more robust in-memory database implementation (it would also knock out the many O(N) query operations), maybe something like https://github.com/hashicorp/go-memdb My primary criteria is avoiding persistence/durability.

WDYT?

alexzorin:
I think MemoryStore could be reworked by banning pointers and requiring everything to be passed around by value. That should allow us to eliminate all of the locks except a single RW lock on the store, right?

go-memdb [would appear to be](https://play.golang.org/p/VZLk3QLFrdq) a sidegrade from that perspective - the same pitfall with leaky pointers exists there.

I don't have any strong feelings about fixing the O(N) scans. Seems like it's not something that would matter for a few hundred slice elements, but tbh even the addition of `ordersByAccountID` in #260 seemed like an unnecessary optimization to me.

cpu:
>  go-memdb would appear to be a sidegrade from that perspective - the same pitfall with leaky pointers exists there.

Agreed, I think either with the existing memstore or an alternative library we would have to switch to passing things around by value.

>  I don't have any strong feelings about fixing the O(N) scans

Me either. Certainly not a big priority given Pebble's explicit use cases.

> Edit: Now that I think about it, proper transactions are probably going to be required if doing pass-by-value.

This was my thought too. We could switch the existing memstore around to pass by value but without transactions I think there will be some weird inconsistencies made possible (maybe that doesn't matter that much?) with `go-memdb` we could use transactions and avoid that.

I'm interested in hearing what @jsha thinks. Historically he's been able to convince me I'm overthinking things with good regularity :laughing: He's OOO this week but will be back next week. It's unlikely I'll start moving on any of this before then.

jsha:
We talked about this by voice; sorry I forgot to update here. I think pass-by-value is fine, since all we expect to return for a given HTTP request is a snapshot of an object's state at a point in time. If that state happened to change while the HTTP request was inflight, no big deal; the next polling request or the next attempt to update the object will get the updated value.

Note that rewriting everything to be pass-by-value would be pretty annoying, because it means a lot of places that `return nil, err` would have to be rewritten as `return SomeType{}, err`. Instead I think it makes sense to give each relevant type a `.clone()` method and return that, or just use [`reflect.Copy()`](https://golang.org/pkg/reflect/#Copy).

PR #276, opened 2019-08-30
Make sure all intermediate certificates have the same subject

felixfontein:
While working on #269 I noticed that all intermediate certificates (assuming `PEBBLE_ALTERNATE_ROOTS > 0`) have different subjects, whence only the first one's subject equals the issuer of the generated leaf certificates. This makes validation of all alternate chains fail.

This PR ensures that all intermediate certificates share the same subject. (They all have different serial numbers, whence the hex string contained in the intermediate's subject is **no longer related** to the serial number(s).)

(Cherry-pick of a476ec3edc03098b304505f2b6d212db87ccd0ce from #269.)

### DIFF
--- a/ca/ca.go
+++ b/ca/ca.go
@@ -64,14 +64,12 @@ func makeKey() (*rsa.PrivateKey, error) {
 
 func (ca *CAImpl) makeRootCert(
 	subjectKey crypto.Signer,
-	subjCNPrefix string,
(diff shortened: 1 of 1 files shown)


jsha:
Nice fix! :rocket:

felixfontein:
@jsha thanks for reviewing and merging!

PR #279, opened 2019-09-16
Docker: expose ports 14000 and 15000 in the pebble image

sergioaugrod:
Hello,

I exposed ports 14000 and 15000 in the pebble image. I believe this improves readability of Dockerfile and this will help me to run the image on GitlabCI. 

Does that make sense to you?

Thank you!

### DIFF
--- a/docker/pebble/linux.Dockerfile
+++ b/docker/pebble/linux.Dockerfile
@@ -20,3 +20,6 @@ COPY --from=builder /go/bin/pebble /usr/bin/pebble
 COPY --from=builder /pebble-src/test/ /test/
 
 CMD [ "/usr/bin/pebble" ]
+
+EXPOSE 14000
+EXPOSE 15000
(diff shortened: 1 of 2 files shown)


cpu:
Hi @sergioaugrod :wave: Thanks for opening a PR!

> this will help me to run the image on GitlabCI.

Can you explain a little bit more about this part? My understanding was that the `EXPOSE` directive in a Dockerfile was strictly metadata and publishing the port requires additional configuration either way but I might be missing something about your use-case.

sergioaugrod:
Hi @cpu. You are right. 

My problem is that Gitlab CI HealthCheck checks which ports are exposed from the container by default and If I don't expose any port in my Dockerfile, then the pipeline will fail. Currently, I can't expose ports in my `gitlab-ci.yml`.

There is a discussion about this here: https://gitlab.com/gitlab-org/gitlab-runner/issues/2460

Despite this very specific problem of mine, would it be nice to have it in Dockerfile to improve the documentation?

Thank you!

Issue #291, opened 2019-11-29
POST-as-GET for directory and newNonce endpoints


eggsampler:
While doing some testing for the upcoming POST-as-GET changes, I noticed that [RFC8555 Section 6.3](https://tools.ietf.org/html/rfc8555#section-6.3) states that,

> The server MUST allow GET requests for the directory and newNonce resources (see Section 7.1), in addition to POST-as-GET requests for these resources.

From what I can tell, this isn't supported in either pebble or boulder: [repro](https://github.com/eggsampler/acme/blob/fetch/acme_test.go#L54) / [test](https://travis-ci.com/eggsampler/acme/jobs/261478210#L242).

Is this something you think should be supported? I doubt anyone has run into this, but it seems like it should be supported per the rfc.

It's a relatively easy fix in pebble, [this](https://github.com/eggsampler/pebble/commit/ce1b72228059e8250105a7708556bdf68a86f654) seems to do the trick (but may need some adjusting of the linter).

cpu:
Hey @eggsampler, thanks for opening an issue.

> Is this something you think should be supported? I doubt anyone has run into this, but it seems like it should be supported per the rfc.

Yup, I think supporting it makes sense. I don't have any philosophical opposition. I think we probably just overlooked that sentence in the RFC when implementing POST-as-GET support.

Thanks for catching that.

cpu:
> It's a relatively easy fix in pebble, this seems to do the trick (but may need some adjusting of the linter).

LGTM, I went ahead and stole your fix to make a PR: https://github.com/letsencrypt/pebble/pull/292 Thanks for making that easy :-)

I'll look to get that merged next week when my American counterparts are back from holiday. I think this fix merits a new release tag so I'll make a note to do that next week too.

Thanks again!

cpu:
> this isn't supported in either pebble or boulder

Filed https://github.com/letsencrypt/boulder/issues/4584 for fixing the Boulder side. Probably won't get a chance to take a whack at that until later next week.

PR #292, opened 2019-11-29
wfe: allow POST-as-GET to dir. and newNonce endpoints

cpu:
[RFC 8555 §6.3](https://tools.ietf.org/html/rfc8555#section-6.3) says the server's `directory` and `newNonce` endpoints should support POST-as-GET as well as GET.

This PR adopts a fix from eggsampler's [Pebble fork](eggsampler@ce1b722). Thanks @eggsampler !

Along the way to avoid the ire of the linter I replaced the numerous occurrences of "GET", "POST", and "HEAD" with the correct exported `http` package constants.

Resolves https://github.com/letsencrypt/pebble/issues/291

### DIFF
--- a/wfe/wfe.go
+++ b/wfe/wfe.go
@@ -431,10 +431,10 @@ func (wfe *WebFrontEndImpl) handleCertStatusBySerial(
 
 func (wfe *WebFrontEndImpl) Handler() http.Handler {
 	m := http.NewServeMux()
-	// GET only handlers
-	wfe.HandleFunc(m, DirectoryPath, wfe.Directory, "GET")
+	// GET & POST handlers
+	wfe.HandleFunc(m, DirectoryPath, wfe.Directory, "GET", "POST")
 	// Note for noncePath: "GET" also implies "HEAD"
-	wfe.HandleFunc(m, noncePath, wfe.Nonce, "GET")
+	wfe.HandleFunc(m, noncePath, wfe.Nonce, "GET", "POST")
 
 	// POST only handlers
 	wfe.HandleFunc(m, newAccountPath, wfe.NewAccount, "POST")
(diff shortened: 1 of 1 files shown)


wfe/wfe.go
```diff
@@ -431,23 +431,23 @@ func (wfe *WebFrontEndImpl) handleCertStatusBySerial(
 
 func (wfe *WebFrontEndImpl) Handler() http.Handler {
 	m := http.NewServeMux()
-	// GET only handlers
-	wfe.HandleFunc(m, DirectoryPath, wfe.Directory, "GET")
-	// Note for noncePath: "GET" also implies "HEAD"
-	wfe.HandleFunc(m, noncePath, wfe.Nonce, "GET")
+	// GET & POST handlers
+	wfe.HandleFunc(m, DirectoryPath, wfe.Directory, http.MethodGet, http.MethodPost)
```
jsha: It looks like all the HandleFuncs now allow MethodPost. Maybe it's better to change the HandleFunc definition so it automatically adds POST? And then adds whatever methods are left in the varargs?

wfe/wfe.go
```diff
@@ -478,6 +478,20 @@ func (wfe *WebFrontEndImpl) Directory(
 		"keyChange":  keyRolloverPath,
 	}
 
+	// RFC 8555 §6.3 says the server's directory endpoint should support
```
jsha: I like the proper section markers. Hooray for broad Unicode support! :-D

wfe/wfe.go
cpu: I feel like that will make reading the code more confusing and we're unlikely to be adding new handlers with much frequency. Do you feel strongly about it? I have a light preference for the explicit version.

cpu:
> there's no reason to switch over to POST-as-GET for subsequent runs.

Agreed, but the spec says what it says. All of the POST-as-GET stuff is jank and this is just another weird artifact of changing the spec with minimal input at the very end of the process. ¯\\\_(ツ)\_/¯

wfe/wfe.go
jsha: I don't feel strongly about it; the explicit version is fine.

PR #294, opened 2019-12-04
va: fix HTTP01 port number issue when redirected

### LINKED ISSUE #293: Regression when validating HTTP01 with a HTTP->HTTPS redirect
As part of implementing #288 and testing it in our own client, we have 'fast-forwarded' the version of Pebble used in our tests from `2e69bb16af048c491720f23cb284fce685e65fec` to `282a77074e9ef932baf2891d4e84a018a6a31a3d` (HEAD of that PR).

As per https://github.com/jetstack/cert-manager/pull/2392#issuecomment-561562971, we've started noticing this error specifically in our test cases that are configured to redirect traffic to the HTTP01 acme solver to HTTPS (i.e. http://domain.to.solve/... 301 https://domain.to.solve/...) - this is something supported according to the ACME spec - the PR to make this possible was https://github.com/letsencrypt/pebble/pull/196. The actual error in question:

```
  Warning  Failed     4m31s  cert-manager  Accepting challenge authorization failed: acme: authorization error for efsxk.certmanager.kubernetes.network: 400 urn:ietf:params:acme:error:connection: Get https://efsxk.certmanager.kubernetes.network/.well-known/acme-challenge/TRrVcMxk3V2oH6r6WK3kbMzmNvWopGCiRRtsqJHNe3k: http: server gave HTTP response to HTTPS client
```

This is a strange error, as the error message implies that the GET request was against `https://...` - whilst this may suggest that our web server is incorrectly serving HTTP traffic on port 443, I suspect that this is *not* the case given we've only observed it after changing *Pebble's* version.

Looking at recent changes to the `va` package, I can see two PRs that look like they may be relevant:

https://github.com/letsencrypt/pebble/commit/bc4da68d49d78a38bfea869d38faadc8c714654c#diff-4c51d1d7ca3ec3022d14b42809af0d7eR447 - this changes the `fetchHTTP` function, although looking at it closely, does not seem like it'd cause an issue (the `Host` field should be set to the same value regardless).

https://github.com/letsencrypt/pebble/commit/ce13973920007bf4ed40c2ddfd7527d169b541d8#diff-4c51d1d7ca3ec3022d14b42809af0d7eR526 - this one looks more suspect. Notice a custom dialer was added, which hardcodes the port to fetch to `portString`, which is fixed to `80` (the HTTP port) in our test cases.

I believe that this particular segment is wrong, as it doesn't allow to account for redirects returned from the serving being tested to be properly processed.

I'm opening the issue first to ensure that I'm not going mad here (😅) - but also to ask how a fix to this can be approached? I understand the reasoning for the dialer (so we can customise the DNS resolver used when solving HTTP01 requests), but I'm uncertain how we can properly plumb this through whilst still accounting for redirects! (i.e. what if my HTTP01 server redirected to a different *Host* altogether?)

cc @cpu @adferrand

munnerz:
This attempts to fix the issue described in #293, whereby the port number to connect to is hardcoded into the custom dialer.

I've opted to try and actually override the `Resolver` field within the Dialer, so we can specifically hijack requests to DNS servers whilst still maintaining the existing dialer logic.

I've not had a chance to test this just yet, as I'll need to stand up some basic testing environment to ensure requests are properly routed!

fixes #293

### DIFF
--- a/va/va.go
+++ b/va/va.go
@@ -513,18 +513,6 @@ func (va VAImpl) fetchHTTP(identifier string, token string) ([]byte, string, *ac
 	httpRequest.Header.Set("User-Agent", userAgent())
 	httpRequest.Header.Set("Accept", "*/*")
 
-	addrs, err := va.resolveIP(identifier)
-
-	if err != nil {
-		return nil, url.String(), acme.MalformedProblem(
-			fmt.Sprintf("Error occurred while resolving URL %q: %q", url.String(), err))
-	}
-
-	if len(addrs) == 0 {
-		return nil, url.String(), acme.MalformedProblem(
-			fmt.Sprintf("Could not resolve URL %q", url.String()))
-	}
-
 	transport := &http.Transport{
 		// We don't expect to make multiple requests to a client, so close
 		// connection immediately.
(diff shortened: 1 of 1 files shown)


munnerz:
I've run into an error in our own client when using this patch (that slips through CI here somehow....) so don't merge for now!

```
  Reason:      Error accepting challenge: acme: urn:ietf:params:acme:error:connection: Get http://yume.certmanager.kubernetes.network:80/.well-known/acme-challenge/tJlOK2KuD_4u4LsuJUgb1CvSTvSykZqzIPqU-9gBDfE: dial tcp: lookup yume.certmanager.kubernetes.network on 10.0.0.10:53: dial udp: missing address
```

munnerz:
I’ve updated the PR to no longer override the Resolver and instead extended the previous mechanism to resolve within DialContext if necessary, with an extremely minimal caching mechanism. It now passes in the corresponding PRs to cert-manager 🥳

cpu:
@munnerz Do you feel strongly about removing the resolution cache per @jsha's suggestion? I've been waiting on reviewing this PR for that change to land.

Thanks!

munnerz:
@cpu @jsha I've updated the PR to remove the resolution cache! Thanks 😄

jsha:
Thanks for the PR! :hammer: :wrench:

PR #299, opened 2020-02-08
ci: update golangci-lint and check go modules

ldez:
- update golangci-lint to v1.23.3
- validate go module files and vendor folder.

Maybe it's time to remove the vendor folder?


cpu:
> Maybe it's time to remove the vendor folder?

I'm open to the idea but don't have a strong opinion. With the new Go module proxy a vendor dir is less valuable than before. @jsha what do you think? It seems unlikely the Boulder vendor dir will get dropped anytime soon so perhaps it's best to just leave the two in the same state?

jsha:
I also don't have a strong opinion on whether to keep the vendor folder for Pebble, but I do want to keep it for Boulder. I agree with the idea that it makes sense to just keep the two the same (at least for now).

PR #301, opened 2020-02-24
Orders don't have a "deactivated" status.

### LINKED ISSUE #300: wrong implementation of ACME spec: Orders should transition to "invalid",  not "deactivated"
I spent the day investigating failing tests, because of a single line in Pebble :(

https://github.com/letsencrypt/pebble/blob/master/core/types.go#L66-L69

	// An order is deactivated if **any** of its authzs are deactivated
	if authzStatuses[acme.StatusDeactivated] > 0 {
		return acme.StatusDeactivated, nil
	}

This is not a correct implementation of the ACME spec.  The Order object does not have a "deactivated" state.  The spec states it should transition to "invalid".

[7.1.3 Order Objects](https://tools.ietf.org/html/rfc8555#section-7.1.3)

>    status (required, string):  The status of this order.  Possible
>      values are "pending", "ready", "processing", "valid", and
>      "invalid".  See Section 7.1.6.

[7.1.6.  Status Changes](https://tools.ietf.org/html/rfc8555#section-7.1.6)
[Page 32: text following "State Transitions for Authorization Objects"]( https://tools.ietf.org/html/rfc8555#page-32)
> The order also moves to the "invalid"
> state if it expires or one of its authorizations enters a final state
> other than "valid" ("expired", "revoked", or "deactivated")

[Page 33: chart "State Transitions for Order Objects"]( https://tools.ietf.org/html/rfc8555#page-33)

```
    pending --------------+
       |                  |
       | All authz        |
       | "valid"          |
       V                  |
     ready ---------------+
       |                  |
       | Receive          |
       | finalize         |
       | request          |
       V                  |
   processing ------------+
       |                  |
       | Certificate      | Error or
       | issued           | Authorization failure
       V                  V
     valid             invalid
```

jsha:
Use "invalid" instead, and add comments linking to the spec.

Fixes #300


cpu:
:wave: FWIW Boulder has the same divergence: https://github.com/letsencrypt/boulder/blob/9a86abc1550425a73f912c25a11c29160f58a85a/sa/sa.go#L1218-L1221

jvanasco:
I think I managed to still trigger this. I need to do a few more tests to make sure i'm running the right code - I'm not very good with `go`

jsha:
Ok, thanks so much for testing! If you need any help, let me know.

Zola-github (dismissed):

jvanasco:
I finally got a chance to test. This works as expected from the RFC.

core/types.go
```diff
@@ -53,22 +53,16 @@ func (o *Order) GetStatus() (string, error) {
 		}
 	}
 
-	// An order is invalid if **any** of its authzs are invalid
-	if authzStatuses[acme.StatusInvalid] > 0 {
+	// An order is invalid if **any** of its authzs are invalid,
+	// deactivated, or expired
+	// https://tools.ietf.org/html/rfc8555#page-32
```
beautifulentropy, line 58:
```suggestion
	// An order is invalid if ANY of its authzs are invalid, deactivated, or
	// expired per: https://tools.ietf.org/html/rfc8555#page-32
```

core/types.go
```diff
@@ -53,22 +53,16 @@ func (o *Order) GetStatus() (string, error) {
...
-		return acme.StatusInvalid, nil
-	}
-
-	// An order is deactivated if **any** of its authzs are deactivated
-	if authzStatuses[acme.StatusDeactivated] > 0 {
-		return acme.StatusDeactivated, nil
-	}
-
-	// An order is pending if **any** of its authzs are pending
+	// Otherwise, an order is pending if **any** of its authzs are pending
```
beautifulentropy, line 65:
```suggestion
	// Otherwise, an order is pending if ANY of its authzs are pending.
```

beautifulentropy (changes requested):
Comment nits.

aarongable (approved):

beautifulentropy (approved):

cpu:
@aarongable @beautifulentropy @mcpherrinm I bumped into this in some client tests and was wondering if one of you could merge this :bow:

jvanasco:
>  I bumped into this in some client tests and was wondering if one of you could merge this

I now consider this a feature - not a bug - helping me expect/handle broken or malformed responses ;)

aarongable:
Lol, no idea why this never got merged. While pebble is meant to implement ACME differently than Boulder, it _isn't_ meant to implement ACME _incorrectly_. Thanks for the prod!

PR #306, opened 2020-04-03
Support for .onion address

orangepizza:
first pass - I'm sure this won't work but let's see where is breaks.

P.S 
1. what should be challenge type name for new CSR challenge? (currently onioncsr-01)
2. currently this accepts http and tls-alpn challenge as if it's normal domain, as I think other CA may want to run http challenge over tor network: but I don't know external program can talk ever tor daemon
3. added payload as challenge struct as string type, and decoded into der format by base64url. (as same with csr sent at finalization stap)
4. I should've write proper protocol extension first then write server side intuitively.


orangepizza:
I wonder why autometic checker can't find sha3 package? it's not no golang 1.13?

alexzorin:
>I wonder why autometic checker can't find sha3 package? it's not no golang 1.13?

I think it's because this project uses vendoring (look in `vendor/` and you'll see the crypto directory tree, without a `sha3` directory).

What I think will fix this is running:

    go mod download
    go mod vendor

and then committing `vendor/golang.org/x/crypto/sha3/` (new) and `vendor/modules.txt` (updated).

I gleaned this from the `.travis.yml` file, not sure if it's totally right. Maybe jsha can clarify once he gets a chance.

alexzorin:
Did you see [APPENDIX C - Issuance of Certificates for .onion Domain Names](https://github.com/cabforum/documents/blob/master/docs/BR.md#appendix-c--issuance-of-certificates-for-onion-domain-names) in the updated BRs?

For the CSR-based validation the signed CSR needs to include two additional extensions that carry nonces:

>b. The CA MAY verify the Applicant's control over the .onion service by having the Applicant provide a Certificate Request signed using the .onion public key if the Attributes section of the certificationRequestInfo contains:
>(i) A caSigningNonce attribute that contains a Random Value that is generated by the CA; and (ii) An applicantSigningNonce attribute that contains a single value with at least 64-bits of entropy that is generated by the Applicant.

I am guessing we could use the ACME challenge `token` as the `caSigningNonce`. Though this means that the CSR would not be bound to the ACME account in any way, because we wouldn't be using the key authorization. (?)

orangepizza:
> 
> 
> Did you see [APPENDIX C - Issuance of Certificates for .onion Domain Names](https://github.com/cabforum/documents/blob/master/docs/BR.md#appendix-c--issuance-of-certificates-for-onion-domain-names) in the updated BRs?
> 
> For the CSR-based validation the signed CSR needs to include two additional attributes that carry nonces:
> 
> > b. The CA MAY verify the Applicant's control over the .onion service by having the Applicant provide a Certificate Request signed using the .onion public key if the Attributes section of the certificationRequestInfo contains:
> > (i) A caSigningNonce attribute that contains a Random Value that is generated by the CA; and (ii) An applicantSigningNonce attribute that contains a single value with at least 64-bits of entropy that is generated by the Applicant.
> 
> I am guessing we could use the ACME challenge `token` as the `caSigningNonce`. Though this means that the CSR would not be bound to the ACME account in any way, because we wouldn't be using the key authorization. (?)

document says
 cabf-caSigningNonce OBJECT IDENTIFIER ::= { cabf 41 }
but x509 format expects it'd be []int so have no idea how to interprete this oid of extension should be

alexzorin:
I believe it'd be:

```golang
asn1.ObjectIdentifier{2, 23, 140, 41}
```

but at this early stage, I don't think it really matters.

alexzorin:
@orangepizza I'm getting the distinct feeling that Go's `x509` package can't see single attributes like the one we need here. 

I generated a CSR with `openssl` with the attribute - Go can't see it. Attributes, Extensions and ExtraExtensions are all empty.

#15995 looks relevant.

This PR might be stuck, unless it implements its own parsing of the CSR ...

orangepizza:
> 
> 
> @orangepizza I'm getting the distinct feeling that Go's `x509` package can't see (or generate) single request attributes like the one we need here.
> 
> I generated a CSR with `openssl` with the attribute - Go can't see it. Attributes, Extensions and ExtraExtensions are all empty.
> 
> [golang/go#15995](https://github.com/golang/go/issues/15995) looks relevant.
> 
> This PR might be stuck, unless it implements its own parsing of the CSR ...

it's be stupidly unsecure but as this is pebble can we search required nonce itself from raw csr []byte?

alexzorin:
>it's be stupidly unsecure but as this is pebble can we search required nonce itself from raw csr []byte?

We can do a little better. Here's something adapted from the workaround posted in the Go issue: https://play.golang.org/p/IvuS-ZkP8UX

Note that the CSR wasn't generated quite right - the attribute value is meant to be an OctetString and not a UTF8String - I couldn't figure out how to coax `openssl` into using a different encoding.

orangepizza:
and I don't understand why Applicant nonce is ever a thing here: is even possible to verify client used enough entropy for that?

alexzorin:
I think for now, we should check that that the `applicantSigningNonce` is present in the CSR, is a sufficient length and is of the `OctetString` type.

Whether or not the ACME extension will make use of it (I doubt it, since external account binding is already available), it must appear in the CSR no matter what.

PR #312, opened 2020-05-19
wfe: add CORS support.

### LINKED ISSUE #305: Cross-origin issue
I have an application where javascript reads the ACME directory (using XMLHttpRequest). 
With pebble, it fails due to cross-origin issues.  

It would be really helpful if pebble included `Access-Control-Allows-Origin: *` in its responses...

RFC 8555, 6.1 penultimate paragraph says "Such servers SHOULD set the Access-control-Allow-Origin header field to the value "*""...

Please?

alexzorin:
Resolves #305.

### DIFF
--- a/wfe/wfe.go
+++ b/wfe/wfe.go
@@ -246,7 +246,7 @@ func (wfe *WebFrontEndImpl) HandleFunc(
 		&topHandler{
 			wfe: wfeHandlerFunc(func(ctx context.Context, response http.ResponseWriter, request *http.Request) {
 				// Process CORS as necessary. If it's a CORS preflight, no further processing may occur.
-				if wfe.processCORS(request, response, methods) {
+				if wfe.processCORS(request, response, methodsMap) {
 					return
 				}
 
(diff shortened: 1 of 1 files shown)


wfe/wfe.go
```diff
@@ -280,6 +285,33 @@ func (wfe *WebFrontEndImpl) HandleFunc(
...
+		return false
+	}
+
+	// All CORS-aware responses include -Allow-Origin and -Expose-Headers
+	response.Header().Set("Access-Control-Allow-Origin", "*")
+	response.Header().Set("Access-Control-Expose-Headers", "Link, Replay-Nonce, Location")
+
+	// Preflight responses additionally include -Max-Age, -Allow-Methods, -Allow-Headers
+	// and terminate the response.
+	if request.Method == http.MethodOptions {
```
rolandshoemaker, line 304: It's only a valid preflight request if `Access-Control-Request-Method` is set in the request, and the value is a member of allowedMethods (in boulder we assume `GET` if the header is not set, which is probably fine to do here as well).

alexzorin: You're right. I've tried to follow the spec a bit more closely on this attempt. I think there's some value in not copying Boulder exactly, so thanks for your review and your patience.

PR #314, opened 2020-06-07
GolangCI closed

szepeviktor:
https://medium.com/golangci/golangci-com-is-closing-d1fc1bd30e0e


jsha:
Thanks for the PR! I'd like to keep `.golangci.yaml` because it's used by the commandline tool `golangci-lint`, which we run in Travis.

However, this reminds me we do need to remove the GolangCI badge from our README - would you like to do that?

szepeviktor:
> to remove the GolangCI badge from our README

Done.

Commit disappeared :)

Issue #303, opened 2020-02-28
Expired Authorizations/Challenges are still pending


jvanasco:
If an Authorization has expired, loading it will show a "pending" status.
If a Challenge belongs to an Authorization which has expired, loading it will show a "pending" status.

Triggering the expired Challenge will result in an error json payload, as expected, such as:

     {u'status': 400, u'type': u'urn:ietf:params:acme:error:malformed', u'detail': u'Authorization expired 2020-02-28T20:25:02Z'}

Attempting to load the Authorization or Challenge after receiving this payload will still return a "pending" status.

PR #315, opened 2020-06-12
authorization update

jvanasco:
This is a partial fix to #303

I am not sure if this is the correct place or way (does it persist?) to implement this. I would not be offended (and actually happy) if a maintainer were to better implement this PR.

This simply effects the correct state transition for the Authorization object (see https://tools.ietf.org/html/rfc8555#section-7.1.6) where the state is rendered to clients as `expired` after the expiry time.


jsha:
Thanks for filing @jvanasco! Things have been a bit busy here but this is on our radar to review.

Issue #297, opened 2020-01-19
Feature suggestion: allow setting a subset of configuration


codeclown:
For my project I had to set a custom value for `httpPort` in `pebble-config.json`.

Currently in my custom JSON I need to specify the rest of the configuration as well:

```json
{
  "pebble": {
    "listenAddress": "0.0.0.0:14000",
    "managementListenAddress": "0.0.0.0:15000",
    "certificate": "test/certs/localhost/cert.pem",
    "privateKey": "test/certs/localhost/key.pem",
    "httpPort": 80,
    "tlsPort": 5001,
    "ocspResponderURL": "",
    "externalAccountBindingRequired": false
  }
}
```

It would be convenient if I could omit all properties but the ones I need to modify. So my JSON would look like this:

```json
{
  "pebble": {
    "httpPort": 80
  }
}
```

Pebble could pull the default values for the rest.

cpu:
Seems like a reasonable request :+1: I probably won't have a chance to implement this short term but would accept a PR from the community.

PR #316, opened 2020-07-24
Add default configuration values

AdamBark:
Address issue #297

### DIFF
--- a/cmd/pebble/main.go
+++ b/cmd/pebble/main.go
@@ -29,6 +29,17 @@ type config struct {
 	}
 }
 
+func setConfigDefaults() (config) {
+	var conf config
+	conf.Pebble.ListenAddress = "0.0.0.0:14000"
+	conf.Pebble.ManagementListenAddress = "0.0.0.0:15000"
+	conf.Pebble.Certificate = "test/certs/localhost/cert.pem"
+	conf.Pebble.PrivateKey = "test/certs/localhost/key.pem"
+	conf.Pebble.HTTPPort = 5002
+	conf.Pebble.TLSPort = 5001
+	return conf
+}
+
 func main() {
 	configFile := flag.String(
 		"config",
@@ -52,7 +63,7 @@ func main() {
 	logger := log.New(os.Stdout, "Pebble ", log.LstdFlags)
 	logger.Printf("Starting Pebble ACME server")
 
-	var c config
+	c := setConfigDefaults()
 	err := cmd.ReadConfigFile(*configFile, &c)
 	cmd.FailOnError(err, "Reading JSON config file into config structure")
 


AdamBark:
I imagine this needs some more work but I thought it was a good place to start.

cmd/pebble/main.go
```diff
@@ -29,6 +29,17 @@ type config struct {
...
 
+func setConfigDefaults() (config) {
+	var conf config
+	conf.Pebble.ListenAddress = "0.0.0.0:14000"
+	conf.Pebble.ManagementListenAddress = "0.0.0.0:15000"
+	conf.Pebble.Certificate = "test/certs/localhost/cert.pem"
+	conf.Pebble.PrivateKey = "test/certs/localhost/key.pem"
+	conf.Pebble.HTTPPort = 5002
+	conf.Pebble.TLSPort = 5001
+	return conf
```
aarongable, line 40: There are two values (`ocspResponderUrl` and `externalAccountBindingRequired`) which don't get initialized here. Sure, their defaults are also their zero-values, but being explicit is good.

cmd/pebble/main.go
```diff
@@ -29,6 +29,17 @@ type config struct {
 	}
 }
 
+func setConfigDefaults() (config) {
+	var conf config
+	conf.Pebble.ListenAddress = "0.0.0.0:14000"
```
aarongable, line 34: Add a comment showing readers that these values are identical to the values in `/test/config/pebble-config.json`, and that they should be kept in sync.

AdamBark:
Yes, I was wondering whether it was better to read the defaults from a file. Should I fix up this PR or open a new one?

cmd/pebble/main.go
```diff
@@ -29,6 +29,17 @@ type config struct {
 	}
 }
 
+func setConfigDefaults() (config) {
```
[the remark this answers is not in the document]
AdamBark: That's what you get for not updating your function names when you change what it does :facepalm:

aarongable:
I'd go ahead and keep using this one -- we just ask that you update it by pushing additional commits, not by rewriting and then force-pushing. If you'd rather start over from scratch, opening a new one is fine too! Doesn't matter much on our end :)

cmd/pebble/main.go
```diff
@@ -29,10 +29,19 @@ type config struct {
 	}
 }
 
+const DefaultConfigPath = "test/config/default-config.json"
+
+func getDefaultConfig() (config) {
```
aarongable, line 34: The automation checks are failing because `go fmt` says you don't need parens around the return type.
```suggestion
func getDefaultConfig() config {
```

felixfontein:
> I was wondering whether it was better to read the defaults from a file. Should I fix up this PR or open a new one?

I'm not sure whether it is a good idea to expect the file `test/config/default-config.json` to be there. Until now, Pebble was working fine without that file existing if you pointed it to an alternate configuration file.

AdamBark:
> > I was wondering whether it was better to read the defaults from a file. Should I fix up this PR or open a new one?
> 
> I'm not sure whether it is a good idea to expect the file `test/config/default-config.json` to be there. Until now, Pebble was working fine without that file existing if you pointed it to an alternate configuration file.

I could swap the `FailOnError` for just a notification?

felixfontein:
I personally would explicitly set the values (as in a previous version of the PR), and remove all values having the default settings from the test config files. That's pretty much incompatible to what @aarongable wrote though :)

aarongable:
It's not wholly incompatible -- it's fine with me if the default values are _only_ specified in code and the default config JSON doesn't exist at all. My primary goal is that they not be specified in two places (and therefore have to be kept in sync). I just think that that solution is likely to be more confusing in the long run, as it is less "self-documenting" for consumers who do want to change values. But if it has other material advantages -- for example, not requiring the default-config.json file to exist at all -- then it is worth considering.

AdamBark:
How about having the defaults baked in and command line flags to change them?

aarongable:
I'd prefer not to change the CLI of pebble at this time. A change like this imo isn't worth an API-breaking change; sticking with the existing `-config` flag and config file format seems like the best way forward.

One possibility here would be for `getDefaultConfig` to _try_ to read the config file. If it fails to find the config file; that's fine, just return an empty config. If it finds the config file but fails to parse it, that's a problem, error out. That way someone who doesn't have the default config file on disk gets the same behavior they've always had: they have to specify everything in their config or else they'll end up with an incomplete config. But someone who wants to only customize a few things can just make sure they have the default file around, and then keep their custom config file simpler.

aarongable:
I'm not sure that I agree that it is a breaking change -- the behavior when no flags are passed is identical, and when specific flags are passed, it is the user's responsibility to ensure that those flags are good.

That said, in order to correctly change the file to default-config.json, the README.md and docker-compose.yml also need to be updated.

PR #320, opened 2020-08-15
Add PEBBLE_CHAIN_LENGTH.

### LINKED ISSUE #318: Add alternate chains of varying path length
@jvanasco suggests in [this forum post](https://community.letsencrypt.org/t/workflow-for-using-alternate-link-relation/129321/7?u=_az) that ACME clients might not be able to handle downloading a certificate chain that is longer than just end-entity+issuer.

We could add this variation to strict mode.

These chains do happen in real life (check what certificates are sent by `gogetssl.com` for example) and since pretty much the entire point of Pebble is to be a stickler for the rules, I think it should be added.

alexzorin:
This commit adds a way to control how long the certificate issuance
chains that Pebble uses are. The default and minimum value of 1 means
that there is one intermediate certificate in the final certificate
chain.

Fixes #318.

--- 

The resulting certificate chain (at a value of `2`) looks like this:

    $ cat b.pem | openssl crl2pkcs7 -nocrl -certfile /dev/stdin | openssl pkcs7 -print_certs -noout
    subject=CN = xfoobazx.com

    issuer=CN = Pebble Intermediate CA 2f1dd1

    subject=CN = Pebble Intermediate CA 2f1dd1

    issuer=CN = Pebble Intermediate CA 5a564f

    subject=CN = Pebble Intermediate CA 5a564f

    issuer=CN = Pebble Root CA 0335fe

The change is a little bit awkward because Pebble has been, in parts, written with an assumption about there being one intermediate.

The additional intermediates are not exposed via the management interface. It might be possible to make a non-breaking change to `:15000/intermediate-keys/`, but I haven't been able to think of a use-case for exposing them.

### DIFF
--- a/ca/ca.go
+++ b/ca/ca.go
@@ -202,7 +202,7 @@ func (ca *CAImpl) newChain(intermediateKey crypto.Signer, intermediateSubject pk
 	// If numIntermediates is only 1, then no intermediates will be generated here.
 	prev := root
 	intermediates := make([]*issuer, numIntermediates)
-	for i := 0; i < numIntermediates-1; i++ {
+	for i := numIntermediates - 1; i > 0; i-- {
 		k, ski, err := makeKey()
 		if err != nil {
 			panic(fmt.Sprintf("Error creating new intermediate issuer: %v", err))
@@ -222,12 +222,7 @@ func (ca *CAImpl) newChain(intermediateKey crypto.Signer, intermediateSubject pk
 	if err != nil {
 		panic(fmt.Sprintf("Error creating new intermediate issuer: %s", err.Error()))
 	}
-	intermediates[numIntermediates-1] = intermediate
-
-	// Reverse the order of the intermediates, from leaf signer to root.
-	for i, j := 0, len(intermediates)-1; i < j; i, j = i+1, j-1 {
-		intermediates[i], intermediates[j] = intermediates[j], intermediates[i]
-	}
+	intermediates[0] = intermediate
 
 	return &chain{
 		root:          root,
--- a/ca/ca.go
+++ b/ca/ca.go
@@ -35,8 +35,8 @@ type CAImpl struct {
 }
 
 type chain struct {
-	root         *issuer
-	intermediate *issuer
+	root          *issuer
+	intermediates []*issuer
 }
 
 type issuer struct {
@@ -138,8 +138,8 @@ func (ca *CAImpl) makeRootCert(
 		DER:  der,
 	}
 	if signer != nil && signer.cert != nil {
-		newCert.Issuers = make([]*core.Certificate, 1)
-		newCert.Issuers[0] = signer.cert
+		newCert.IssuerChains = make([][]*core.Certificate, 1)
+		newCert.IssuerChains[0] = []*core.Certificate{signer.cert}
 	}
 	_, err = ca.db.AddCertificate(newCert)
 	if err != nil {
@@ -163,7 +163,7 @@ func (ca *CAImpl) newRootIssuer() (*issuer, error) {
 		return nil, err
 	}
 
-	ca.log.Printf("Generated new root issuer with serial %s and SKI %x\n", rc.ID, subjectKeyID)
+	ca.log.Printf("Generated new root issuer %s with serial %s and SKI %x\n", rc.Cert.Subject, rc.ID, subjectKeyID)
 	return &issuer{
 		key:  rk,
 		cert: rc,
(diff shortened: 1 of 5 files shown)


felixfontein:
@alexzorin one use of exposing the intermediate keys and certs via the management API is that it is possible to implement an OCSP responder for all certificates in the chain, and not just for the leaf certificates. (I'm not sure whether the OCSP URL is actually included in the intermediate/root certificates if it is provided, though.)

ca/ca.go
```diff
@@ -186,18 +186,52 @@ func (ca *CAImpl) newIntermediateIssuer(root *issuer, intermediateKey crypto.Sig
 	}, nil
 }
 
-func (ca *CAImpl) newChain(intermediateKey crypto.Signer, intermediateSubject pkix.Name, subjectKeyID []byte) *chain {
+// newChain generates a new issuance chain, including a root certificate and numIntermediates intermediates (at least 1).
+// The first intermediate will use intermediateKey, intermediateSubject and subjectKeyId.
+// Any intermediates between the first intermediate and the root will have their keys and subjects generated automatically.
+func (ca *CAImpl) newChain(intermediateKey crypto.Signer, intermediateSubject pkix.Name, subjectKeyID []byte, numIntermediates int) *chain {
+	if numIntermediates <= 0 {
+		panic("At least one intermediate must be present in the certificate chain")
```
felixfontein, line 194: Hmm, why not also support zero intermediates? But I guess that could also be added later (in another PR), because right now there's probably too much code which assumes there is at least one (like the management interface).

ca/ca.go
```diff
@@ -186,18 +186,52 @@ func (ca *CAImpl) newIntermediateIssuer(root *issuer, intermediateKey crypto.Sig
...
+	// The last N-1 intermediates build a path from the root to the leaf signing certificate.
+	// If numIntermediates is only 1, then no intermediates will be generated here.
+	prev := root
+	intermediates := make([]*issuer, numIntermediates)
+	for i := 0; i < numIntermediates-1; i++ {
+		k, ski, err := makeKey()
+		if err != nil {
+			panic(fmt.Sprintf("Error creating new intermediate issuer: %v", err))
+		}
+		intermediate, err := ca.makeRootCert(k, pkix.Name{
```
felixfontein, line 210: Why aren't you using `newIntermediateIssuer` here?
```suggestion
		intermediates[i], err := ca.newIntermediateIssuer(prev, k, pkix.Name{
```

alexzorin:
>I'm not sure whether the OCSP URL is actually included in the intermediate/root certificates if it is provided, though

Doesn't look like it - only in the leaf right now.

felixfontein:
From @jsha's [post](https://community.letsencrypt.org/t/detailed-2020-hierarchy/131019/1) it looks like intermediates are no longer required to have OCSP URLs anyway, so I guess it's fine. In any case, we can still add an API to retrieve the other chain certificates / keys if someone actually needs them.

ca/ca.go
```diff
@@ -179,25 +179,59 @@ func (ca *CAImpl) newIntermediateIssuer(root *issuer, intermediateKey crypto.Sig
...
+	// If numIntermediates is only 1, then no intermediates will be generated here.
+	prev := root
+	intermediates := make([]*issuer, numIntermediates)
+	for i := 0; i < numIntermediates-1; i++ {
+		k, ski, err := makeKey()
+		if err != nil {
+			panic(fmt.Sprintf("Error creating new intermediate issuer: %v", err))
+		}
+		intermediate, err := ca.newIntermediateIssuer(prev, k, pkix.Name{
+			CommonName: intermediateCAPrefix + hex.EncodeToString(makeSerial().Bytes()[:3]),
```
rolandshoemaker, line 211: Would probably be useful to specify what position in the chain this intermediate is, i.e. `fmt.Sprintf("%s %d", intermediateCAPrefix, i)` and skip the random suffix.

ca/ca.go
```diff
@@ -179,25 +179,59 @@ func (ca *CAImpl) newIntermediateIssuer(root *issuer, intermediateKey crypto.Sig
...
+
+	// The first issuer is the one which signs the leaf certificates
+	intermediate, err := ca.newIntermediateIssuer(prev, intermediateKey, intermediateSubject, subjectKeyID)
 	if err != nil {
 		panic(fmt.Sprintf("Error creating new intermediate issuer: %s", err.Error()))
 	}
+	intermediates[numIntermediates-1] = intermediate
+
+	// Reverse the order of the intermediates, from leaf signer to root.
+	for i, j := 0, len(intermediates)-1; i < j; i, j = i+1, j-1 {
```
rolandshoemaker, line 228: This could be skipped by just reversing the direction of the for loop above, i.e. `for i := numIntermediates-1; i > 0; i--` (when constructing the common name you need to be careful you inverse the index so the numbering makes sense). You'd then stick the last intermediate in `intermediates[0]`.

ca/ca.go
alexzorin: I don't think that eliminating the random suffix would combine well with `PEBBLE_ALTERNATE_ROOTS`. We would end up with a "CN=Pebble Intermediate CA 2" in every chain, suggesting they're cross-signs, but they're actually totally unrelated with different subject keys.

What if we logged each issuer chain on startup, instead of logging the generation of each issuer? Helps the user while keeping opaque things opaque.

felixfontein: How about having both the random suffix and the position? I.e. `Pebble Intermediate CA 3 2a9423`? Or even include the alternative root number, so we have `Pebble Root CA 2 3a74f` and `Pebble Intermediate CA 2:3 2a9423`?

rolandshoemaker: I'm not strongly opinionated either way.

Another option would be to generate a unique chain ID, so with two chains you'd have i.e. `root (2a9423) -> intermediate 1 (2a9423) -> intermediate 2 (2a9423) -> leaf` and `root (84799db) -> intermediate 1 (84799db) -> intermediate 2 (84799db) -> leaf`

alexzorin: The leaf's issuer would need to have a suffix-less name ("Pebble Intermediate XYZ") as it must have the same subject across every chain.

I'll try experiment with a chain ID for all the issuers except that one, but otherwise, I'm probably just going to log the issuance chains and keep the names as-is.

ca/ca.go
```diff
@@ -179,26 +191,61 @@ func (ca *CAImpl) newIntermediateIssuer(root *issuer, intermediateKey crypto.Sig
 	if err != nil {
 		return nil, err
 	}
-	ca.log.Printf("Generated new intermediate issuer with serial %s and SKI %x\n", ic.ID, subjectKeyID)
+	ca.log.Printf("Generated new intermediate issuer %s with serial %s and SKI %x\n", ic.Cert.Subject, ic.ID, subjectKeyID)
```
alexzorin, line 194: Pebble's startup logging is imo too spammy now.

    Pebble 2020/08/20 20:47:56 Generated new root issuer CN=Pebble Root CA 6ee3a9 with serial 7763e13ec3f24991 and SKI 62675f56d68b4a74e93aaed38bfbfdf375b2480a
    Pebble 2020/08/20 20:47:57 Generated new intermediate issuer CN=Pebble Intermediate CA 6ee3a9 #1 with serial 1368b1b7a776a696 and SKI fd393217f77da95999769563c8db1d24af99a3c2
    Pebble 2020/08/20 20:47:57 Generated new intermediate issuer CN=Pebble Intermediate CA 335fb8 with serial 5413643aad42ad63 and SKI bcc3646f6e7bc060462ea54556c697784395c128
    Pebble 2020/08/20 20:47:57 Generated issuance chain: Pebble Root CA 6ee3a9 -> Pebble Intermediate CA 6ee3a9 #1 -> Pebble Intermediate CA 335fb8
    Pebble 2020/08/20 20:47:57 Generated new root issuer CN=Pebble Root CA 31d5c2 with serial 7d145f4fb93c1812 and SKI 6569baa2cf24f15a95c22a9939e6bce45536c76b
    Pebble 2020/08/20 20:47:57 Generated new intermediate issuer CN=Pebble Intermediate CA 31d5c2 #1 with serial 09f5bc1b9af57921 and SKI 4179a17e4ecfe06a98a659c708966dacdbaf0386
    Pebble 2020/08/20 20:47:57 Generated new intermediate issuer CN=Pebble Intermediate CA 335fb8 with serial 662fb5ce4a0bb6bb and SKI bcc3646f6e7bc060462ea54556c697784395c128
    Pebble 2020/08/20 20:47:57 Generated issuance chain: Pebble Root CA 31d5c2 -> Pebble Intermediate CA 31d5c2 #1 -> Pebble Intermediate CA 335fb8

My personal preference would be to only log the chain and not the individual issuers being generated. But maybe others find it useful, so I welcome a comment on that.

felixfontein: I prefer it to be a bit on the too spammy side than having it output too litle. This way, I can always scroll up in the CI logs to see which names are used in all involved intermediate/root certs, if problems arise. And the default values of `PEBBLE_CHAIN_LENGTH` and `PEBBLE_ALTERNATE_ROOTS` are so that there's no spam in the default config.

felixfontein (approved):
Current version still works fine with my client's CI tests.

ca/ca.go
alexzorin: Fully agree that the names should be logged - those would stay as part of the "Generated issuance chain" line. I was more wondering about the individual logging of each serial and SKI. If those are useful, I won't touch it.

jsha:
Looks great, thanks for writing this up! BTW, what are your thoughts on offering chains > 1 by default?

alexzorin:
>Looks great, thanks for writing this up! BTW, what are your thoughts on offering chains > 1 by default?

Don't know. I'm not sure whether it would a useful default, as clients will just ignore it if they're not aware of it.

PR #322, opened 2020-08-26
Use rand.Seed to prevent predictable nonce rejections

### LINKED ISSUE #321: Authorization reuse and nonce rejection use deterministic "rand"
According to the README, "Pebble will reuse valid authorizations in new orders, if they exist, 50% of the time" and `PEBBLE_AUTHZREUSE` can be set to change the probability. The relevant code agrees with this: https://github.com/letsencrypt/pebble/blob/4670b15b155dbd5138d5cbecf4966e497e9d48ac/wfe/wfe.go#L1489 The same applies to nonce rejection.

However, it seems that nothing in Pebble ever calls `rand.Seed()` and thus the PRNG emits the same sequence of values every run. (See https://golang.org/pkg/math/rand/#Seed for the documented behaviour.)

This can be a problem for automated test suites (like mine) that always make the same sequence(s) of calls to Pebble with very few opportunities for authorization reuse because multiple test runs (with a freshly-started Pebble for each) don't increase the chances of uncovering authorization reuse or nonce rejection bugs.

Given that Pebble is explicitly a testing server and generates new CA credentials every run, I think it would make sense to seed `rand` at startup so that every test run provides new opportunities to uncover bugs.

aarongable:
Fixes #321

### DIFF
--- a/README.md
+++ b/README.md
@@ -232,7 +232,7 @@ failed request, rather than quitting outright.
 
 Experience from Boulder indicates that many ACME clients do not gracefully retry
 on invalid nonce errors. To help ensure future ACME clients are able to
-gracefully handle these errors by default **Pebble rejects 15% of all valid
+gracefully handle these errors by default **Pebble rejects 5% of all valid
 nonces as invalid**.
 
 The percentage of valid nonces that are rejected can be configured using the
--- a/wfe/wfe.go
+++ b/wfe/wfe.go
@@ -165,6 +165,10 @@ func New(
 	va *va.VAImpl,
 	ca *ca.CAImpl,
 	strict, requireEAB bool) WebFrontEndImpl {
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -232,7 +232,7 @@ failed request, rather than quitting outright.
 
 Experience from Boulder indicates that many ACME clients do not gracefully retry
 on invalid nonce errors. To help ensure future ACME clients are able to
-gracefully handle these errors by default **Pebble rejects 15% of all valid
+gracefully handle these errors by default **Pebble rejects 5% of all valid
```
felixfontein, line 235: Good catch!

felixfontein (approved):

felixfontein:
Ha, two reviews before the initial CI run finished. I guess that's a new record for this repo ;-)

Issue #324, opened 2020-09-21
Boulder Divergences: Duplicate Orders and Authorizations


jvanasco:
There are several divergences in the behavior of Boulder and Pebble.

Divergence 1- Authorizations

* Account A submits an order for `a.example.com`. No challenge is triggered.
* Account A submits a second order for `a.example.com`

Pebble will generate a Unique Authorization (Unique Authorization URL + Unique Challenges) for each order.
Boulder will use the same Authorization URL + Challenges for both orders.

Divergence 2- Orders

* Account A submits an order for `a.example.com` or `a.example.com+b.example.com`. One challenge is triggered and fails.
* Account A submits a second order for `a.example.com, b.example.com`.

Pebble will generate a Unique Order URL for each order.
Boulder is unpredictable and may re-use the Order URL.  This seems to happen on single domain orders too, but this has been unpredictable in my tests and hard to reproduce.

Just to be clear: I am not making an opinion on the behavior of either server. These are significant divergences, and are not documented. The Pebble documentation teaches away from expecting this, as it only speaks of similar divergences to Boulder in regards to re-using validated authorizations, not pending ones.

Edit: After additional testing, Boulder seems to re-use the Order URL with 'random domains', but not with "known" domains. Either triggering a challenge or touching/deactivating an authorization appears to change something and issue a new order URL.

felixfontein:
Why should that be an issue?

jvanasco:
@felixfontein this submitted by accident, without all the text.  refiling.

aarongable:
Out of curiosity, why are these divergences seen as significant? In what ways would a client's behavior differ based on whether a given authorization or order object was re-used?

Interestingly, re-use of these objects is neither specifically allowed nor specifically forbidden by the ACME spec. Where possible, I'd suggest that clients always behave as though a received authorization or order object is brand new.

jvanasco:
> Where possible, I'd suggest that clients always behave as though a received authorization or order object is brand new.

That is exactly what I did with the Order, and what caused our problem.  The correct assumption is to behave as if anything might reused.  Lesson learned!

> Interestingly, re-use of these objects is neither specifically allowed nor specifically forbidden by the ACME spec.

[...]

PR #325, opened 2020-09-25
updated readme with reuse information

jvanasco:
Re #324 

As these are the only divergences I know of, I just altered the "authorization reuse" section, instead of making a "divergences" section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,12 +251,13 @@ The RFC allows for several objects to be re-used.
 
 **Clients should be prepared an ACME server may re-use any given object type, regardless of Pebble implementing a reuse policy for that object.**
 
+Pebble and Boulder __may__ or __may not__ implement the same object re-use policies at any given time.  There exists an [ACME Implementation Details](https://github.com/letsencrypt/boulder/blob/main/docs/acme-implementation_details.md) document for Boulder which contains some information on how Boulder handles this.
+
 #### Order Reuse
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -245,16 +245,36 @@ To **never** reject a valid nonce as invalid run:
...
+The RFC allows for several objects to be re-used.
+
+**Clients should be prepared an ACME server may re-use any given object type, regardless of Pebble implementing a reuse policy for that object.**
+
+#### Order Reuse
+
+The RFC allows ACME servers to reuse orders. Pebble does not reuse orders at this time; however Boulder does reuse Orders in at least two scenarios:
+
+1. If an Account requests a new Order that is identical to an already existing "pending" order for that same Account, the Order will be re-used.
+2. If an Account requests a new Order that is identical to a recently failed "invald" order for that same Account, the Order will be re-used and return to a "pending" state.
```
jsha, line 259: This (2) is not accurate. The RFC does not allow Orders to go backwards from invalid to pending, and Boulder doesn't implement that.

For (1) I would say "...already existing "pending" or "ready" order."

By the way, I'm okay with landing this documentation here but it might be better to document it in Boulder and link to that from here. I'd welcome a PR!

jvanasco: i will adjust and PR to both!

jvanasco:
updated with references to new implementation details document in Boulder

jsha (approved):
Thanks! And sorry for the very long lag on getting this merged. :-)

PR #329, opened 2020-10-21
Add support for NotBefore and NotAfter in Order

meyskens:
When testing against Pebble I found it doesn't reject NotBefore and NotAfter in orders but it wasn't using them causing to have confusing results.

This PR adds support for issuing a cert with NotBefore/NotAfter set to specific values set in the order.
Feedback welcome :)


jsha (approved):
Welcome to the project! This change looks good.

PR #331, opened 2020-11-13
Add a blocklist for Order identifiers

meyskens:
This adds a small version of the domain policies of boulder in the version of a block list which can deny certain domains (or IPs) with the `rejectedIdentifier` error.
This allows to tes the behaviour of having an order rejected by policy rules.

Let me know if it needs some changes, tests or documentation 😄

### DIFF
--- a/cmd/pebble/main.go
+++ b/cmd/pebble/main.go
@@ -27,7 +27,7 @@ type config struct {
 		ExternalAccountBindingRequired bool
 		ExternalAccountMACKeys         map[string]string
 		// Configure policies to deny certain domains
-		BlockList []string
+		DomainBlocklist []string
 	}
 }
 
@@ -78,7 +78,7 @@ func main() {
 		cmd.FailOnError(err, "Failed to add key to external account bindings")
 	}
 
-	for _, domainName := range c.Pebble.BlockList {
+	for _, domainName := range c.Pebble.DomainBlocklist {
 		err := db.AddBlockedDomain(domainName)
 		cmd.FailOnError(err, "Failed to add domain to block list")
 	}
--- a/cmd/pebble/main.go
+++ b/cmd/pebble/main.go
@@ -26,6 +26,8 @@ type config struct {
 		// Require External Account Binding for "newAccount" requests
 		ExternalAccountBindingRequired bool
 		ExternalAccountMACKeys         map[string]string
+		// Configure policies to deny certain domains
+		BlockList []string
 	}
 }
 
@@ -74,6 +76,11 @@ func main() {
 	for keyID, key := range c.Pebble.ExternalAccountMACKeys {
 		err := db.AddExternalAccountKeyByID(keyID, key)
 		cmd.FailOnError(err, "Failed to add key to external account bindings")
+	}
+
+	for _, domainName := range c.Pebble.BlockList {
+		err := db.AddBlockedDomain(domainName)
+		cmd.FailOnError(err, "Failed to add domain to block list")
 	}
 
 	wfeImpl := wfe.New(logger, db, va, ca, *strictMode, c.Pebble.ExternalAccountBindingRequired)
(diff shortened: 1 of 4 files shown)


cmd/pebble/main.go
```diff
@@ -26,6 +26,8 @@ type config struct {
 		// Require External Account Binding for "newAccount" requests
 		ExternalAccountBindingRequired bool
 		ExternalAccountMACKeys         map[string]string
+		// Configure policies to deny certain domains
+		BlockList []string
```
aarongable, line 30: Since config files are part of the "public API" of Pebble, make this name more descriptive, like "IdentifierBlocklist" or "DomainBlocklist", just in case someone wants to add an account key blocklist or similar later on.

aarongable (changes requested):
Thanks so much for this PR! Just a few small comments inline, and one high-level comment here:

The intent and names used (e.g. `blocklistByDomain`) here suggest that the purpose of this is to allow users to configure a list of blocked DNS identifiers. But the implementation actually allows users to block any kind of identifier: the Add and Get methods are just doing untyped string comparisons, and the `GetBlockedDomain` check inside `verifyOrder` happens outside any conditionals which check what kind of identifier is currently being processed.

As such, I'd suggest either:
a) Changing all of the method and variable names to make it clear that this is a generic Identifier blocklist, not just a domain name blocklist; or
b) Moving the call to check the blocklist down to line wfe.go:1424, after we already know that it isn't of type `IdentifierIP` (because we would have already returned or continued if it were).

db/memorystore.go
```diff
@@ -437,3 +440,25 @@ func (m *MemoryStore) GetExtenalAccountKeyByID(keyID string) ([]byte, bool) {
 	key, ok := m.externalAccountKeysByID[keyID]
 	return key, ok
 }
+
+// AddBlockedDomain will add the domain name to the block list
+func (m *MemoryStore) AddBlockedDomain(name string) error {
```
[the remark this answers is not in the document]
meyskens: Great idea!

meyskens:
Thanks for the quick review @aarongable 
I took on the comments and made it more domain specific implementing it for domains so I can do the reversing, an IP specific implementation would be nice to follow later (thinking about subnetting there for example)

Feedback on the logic more than welcome!

db/memorystore.go
```diff
@@ -437,3 +441,51 @@ func (m *MemoryStore) GetExtenalAccountKeyByID(keyID string) ([]byte, bool) {
...
+	}
+
+	for _, blockedParts := range m.blockListByDomain {
+		isMatch := true
+		for i := range blockedParts {
+			if blockedParts[i] != domainParts[i] {
+				isMatch = false
+				break
+			}
+		}
```
aarongable, line 484: An alternate approach here would be to store the blocked domains as strings, concatenate domainParts back into a single string, and then do a prefix check: https://golang.org/pkg/strings/#HasPrefix

I'm honestly not sure which approach is preferable here. It would avoid this loop here, and the logic might be more readable, but it likely has weirder edge cases (e.g. if "example.com" is blocked, we should prevent issuance for "foo.example.com" but not for "fooexample.com"). I'd be happy with either; just wanted to put this idea out there in case you liked it.

aarongable (dismissed):
Looks good! Two small nits and one potential alternate suggestion which you're free to ignore.

db/memorystore.go
meyskens: I wrote it exactly like that till my head made me think about the domain prefix being problematic as you said. I was thinking about ways around it by adding more dots etc but always ended up with even more complex issues.

meyskens:
`meyskens dismissed aarongable’s stale review via e638f39 7 minutes ago`

that is new 👀

aarongable:
Heh yeah, that's what happens when you have certain kinds of branch protection enabled:
image

Thanks again for this PR!

aarongable (approved):

meyskens:
This is interesting! should enable that on our repos too :)

PR #334, opened 2020-12-22
wfe: surface errors related to JWS parsing

alexzorin:
This is very much a matter of opinion, but given Pebble's role as a development server, I think it would help to expose more error detail where possible.

JWS can be particularly tedious to debug. I feel that this:

```json
{
   "type": "urn:ietf:params:acme:error:malformed",
   "detail": "Parse error reading JWS: square/go-jose: invalid embedded jwk, must be public key",
   "status": 400
}
```

is a better experience than:

```json
{
   "type": "urn:ietf:params:acme:error:malformed",
   "detail": "Parse error reading JWS",
   "status": 400
}
```


jsha (approved):
Good idea, thanks!

PR #337, opened 2021-02-09
Add an option to listen on HTTP.

ralt:
This is needed when deploying Pebble behind other proxies doing TLS
termination. In my use case, I've been deploying a test instance on a
PaaS hosting platform that provides the TLS termination in front of
Pebble, so I have to make it listen on HTTP only.

### DIFF
--- a/README.md
+++ b/README.md
@@ -272,6 +272,11 @@ store or to any production systems/codebases. The private key for this CA is
 intentionally made [publicly available in this
 repo](test/certs/pebble.minica.key.pem).**
 
+### Listening on HTTP
+
+In order to make Pebble accessible over HTTP, you can set the
+`listenHttp` setting of the `pebble-config.json` file.
+
 ### Management interface
 
 In order to ease the interaction of Pebble with testing systems, a specific HTTP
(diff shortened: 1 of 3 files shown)


aarongable:
Hi ralt, and thanks for the PR!

Unfortunately for your use-case, use of HTTPS is required as per [RFC 8555 Section 6.1](https://tools.ietf.org/html/rfc8555#section-6.1):

> Each ACME function is accomplished by the client sending a sequence
   of HTTPS requests to the server [RFC2818], carrying JSON messages
   [RFC8259].  Use of HTTPS is REQUIRED.

Section 10 has further discussion of the security considerations, including a potential MitM attack against the ACME requests channel (e.g. if your TLS-terminating proxy was malicious). Although the RFC puts various mitigations in place to prevent such MitM attacks, it still requires that the CA be an HTTPS service.

Because of the above, and because one of the explicit goals of pebble is to remain as simple as possible, I think at this time we'd like to keep it to HTTPS-only. It's somewhat surprising that your PaaS supports neither HTTPS as a backend, nor packet-level proxying?

ralt:
From the client's POV, it is definitely HTTPS, so the RFC is respected, I'd say. I was also assuming that pebble is a test-bed, and as such considerations such as MiTM attacks were less of an issue.

But it's fine, I thought this would be useful to other people, as surely I'm not the only one hitting those sort of limitations in hosting platforms. If you say no, then so be it. :)

PR #338, opened 2021-03-06
Follow CNAME records for DNS challenge

andrewda:
The ACME protocol allows CNAMEs from `_acme-challenge` subdomains to domains with the proper challenge tokens. For example, the following configuration would be valid:

```
_acme-challenge.example.com		CNAME	test1.example.com.
test1.example.com				TXT		srFQeeM4d8iP7FpzGcxrc4mmRtKdJ_uJJBo5DlWHYSs
```

Pebble should abide by this by following CNAME records and reading any TXT records from the target domain.

This feature is used extensively by tools such as https://github.com/joohoi/acme-dns.

### DIFF
--- a/va/va.go
+++ b/va/va.go
@@ -596,12 +596,37 @@ func (va VAImpl) getTXTEntry(name string) ([]string, error) {
 	}
 
 	if in.Rcode != dns.RcodeSuccess {
-		return nil, fmt.Errorf("DNS lookup for %q returned an unsuccessful response: %q", name, in.Rcode)
+		return nil, fmt.Errorf("DNS TXT lookup for %q returned an unsuccessful response: %q", name, in.Rcode)
 	}
 
 	for _, record := range in.Answer {
 		if t, ok := record.(*dns.TXT); ok {
 			txts = append(txts, t.Txt...)
+		}
+	}
+
+	// If no TXT records are present, follow CNAME
+	// Per RFC 1034, no CNAME exists if a TXT record is present
+	if len(txts) == 0 {
+		message.SetQuestion(dns.Fqdn(name), dns.TypeCNAME)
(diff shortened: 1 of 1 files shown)


felixfontein:
Hmm, is this actually necessary? The documentation of `getTXTEntry` says:
```
// getTXTEntry fetches TXT entries for the given domain name using the recursive resolver located at
// `va.customResolverAddr`, or the default system resolver if no custom resolver addr is specified
```
So it assumes that the custom resolver is a *recursive* resolver, and that one is supposed to follow CNAMEs automatically.

felixfontein:
Also, it looks like CI is currently broken (independent of this PR): certbot seems to no longer support Python 3.5 and 2.7, and that seem to be the Python versions available in CI.

andrewda:
You're likely correct for a production setup, but many test/local DNS servers don't have recursive capabilities. Specifically, my testing setup uses `dnsmasq` and https://github.com/joohoi/acme-dns -- `dnsmasq` to create the CNAME records and `acme-dns` as a nameserver to resolve the TXT records. `dnsmasq` doesn't resolve recursively, so the TXT records are never found by pebble, unless something like this is added.

I might try replacing `dnsmasq` with Unbound or BIND to see if I can get either of those to work recursively.

jsha:
I think replacing `dnsmasq` with Unbound or BIND in your test setup is the right move here. Unbound is pretty easy to set up! If you don't like the packaged version provided by your OS, I have a very small config for a standalone user-mode unbound as part of unboundtest: https://github.com/jsha/unboundtest

jsha:
Closing. As discussed above, we shouldn't implement CNAME chasing in Pebble. That's the recursive resolver's job.

PR #348, opened 2021-06-15
Randomize the account ID generated

### LINKED ISSUE #340: Feature Request - Random Account URLs
The account URLs appear to be sequential based.  It would be useful if the urls were randomized, or potentially randomized by an environment variable.  Alternatively, the URLs could contain a hash of the contact email address.  (These are just the two simplest ideas i had)

     -  https://0.0.0.0:14000/my-account/1
     +  https://0.0.0.0:14000/my-account/c556d4  # where `c556d4` is random

or 

     -  https://0.0.0.0:14000/my-account/1
     +  https://0.0.0.0:14000/my-account/1/7fc31f   # where `7fc31f` is a hash of the contact

Use Case:

Our test systems typically use an isolated pebble server per test suite in an attempt to try and break things, then reconcile data with the server if possible or handle "this is broken!".  This works perfectly, except for accounts.

When restarting the pebble server, the first account is always "https://0.0.0.0:14000/my-account/1". This creates an issue as, in the context of a pebble test server, this URL represents a uniquely new account -- but in the context of a production boulder server with persistent storage, this url represents an existing account that needs to be reconciled on our end.

While a workaround is to alter the database items after tests, this is less than ideal.

squizzling:
This also switches account IDs to hex, so there's no assumptions about them being integers.

Closes #340

### DIFF
--- a/db/memorystore.go
+++ b/db/memorystore.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 	"math/big"
+	"math/rand"
 	"reflect"
 	"strconv"
 	"strings"
@@ -37,7 +38,7 @@ func (e ExistingAccountError) Error() string {
 type MemoryStore struct {
 	sync.RWMutex
 
-	accountIDCounter int
+	accountRand *rand.Rand
 
 	accountsByID map[string]*core.Account
 
@@ -62,7 +63,7 @@ type MemoryStore struct {
 
 func NewMemoryStore() *MemoryStore {
 	return &MemoryStore{
-		accountIDCounter:        1,
+		accountRand:             rand.New(rand.NewSource(time.Now().UnixNano())),
 		accountsByID:            make(map[string]*core.Account),
 		accountsByKeyID:         make(map[string]*core.Account),
 		ordersByID:              make(map[string]*core.Order),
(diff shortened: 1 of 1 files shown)


aarongable (commented):
Thanks! LGTM in general, but I'd like a second set of Boulder eyes on this. @jsha?

In particular, I think that this approach is clean, simple, and accomplishes the goal. That said, Boulder (though not RFC 8555) does in fact guarantee that account IDs will be integers, and should never fail with "account already exists" because the ID is generated at the database layer via an auto-increment column. So to most accurately emulate Boulder, perhaps the right thing to do would be to generate integers, and if the check on line 129 fails, try again. But maybe that level of fidelity isn't really what we want or need here.

felixfontein:
I would avoid generating integers (since RFC 8555 does not promise it, it's good that we have a test server which is more liberal than boulder :) ). I'd also add a repeat logic for the check in line 129, so that `account %q already exists` never happens. Getting such an error would look like a server error to me; creating a new account should never result in that error.

squizzling:
Groovy.  I'll get Travis working, then get the loop going and update this PR.  CI > features.

squizzling:
A re-run of the CI test should fix this (#351)

jsha (approved):
This looks good to me. I agree generating random strings rather than integers is even nicer, but this is a step forward and doesn't get in the way of doing better later.

jvanasco:
Thank you everyone for the work on developing and supporting this PR!  This will allow me to remove some very unfortunate code from a few test harnesses!

PR #353, opened 2021-08-08
Add env var to set certificate lifetime

osirisinferi:
This commit/PR adds the `PEBBLE_CERT_LIFETIME` environment variable to be able to set the end leaf certificate lifetime.

Currently, it uses the current 5 years as default, but does not have an upper limit. No idea what happens if the user sets `PEBBLE_CERT_LIFETIME` to something HUGE or very strange (like a non-integer variable), pebble probably crashes.

Mainly for personal use, but perhaps you might be interested :slightly_smiling_face: 

An alternative by using the configuration file instead of an environment variable can be found in #361


aarongable:
Superceded by #361

PR #355, opened 2021-08-14
Copy over extensions (e.g. must-staple) from offered CSR into final cert

osirisinferi:
Just a tiny simple modification to Pebble so if the client requests a certain extension in the CSR, such as must-staple, it will end up in the certificate too.

This code doesn't check the requested extensions in the orders CSR, it just c/ps it.. My Golang skills aren't good enough for any of that and frankly I'm not going/willing to learn Go either. I just wanted this feature in Pebble so I could work on some must-staple features in certbot, so I implemented this.

### DIFF
--- a/ca/ca.go
+++ b/ca/ca.go
@@ -248,7 +248,7 @@ func (ca *CAImpl) newChain(intermediateKey crypto.Signer, intermediateSubject pk
 	return c
 }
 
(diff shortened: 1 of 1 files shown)


aarongable:
Copying extensions has historically been one of the most common causes of misissuance. Boulder categorically refuses to do so: it only looks for the presence of a small number of extensions in the CSR (e.g. SANs, must-staple), and then re-creates those in the final certificate from the group up. While respecting must-staple would be a great thing for Pebble to have, it definitely shouldn't copy over extensions blindly.

osirisinferi:
Noted :slightly_smiling_face:

PR #357, opened 2021-08-16
Use snap instead of certbot-auto in travis CI

### LINKED ISSUE #354: Travis relies on certbot-auto causing Travis to fail
It seems the Travis CI uses the deprecated `certbot-auto` command:

However, as we all know, `certbot-auto` has been [deprecated](https://github.com/certbot/certbot/pull/8489) and consequently [removed](https://github.com/certbot/certbot/pull/8938). This causes Travis to fail:

```
$ git clone --single-branch --depth=1 -b master https://github.com/certbot/certbot
$ cd certbot
$ ./certbot-auto --os-packages-only -n
/home/travis/.travis/functions: line 109: ./certbot-auto: No such file or directory
The command "./certbot-auto --os-packages-only -n" failed and exited with 127 during .
```

aarongable:
Fixes #354

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ services:
   - docker
 
 before_install:
-  - sudo apt-get install snapd
+  - sudo apt-get install snapd python3-josepy
   - sudo snap install core
   - sudo snap refresh core
   - sudo snap install --classic certbot
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,19 +19,19 @@ services:
 
 before_install:
   - sudo apt-get install snapd python3-venv
-  # We need certbot to be available on $PATH, so we use the recommended (snap)
-  # installation method.
-  - sudo snap install core
-  - sudo snap refresh core
-  - sudo snap install --classic certbot
-  - sudo ln -s /snap/bin/certbot /usr/bin/certbot
-  # We also need the `acme` and `josepy` python libraries for chisel2.py, so we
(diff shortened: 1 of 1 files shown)


aarongable:
Looks like this fixes the certbot install issue, but it also looks like a side-effect of that certbot installation method was activating the certbot python venv, which was indirectly providing the josepy package for chisel2.py to use.

osirisinferi:
> it also looks like a side-effect of that certbot installation method was activating the certbot python venv, which was indirectly providing the josepy package for chisel2.py to use.

Can't you just do a `apt-get install python3-josepy` for that?

aarongable:
That was a good idea, but looks like chisel2.py actually relies on the entirety of the `acme` python library... which is part of the certbot repo. So I guess I'm gonna do both: install certbot from snap, *and* clone the certbot repo and activate its venv.

osirisinferi:
> So I guess I'm gonna do both: install certbot from snap, _and_ clone the certbot repo and activate its venv.

I don't know which distribution Travis is running on, but I think Debian and Ubuntu both also have a package python3-acme :stuck_out_tongue: But I guess just pulling the git repo works too :wink: Don't know which is quicker/easier tho.

aarongable:
Ooh interesting, I didn't expect that! Giving it a shot, thanks.

osirisinferi:
Huh, and suddenly you don't need augeas any longer? How does that compute? :stuck_out_tongue: Ah well, the checks have passed :grin:

beautifulentropy (approved):

PR #361, opened 2021-08-22
Add certificate lifetime configuration option

osirisinferi:
An alternative to #353, probably not necessary to add both :stuck_out_tongue:

### DIFF
--- a/ca/ca.go
+++ b/ca/ca.go
@@ -349,8 +349,9 @@ func (ca *CAImpl) newCertificate(domains []string, ips []net.IP, key crypto.Publ
 
 func New(log *log.Logger, db *db.MemoryStore, ocspResponderURL string, alternateRoots int, chainLength int, certificateLifetime int) *CAImpl {
 	ca := &CAImpl{
-		log: log,
-		db:  db,
+		log:          log,
+		db:           db,
+		certLifetime: defaultCertLifetime,
 	}
 
 	if ocspResponderURL != "" {
@@ -375,11 +376,9 @@ func New(log *log.Logger, db *db.MemoryStore, ocspResponderURL string, alternate
 			ca.certLifetime = certificateLifetime
 			ca.log.Printf("Using user defined certificate lifetime of %d days", certificateLifetime)
 		} else {
-			ca.certLifetime = defaultCertLifetime
-			ca.log.Printf("certificateLifetime (%d) is not a valid value (valid values: integer > 0), using default certificate lifetime of %d days", certificateLifetime, defaultCertLifetime)
+			ca.log.Printf("certificateLifetime (%d) is not a valid value (valid values: integer > 0), using default certificate lifetime of %d days", certificateLifetime, ca.certLifetime)
 		}
 	} else {
-		ca.certLifetime = defaultCertLifetime
 		ca.log.Printf("Using default certificate lifetime of %d days", ca.certLifetime)
 	}
 
--- a/ca/ca.go
+++ b/ca/ca.go
@@ -25,6 +25,7 @@ import (
 const (
 	rootCAPrefix         = "Pebble Root CA "
 	intermediateCAPrefix = "Pebble Intermediate CA "
+	defaultCertLifetime  = 1826
 )
 
 type CAImpl struct {
@@ -33,6 +34,8 @@ type CAImpl struct {
 	ocspResponderURL string
 
 	chains []*chain
+
+	certLifetime int
 }
 
 type chain struct {
@@ -277,7 +280,7 @@ func (ca *CAImpl) newCertificate(domains []string, ips []net.IP, key crypto.Publ
 		}
 	}
 
-	certNotAfter := time.Now().AddDate(5, 0, 0)
+	certNotAfter := time.Now().AddDate(0, 0, ca.certLifetime)
 	if notAfter != "" {
 		certNotAfter, err = time.Parse(time.RFC3339, notAfter)
 		if err != nil {
@@ -340,7 +343,7 @@ func (ca *CAImpl) newCertificate(domains []string, ips []net.IP, key crypto.Publ
 	return newCert, nil
 }
 
-func New(log *log.Logger, db *db.MemoryStore, ocspResponderURL string, alternateRoots int, chainLength int) *CAImpl {
+func New(log *log.Logger, db *db.MemoryStore, ocspResponderURL string, alternateRoots int, chainLength int, certificateLifetime int) *CAImpl {
 	ca := &CAImpl{
 		log: log,
 		db:  db,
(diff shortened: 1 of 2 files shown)


ca/ca.go
```diff
@@ -25,6 +25,7 @@ import (
 const (
 	rootCAPrefix         = "Pebble Root CA "
 	intermediateCAPrefix = "Pebble Intermediate CA "
+	defaultCertLifetime  = 1826
```
aarongable, line 28: Let's express this in seconds, rather than days, because certificate validity periods are measured at second-level granularity. Let's also call it `defaultValidityPeriod` (and use "validity period" throughout the rest of the change), as that is the technical name for the thing we're controlling here.

ca/ca.go
```diff
@@ -33,6 +34,8 @@ type CAImpl struct {
 	ocspResponderURL string
 
 	chains []*chain
+
+	certLifetime int
```
aarongable, line 38: Make this a `uint` so you don't have to check for `> 0` later on.

ca/ca.go
```diff
@@ -277,7 +280,11 @@ func (ca *CAImpl) newCertificate(domains []string, ips []net.IP, key crypto.Publ
 		}
 	}
 
-	certNotAfter := time.Now().AddDate(5, 0, 0)
+	certNotAfter := time.Now().AddDate(0, 0, ca.certLifetime)
```
aarongable, line 283: Once `ca.certValidityPeriod` is measured in seconds, this can become `.Add(ca.certValidityPeriod * time.Second)`.

Also, this is unrelated to the change you're making now, but it'd be great if you did it while you're here: subtract one second, to account for the way validity periods are actually computed. So really this should be `.Add(ca.certValidityPeriod * time.Second).Add(-1 * time.Second)` (or of course you could combine them into a single operation if you think that's still clear).

ca/ca.go
```diff
@@ -277,7 +280,11 @@ func (ca *CAImpl) newCertificate(domains []string, ips []net.IP, key crypto.Publ
...
+	maxNotAfter := time.Date(9999, 12, 31, 0, 0, 0, 0, time.UTC)
+	if certNotAfter.After(maxNotAfter) {
+		certNotAfter = maxNotAfter
+	}
```
aarongable, line 287: Is this check necessary? Go's `time.Time` can store values much higher than this, and timestamps in certificates are just encoded as strings and so do not have true max values.

ca/ca.go
```diff
@@ -340,10 +347,11 @@ func (ca *CAImpl) newCertificate(domains []string, ips []net.IP, key crypto.Publ
 	return newCert, nil
 }
 
-func New(log *log.Logger, db *db.MemoryStore, ocspResponderURL string, alternateRoots int, chainLength int) *CAImpl {
+func New(log *log.Logger, db *db.MemoryStore, ocspResponderURL string, alternateRoots int, chainLength int, certificateLifetime int) *CAImpl {
```
aarongable, line 350: similarly make `certValidityPeriod` a `uint` here.

ca/ca.go
```diff
@@ -362,6 +370,18 @@ func New(log *log.Logger, db *db.MemoryStore, ocspResponderURL string, alternate
...
+	if certificateLifetime != 0 {
+		if certificateLifetime > 0 {
+			ca.certLifetime = certificateLifetime
+			ca.log.Printf("Using user defined certificate lifetime of %d days", certificateLifetime)
+		} else {
+			ca.log.Printf("certificateLifetime (%d) is not a valid value (valid values: integer > 0), using default certificate lifetime of %d days", certificateLifetime, ca.certLifetime)
+		}
+	} else {
+		ca.log.Printf("Using default certificate lifetime of %d days", ca.certLifetime)
+	}
```
aarongable, line 383: Honestly I'd just simplify this to
```suggestion
	if certValidityPeriod != 0 {
		ca.certValidityPeriod = certValidityPeriod
	}
	ca.log.Printf("Using certificate validity period of %d seconds", ca.certValidityPeriod)
```
I don't think it's super important for the log message to indicate whether or not it's using the default value; it should just indicate what the value is.

cmd/pebble/main.go
```diff
@@ -28,6 +28,8 @@ type config struct {
 		ExternalAccountMACKeys         map[string]string
 		// Configure policies to deny certain domains
 		DomainBlocklist []string
+
+		CertificateLifetime int
```
aarongable, line 32: One last place to make it a `uint`.

aarongable (commented):
Of the three mechanisms you've proposed for making this configurable, I like this approach by far the best. Just a few small comments (nothing substantive, mostly naming nits and slight simplifications) and this'll be ready to go!

ca/ca.go
osirisinferi: Yeah, I tried using `uing` in #362 I believe and something broke in that PR.. So I didn't bother looking at the other PRs to be honest. Should probably work here.

ca/ca.go
osirisinferi: I'm afraid it is. If you go one second further than 9999-12-31 23:59:59 you'll overflow the "notAfter" field, which is a "GeneralizedTime":

> 4.1.2.5.2.  GeneralizedTime
> 
>    The generalized time type, GeneralizedTime, is a standard ASN.1 type
>    for variable precision representation of time.  Optionally, the
>    GeneralizedTime field can include a representation of the time
>    differential between local and Greenwich Mean Time.
> 
>    For the purposes of this profile, GeneralizedTime values MUST be
>    expressed in Greenwich Mean Time (Zulu) and MUST include seconds
>    (**i.e., times are YYYYMMDDHHMMSSZ**), even where the number of seconds
>    is zero.  GeneralizedTime values MUST NOT include fractional seconds.

The reason why I build this check in the first place is because otherwise Pebble would fail to issue the certificate and my certbot would just keep polling Pebble without any chance of success.

ca/ca.go
osirisinferi: Although I agree the granularity of certs is at the second level, I'm not sure what benefit it'll have to also for the cert validity period to have such granularity. But I also have no objection obviously.

ca/ca.go
osirisinferi: Had to use `time.Duration()` in combination with `time.Second()` otherwise Go would complain about `uint` not being type `time.Duration`...

ca/ca.go
```diff
@@ -280,7 +280,7 @@ func (ca *CAImpl) newCertificate(domains []string, ips []net.IP, key crypto.Publ
 		}
 	}
 
-	certNotAfter := time.Now().AddDate(0, 0, ca.certLifetime)
+	certNotAfter := time.Now().Add(time.Duration(ca.certValidityPeriod-1) * time.Second)
```
aarongable, line 283: Oh my goodness I didn't even notice that it does two separate calls to `time.Now()`! Obviously you don't need to fix this now but this should definitely be `certNotBefore.Add(...)`

aarongable (approved):

ca/ca.go
aarongable: To me, the two biggest benefits are
1) Then all the math (especially subtracting one second to compute the validity period correctly) is done in the same units; and
2) Boulder's config files measure this in seconds, too.

ca/ca.go
osirisinferi: Hm, that's a sharp eye, I haven't even looked at `certNotBefore` :stuck_out_tongue:

ca/ca.go
osirisinferi: By the way, I'm not sure if this is still necessary since we're using `time.Duration()` with `time.Add()`:

> A Duration represents the elapsed time between two instants as an int64 nanosecond count. The representation limits the largest representable duration to approximately 290 years. ([type Duration](https://pkg.go.dev/time#Duration))

osirisinferi:
Hm, not sure if you'd see my latest comment above, so I'll repeat it here:

`time.Duration()` uses an `int64` internally which will overflow the NotAfter date if the duration is set to 9223372038 or more (the floored max. of an `int64` plus 1 second (because we do -1) and an extra second to overflow it.

Which results in this:

```
        Validity
            Not Before: Aug 23 19:47:10 2021 GMT
            Not After : May 14 19:59:54 1729 GMT
```

aarongable (approved):

andygabby (approved):
Thanks!

PR #362, opened 2021-08-22
Add management server handler to change certificate lifetime

osirisinferi:
Partly related to #353 and #361

This PR adds a new endpoint to the management interface: `/change-cert-lifetime/`

It takes an integer as value (i.e. `/change-cert-lifetime/123`) which will set the certificate lifetime to the value amount of days.

The output will show a JSON structure with "Old" and "New" containing the previous and new certificate lifetimes if successful. It will return a HTTP 400 if the requested value wasn't an integer > 0 and it will return a HTTP 500 if something else went wrong.


aarongable:
Superceded by #361

PR #372, opened 2022-02-17
ci: generate binaries for macOS arm64

### LINKED ISSUE #343: Include macOS binary in releases
Currently, there are only release binaries for Windows and Linux. It would help if there were binaries for macOS also. 

(e.g., in my package, [node-pebble](https://github.com/small-tech/node-pebble), I download the correct release at install time to enable testing with Pebble via Node.js but it only supports Linux and Windows right now as there isn’t a means to get an official binary for macOS).

Thanks :)

alexzorin:
This would help with getting Certbot integration tests running natively on macOS.

Fixes #343.﻿

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,6 +43,7 @@ install:
   - go mod vendor
   - git diff --exit-code vendor/
   - go install -v -race -mod=vendor ./...
+  - GOOS=darwin GOARCH=arm64 go install -v -mod=vendor ./...
 
 before_script:
   - GORACE="halt_on_error=1" PEBBLE_WFE_NONCEREJECT=0 pebble &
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -43,6 +43,7 @@ install:
   - go mod vendor
   - git diff --exit-code vendor/
   - go install -v -race -mod=vendor ./...
+  - GOOS=darwin GOARCH=arm64 go install -v -mod=vendor ./...
```
aarongable, line 46: Any reason not to pass `-race` here, like we do in the amd64 build on the line above?

alexzorin: Yes, it would seem that cgo is required for `-race`, which is not available with cross-compilation.

In view of this pipeline being eventually migrated to GitHub Actions, it seemed like cross-compiling was preferable to running a separate macOS-native build.

aarongable (approved):

PR #374, opened 2022-03-31
remove superfluous calls to net/http.Response.WriteHeader

alexzorin:
WebFrontEndImpl.sendError is already responsible for sending the
HTTP status code related to these ACME problems.

---

While playing with Certbot's handling of ACME errors in https://github.com/certbot/certbot/pull/9255, I noticed some output like:

>2022/03/31 12:55:53 http: superfluous response.WriteHeader call from github.com/letsencrypt/pebble/wfe.(*WebFrontEndImpl).sendError (wfe.go:346)

Each of the removed calls match the HTTP status code that `sendError` would have sent anyway.

### DIFF
--- a/wfe/wfe.go
+++ b/wfe/wfe.go
@@ -1782,7 +1782,6 @@ func (wfe *WebFrontEndImpl) Order(
 	// authenticated account owns the order being requested
 	if account != nil {
 		if orderAccountID != account.ID {
-			response.WriteHeader(http.StatusForbidden)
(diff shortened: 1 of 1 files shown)


aarongable (approved):
Heh, I had to go double-check that `acme.UnauthorizedProblem` results in 403 (Forbidden), not 401 (Unauthorized), but it does! So this means that a couple codepaths (lines 2137, 2290, and 2411 of the original) might start sending 403s instead of 401s, but that seems like the correct thing to me anyway. LGTM.

Issue #83, opened 2017-12-28
Retry-After headers for polling orders and authorizations


sophie-h:
Eventually, the server knows when it will retry a challenge validation etc. and therefore when it makes sense for the client to recheck the status.

PR #380, opened 2022-05-03
Add Retry-After header when responding to Order and Authorization object

moratori:
Address the issue #83
Fixed to add Retry-After header when responding to requests for Order and Authorization objects.

Although it is a static, The value of Retry-After is defined in the configuration file.
I would appreciate it if you could review this PR.

### DIFF
--- a/test/config/pebble-config-external-account-bindings.json
+++ b/test/config/pebble-config-external-account-bindings.json
@@ -7,6 +7,10 @@
     "httpPort": 5002,
     "tlsPort": 5001,
     "ocspResponderURL": "",
+    "retryAfter": {
+        "authz": 3,
+        "order": 5
+    }
     "externalAccountBindingRequired": true,
     "externalAccountMACKeys": {
       "kid-1": "zWNDZM6eQGHWpSRTPal5eIUYFTu7EajVIoguysqZ9wG44nMEtx3MUAsUDkMTQ12W",
--- a/wfe/wfe.go
+++ b/wfe/wfe.go
@@ -2094,7 +2094,10 @@ func (wfe *WebFrontEndImpl) Authz(
 			// Check for the existence of a challenge which state is processing, and add the
 			// Retry-After header if there is one.
 			for _, c := range authz.Challenges {
-				if c.Status == acme.StatusProcessing {
+				c.RLock()
+				challengeStatus := c.Status
+				c.RUnlock()
+				if challengeStatus == acme.StatusProcessing {
 					addRetryAfterHeader(response, wfe.retryAfterAuthz)
 					break
 				}
--- a/wfe/wfe.go
+++ b/wfe/wfe.go
@@ -155,6 +155,8 @@ type WebFrontEndImpl struct {
 	ca                *ca.CAImpl
 	strict            bool
 	requireEAB        bool
+	retryAfterAuthz   int
+	retryAfterOrder   int
 }
 
 const ToSURL = "data:text/plain,Do%20what%20thou%20wilt"
@@ -164,7 +166,7 @@ func New(
 	db *db.MemoryStore,
 	va *va.VAImpl,
 	ca *ca.CAImpl,
-	strict, requireEAB bool) WebFrontEndImpl {
+	strict, requireEAB bool, retryAfterAuthz int, retryAfterOrder int) WebFrontEndImpl {
 	// Seed rand from the current time so test environments don't always have
 	// the same nonce rejection and sleep time patterns.
 	rand.Seed(time.Now().UnixNano())
@@ -225,6 +227,8 @@ func New(
 		ca:                ca,
 		strict:            strict,
 		requireEAB:        requireEAB,
+		retryAfterAuthz:   retryAfterAuthz,
+		retryAfterOrder:   retryAfterOrder,
 	}
 }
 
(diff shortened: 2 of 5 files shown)


wfe/wfe.go
```diff
@@ -2079,6 +2085,8 @@ func (wfe *WebFrontEndImpl) Authz(
 				response)
 			return
 		}
+
+		response.Header().Add("Retry-After", strconv.Itoa(wfe.retryAfterAuthz))
```
aarongable, line 2089: Similar comment here as above: the Retry-After header is only really relevant when the Authz is in the "pending" state; consider checking for `authz.Status == acme.StatusPending` and only adding the header in that case.

alexzorin: @aarongable since authorizations begin life in the `pending` state, should it also check that at least one of the challenges is undergoing validation?

Otherwise Pebble will send a `Retry-After` on a brand new authorization, where the client hasn't responded to any of the challenges, and there is nothing to wait for.

This will probably require implementing the `processing` state on challenges, or an internal flag on the challenge object, though.

wfe/wfe.go
```diff
@@ -1788,6 +1792,8 @@ func (wfe *WebFrontEndImpl) Order(
 		}
 	}
 
+	response.Header().Add("Retry-After", strconv.Itoa(wfe.retryAfterOrder))
```
alexzorin, line 1795: Since `Retry-After` [may be represented by a seconds duration or by an absolute date](https://datatracker.ietf.org/doc/html/rfc7231#section-7.1.3), perhaps it would be good to randomize the representation here.

alexzorin, line 1795: `wfe.retryAfterOrder` may be zero and no `Retry-After` header should be sent.

wfe/wfe.go
alexzorin, line 2089: `wfe.retryAfterAuthz` may be zero and no `Retry-After` header should be sent.

test/config/pebble-config-external-account-bindings.json
```diff
@@ -7,6 +7,10 @@
     "httpPort": 5002,
     "tlsPort": 5001,
     "ocspResponderURL": "",
+    "retryAfter": {
+        "authz": 3,
+        "order": 5
+    }
```
alexzorin, line 13: I am a little worried about how much these defaults will slow down CI pipelines, especially for `authz`. Don't have a great answer about what to do, though.

alexzorin (changes requested):

moratori:
Thank you very much for your feedback.
Maybe I think that I have fixed the above issues.
Could you please review it once again.

aarongable (commented):
Just two small comments, otherwise looks good to me!

moratori:
I finished editing it!
If you have any other comments, please let me know.

aarongable (approved):

wfe/wfe.go
```diff
@@ -2079,6 +2089,17 @@ func (wfe *WebFrontEndImpl) Authz(
 				response)
 			return
 		}
+
+		if authz.Status == acme.StatusPending {
+			// Check for the existence of a challenge which state is processing, and add the
+			// Retry-After header if there is one.
+			for _, c := range authz.Challenges {
+				if c.Status == acme.StatusProcessing {
```
alexzorin, line 2097: As mentioned in [this comment](https://github.com/letsencrypt/pebble/pull/380#discussion_r865422209), the `processing` state for challenges is not currently implemented in Pebble. 

Consequently, the `Retry-After` header is never sent for challenges.

I think we'd need to set the state to `processing` right as the challenge is queued up for validation in `wfe.updateChallenge`. Do we want to do this in a separate PR or here?

alexzorin, line 2097: I think it's likely that we need to do a `c.RLock()` and `c.RUnlock()` around the access to `c.Status`, or it may be a race.

alexzorin (changes requested):

wfe/wfe.go
moratori: @alexzorin 
There was an error in my understanding. My apologies.
I now understand that `processing` status is not currently used for `challenge` in pebble.
Regarding the unused `processing`, I felt that perhaps it would be appropriate to proceed with implementation in another PR.

moratori: @alexzorin 
Thank you for your comment. I will attempt to fix this problem.

moratori: This problem was fixed by commit [6f709f9](https://github.com/letsencrypt/pebble/pull/380/commits/6f709f914d48e7bc067683ddcc851c3701c15744)

moratori:
@alexzorin 
@aarongable 
I'm sorry to bother you again, I would be grateful if you could review it.
Fixed to acquire RLock before reading status of challenge.

alexzorin:
Sorry for disappearing. I'd still like to fix #382 soon, to enable this PR.

moratori:
Thank you for implementing it!

moratori:
@alexzorin 
The PR #382 was merged successfully.
I would appreciate it if you review this PR once again.

alexzorin (approved):
This LGTM and appears to work well, thank you!

The PR will still require another review from @aarongable to merge.

aarongable (approved):

PR #382, opened 2022-05-08
add 'processing' state to challenges

alexzorin:
For https://github.com/letsencrypt/pebble/pull/380#discussion_r867289222. 

Keeping track of when a challenge is in the `processing` state means we will be able to then include a `Retry-After` response header for authorizations, as described in:

>    To check on the status of an authorization, the client sends a POST-
   as-GET request to the authorization URL, and the server responds with
   the current authorization object.  In responding to poll requests
   while the validation is still in progress, the server MUST return a
   200 (OK) response and MAY include a Retry-After header field to
   suggest a polling interval to the client.

One extra side effect here is that clients will no longer be able to cause Pebble to queue up a `pending` challenge multiple times for processing by a VA. This seems like a good thing anyway.

### DIFF
--- a/va/va.go
+++ b/va/va.go
@@ -157,10 +157,6 @@ func New(
 }
 
 func (va VAImpl) ValidateChallenge(ident acme.Identifier, chal *core.Challenge, acct *core.Account) {
-	chal.Lock()
-	chal.Status = acme.StatusProcessing
-	chal.Unlock()
-
 	task := &vaTask{
 		Identifier: ident,
 		Challenge:  chal,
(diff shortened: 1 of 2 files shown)


alexzorin:
On second thought, this is still racey. I guess we have to set the status earlier, when we hold the lock to check that the current status is `pending`.

moratori (approved):

aarongable (approved):
LGTM, but I'd like a second set of eyes on this, since I'm actually not very familiar with Pebble's in-memory datastore and the locking necessary to work with it safely. @jsha?

jsha (approved):
Thanks very much!

PR #386, opened 2022-06-07
Validate CSR signatures and check signature type

### LINKED ISSUE #384: Pebble should fail CSR requests signed using SHA-1
Pebble should fail to finalise a request if it has been signed using the SHA-1 signature algorithm to match:

  - The current behaviour of the Let’s Encrypt staging server (as of at least the time of this writing)
  - [The future behaviour of the Let’s Encrypt production server](https://community.letsencrypt.org/t/rejecting-sha-1-csrs-and-validation-using-tls-1-0-1-1-urls/175144) from September 15, 2022.

mcpherrinm:
With the deprecation of SHA1 signatures on CSRs, we should ensure Pebble rejects them as well.  As well, ensure the signature itself is valid.  Unlike Boulder, which currently allows SHA1 to be configurably accepted, Pebble will always reject SHA1.

There's no tests for the WFE code in this repository, so I didn't add a unit test ensuring SHA1 is rejected.  However, I tested this by running an ACME client against Pebble that was patched to use SHA1 signatures and verifying they were successfully rejected, as well as ensuring SHA256 was still accepted

Fixes #384

### DIFF
--- a/wfe/wfe.go
+++ b/wfe/wfe.go
@@ -1637,7 +1637,7 @@ func (wfe *WebFrontEndImpl) NewOrder(
 	err := json.Unmarshal(postData.body, &newOrder)
 	if err != nil {
 		wfe.sendError(
-			acme.MalformedProblem("Error unmarshaling body JSON: "+err.Error()), response)
+			acme.MalformedProblem(fmt.Sprintf("Error unmarshalling body JSON: %s", err.Error())), response)
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


osirisinferi:
Boulder rejects SHA1 hashes in its `ca` "module", not `wfe`. Could that matter regarding ACME client behavior differences between Pebble and Boulder?

mcpherrinm:
I don't think where exactly the messages are rejected matters:  Pebble is a much smaller codebase with less modules than boulder so it won't ever align the same exactly.  Pebble only handles the CSRs within the WFE (web frontend) code here so it would require restructuring more code to have the check occur elsewhere.

wfe/wfe.go
```diff
@@ -1882,6 +1891,18 @@ func (wfe *WebFrontEndImpl) FinalizeOrder(
 		return
 	}
 
+	// Ensure the signature on the CSR is good
+	if err := parsedCSR.CheckSignature(); err != nil {
+		wfe.sendError(acme.BadCSRProblem("Bad signature on CSR: "+err.Error()), response)
```
aarongable, line 1896: Although I see it is done using concatenation on lines 1883 and 1890 above, we prefer string interpolation (as on line 1876, 1864, and 1857) in general:
```suggestion
		wfe.sendError(acme.BadCSRProblem(fmt.Sprintf("Bad signature on CSR: %s", err.Error())), response)
```

wfe/wfe.go
```diff
@@ -1882,6 +1891,18 @@ func (wfe *WebFrontEndImpl) FinalizeOrder(
...
+		return
+	}
+
+	// Ensure a supported CSR algorithm is used
+	if good := goodSignatureAlgorithms[parsedCSR.SignatureAlgorithm]; !good {
+		wfe.sendError(acme.BadCSRProblem("CSR signature algorithm not supported: "+parsedCSR.SignatureAlgorithm.String()), response)
```
aarongable, line 1902: Same here, and we usually reserve "message: other message" for cases where the bit after the colon is an embedded error:
```suggestion
		wfe.sendError(acme.BadCSRProblem(fmt.Sprintf("CSR signature algorithm %s is not supported", parsedCSR.SignatureAlgorithm.String())), response)
```

wfe/wfe.go
```diff
@@ -1882,6 +1891,18 @@ func (wfe *WebFrontEndImpl) FinalizeOrder(
 		return
 	}
 
+	// Ensure the signature on the CSR is good
+	if err := parsedCSR.CheckSignature(); err != nil {
```
aarongable, line 1895: We prefer to avoid the `if err := ...; err != nil` syntax: https://github.com/letsencrypt/boulder/blob/main/docs/CONTRIBUTING.md#error-handling

```suggestion
	err = parsedCSR.CheckSignature()
	if err != nil {
```

wfe/wfe.go
```diff
@@ -116,6 +116,15 @@ const (
 	defaultOrdersPerPage = 3
 )
 
+var goodSignatureAlgorithms = map[x509.SignatureAlgorithm]bool{
```
aarongable, line 119: If you care about saving a byte here or there, this can instead be a `map[x509.SignatureAlgorithm]struct{}`, which Go will optimize to having zero-byte values instead of one-byte values. But that's likely premature optimization here.

aarongable (commented):
Overall LGTM, just a few formatting comments

wfe/wfe.go
mcpherrinm: If I was writing this from scratch I'd probably do that, but this list was taken verbatim from boulder.  Consistency seemed more important than a byte

mcpherrinm:
Design question:  Should either of these checks be gated on Pebble's "strict mode"?

aarongable (approved):
LGTM. I wouldn't gate rejecting SHA1 behind "strict mode": half the reason we want to remove it is that Go itself is going to remove support, so Pebble needs to deprecate it just as much as Boulder does. Validating the signature could be behind the strict-mode flag but I don't feel strongly either way.

mcpherrinm:
I don't think it's likely to be helpful to anyone to allow invalid CSR signatures, especially since Let's Encrypt and presumably other CAs always require valid CSR signatures.  So I think we can leave it as-is and merge this.

jsha (approved):

PR #408, opened 2023-05-30
Add clock mock plumbing throughout the project

pzduniak:
Hey, it appears that the repo included github.com/jmhodges/clock as a dependency at some point, but it was removed (#231).

As I was trying to integrate pebble into our integration tests, I ran into the issue of the lack of the clock mocks, which was a dealbreaker. I reintroduced the dependency here, unfortunately it breaks the external API and it's _probably_ not done the way you, the maintainers, would approach it. I'm open to spending some time on fine tuning this PR to better match the rest of the codebase stylistically. This was the bare minimum of changes I had to make to make it all work.


cmd/pebble/main.go
```diff
@@ -8,6 +8,7 @@ import (
 	"os"
 	"strconv"
 
+	"github.com/jmhodges/clock"
```
aarongable, line 11:
```suggestion
	"github.com/jmhodges/clock"
	
```

ca/ca.go
```diff
@@ -34,6 +36,7 @@ type CAImpl struct {
 	log              *log.Logger
 	db               *db.MemoryStore
 	ocspResponderURL string
+	clockSource      clock.Clock
```
aarongable, line 39: Here and throughout, let's rename `clockSource` to `clk`, to match [Boulder](https://github.com/letsencrypt/boulder/blob/89b07f4543e067d21e933ec2603e87f399669912/ca/ca.go#L96).

ca/ca.go
```diff
@@ -342,17 +345,21 @@ func (ca *CAImpl) newCertificate(domains []string, ips []net.IP, key crypto.Publ
...
 		certValidityPeriod: defaultValidityPeriod,
 	}
 
 	if ocspResponderURL != "" {
 		ca.ocspResponderURL = ocspResponderURL
 		ca.log.Printf("Setting OCSP responder URL for issued certificates to %q", ca.ocspResponderURL)
 	}
+	if clockSource == nil {
+		ca.clockSource = clock.New()
+	}
```
aarongable, line 362: Two comments:
1) Here and throughout, honestly I think it's fine to leave this check/default off. The clk can just be a required argument and pebble can just panic if it isn't provided. Just like it would if the db argument was nil.
2) But if you really do want to keep it optional, do this check/default before the creation of the CAImpl on line 349.

cmd/pebble/main.go
```diff
@@ -93,9 +94,10 @@ func main() {
 		chainLength = int(val)
 	}
 
-	db := db.NewMemoryStore()
-	ca := ca.New(logger, db, c.Pebble.OCSPResponderURL, alternateRoots, chainLength, c.Pebble.CertificateValidityPeriod)
-	va := va.New(logger, c.Pebble.HTTPPort, c.Pebble.TLSPort, *strictMode, *resolverAddress)
+	clockSource := clock.New()
```
aarongable, line 97: Based on your PR description (you mention "breaking the external API") it sounds like you're using Pebble as a library, rather than as a standalone binary? In that case, your changes here make sense.

But if we're going to go to the trouble of giving Pebble full fake clock support, we should give it to everyone, including people who use Pebble as a binary (which I think is _most_ users of Pebble).

To that end, let's replace this line with something a bit more like [this](https://github.com/letsencrypt/boulder/blob/main/cmd/clock_integration.go), so that the fake clock can be initialized from an environment variable if someone wants to, or defaults to the system clock otherwise.

aarongable (changes requested):
Thanks for the PR, this generally looks good to me! Just a few small comments, but nothing big and no concerns from me. Thanks!

pzduniak:
Thanks for the review, I'll address your comments when I have some free time.

pzduniak:
I've done this from scratch with a slightly different approach in #552. Thanks for the review @aarongable!

PR #409, opened 2023-05-30
Change Go version in go.mod to 1.18, fixing CI and matching the Docker image version

pzduniak:
CI was broken in my branch. This fixes it. Also ran `go mod tidy && go mod vendor`, though it probably wasn't needed. Unclear why the diff was generated.


pzduniak:
looks like appveyor needs a Go update, haven't ever used it so no clue how it works

ldez:
Since #434 this PR can be closed.

PR #420, opened 2023-11-09
Remove CommonName from issued certificates

mcpherrinm:
The Baseline requirements now says Common Name is "Not Recommended" for
subscriber server certificates, so this follows that recommendation.


jsha (approved):
Thanks! (and sorry for the delay)

PR #428, opened 2024-01-18
Add eab key example that enforces base64url

j0ru:
The acme rfc states that clients should interpret supplied eab keys as base64url. To encourage that behavior, this commit adds a key to the example configuration that is only decodable by base64url.


mcpherrinm:
pulled into #459

Issue #425, opened 2024-01-11
Implement the "dns-account-01" Challenge in Pebble


sheurich:
### Objective

Implement the "dns-account-01" challenge in Pebble, setting the groundwork for subsequent Boulder alignment (referencing [boulder#7240](https://github.com/letsencrypt/boulder/issues/7240)).

### Rationale

Establishing this feature in Pebble first is crucial for a consistent and forward-compatible testing framework.

sheurich:
In addition to the ongoing work from @orangepizza, I have prepared two changes to support Pebble integration tests using `dns-account-01`. 
- https://github.com/certbot/certbot/pull/9887
- https://github.com/letsencrypt/pebble/pull/432

In conjunction with my proposed change to https://github.com/letsencrypt/pebble/pull/430 (see https://github.com/letsencrypt/pebble/pull/430#issuecomment-1928497141), the preceding changes to `certbot/acme` and `pebble/test/chisel2.py` enable a successful integration test between pebble, challtestsrv and chisel2. 

Example run:

```shell
REQUESTS_CA_BUNDLE=pebble/test/certs/pebble.minica.pem \
  python3 -c 'import pebble.test.chisel2; pebble.test.chisel2.auth_and_issue(["foo.com"],"dns-account-01")'

…
DEBUG:urllib3.connectionpool:https://localhost:14000 "POST /authZ/AFDXL5tMBBi6GTDqetoTuPArYOTXi7_GGL0CQWCh6Qs HTTP/1.1" 200 440
DEBUG:acme.client:Received response:
HTTP 200
Cache-Control: public, max-age=0, no-cache
Content-Type: application/json; charset=utf-8
Link: <https://localhost:14000/dir>;rel="index"
Replay-Nonce: BAMHTKcD2aNAyopsB-9zZg
Date: Tue, 06 Feb 2024 00:29:17 GMT
Content-Length: 440

{
   "status": "valid",
   "identifier": {
      "type": "dns",
      "value": "foo.com"
   },
   "challenges": [
      {
         "type": "dns-account-01",
         "url": "https://localhost:14000/chalZ/lV0CIU0zxaGkt5Ib7og8d1vkQo1A2EHzMREKbgdsZOI",
         "token": "01Wsj3RQfOinMJ7-lbKRBeElrLBI8OaUv99Jioem0_w",
         "status": "valid",
         "validated": "2024-02-06T00:29:14Z"
      }
   ],
[...]

PR #430, opened 2024-01-31
Implement the "dns-account-01" Challenge 

orangepizza:
Solve https://github.com/letsencrypt/pebble/issues/425

implement https://datatracker.ietf.org/doc/draft-ietf-acme-dns-account-challenge/
as only thing different is from dns-01 is challenge subdomain craft, which needs what hostname user expects:
this PR handed this by sending relative accountURL path from WFE to VA. 
need client impl but as hash craft itself is tested to make same result from RFC draft input, and using same craft function for accountURL craft that used with other url endpoints like new-reg


sheurich:
👋🏽 Hi @orangepizza and thanks for the contributing this feature! The intent with https://github.com/letsencrypt/pebble/issues/425 is to implement the challenge as described in https://github.com/aaomidi/draft-ietf-acme-dns-account-challenge/pull/25. Adding https://github.com/fastly/pebble/commit/2afa9f4697d68973edde30cbea3b33d09c95ef64 will make the validation label match the updated RFC draft.

```go
const dnsacct01Prefix = "_acme-challenge"
…
challengeSubdomain := fmt.Sprintf("_%s.%s.%s", urlhash, dnsacct01Prefix, task.Identifier.Value)
```

orangepizza:
updated for new scoped challenge: but looks too late :(

orangepizza:
superseded by https://github.com/letsencrypt/pebble/pull/435

Issue #419, opened 2023-11-05
Golang, apk and zlib versions are outdated


NoamGaloz:
Hello,

The versions of Golang, apk-tools, zlib, and some more are outdated.
Is there a possibility to make a contribution to upgrading those versions and you'll run a new release?

jsha:
Go for it!

NoamGaloz:
please create a new release @jsha

PR #434, opened 2024-02-17
Upgrade Go and dependencies; CI and Docker fixes

sheurich:
Keeping current:
- Upgrade Go to `v1.21`
- Upgrade Go dependencies
  - Upgrade `github.com/miekg/dns` to `v1.1.58`
  - Upgrade indirect dependencies
- Fixes for appveyor
  - Upgrade to supported image `VS 2019`
  - Upgrade to supported Python `v3.8`
- Fixes for golangci-lint
  - Use supported installation method
- Upgrade Docker base images to be supported and match Go version
  - `Alpine v3.16` and `Nanoserver v1806`

Addresses:
- https://github.com/letsencrypt/pebble/issues/419
- https://github.com/letsencrypt/pebble/pull/409


jsha (approved):
This is great! Thanks so much for the updates.

sheurich:
On this topic:
1. Is Windows Container support needed going forward?
2. Can we use GitHub Actions in lieu of Appveyor and Travis?
3. Is the purpose of the `docker-compose.yml` for testing Pebble or testing other clients against Pebble? If both, should the use-cases be separated out?

jsha:
> Is Windows Container support needed going forward?

I think so. @wgreenberg can you confirm? If it's not needed for Certbot I'm inclined to drop it.

> Can we use GitHub Actions in lieu of Appveyor and Travis?

Yes, very much so; I've been meaning to make this change for a while, and publish containers via ghcr. If you have a chance to do that it would be wonderful.

> Is the purpose of the docker-compose.yml for testing Pebble or testing other clients against Pebble? If both, should the use-cases be separated out?

Looking back at when it was added in #181, "The stand-alone binary with an HTTP management interface can come live in the Pebble repo where more folks can use it without pulling in all of Boulder. I've heard from a few ACME client developers that this would be useful to them." So I interpret that to be "it's useful for testing other clients against Pebble." AFAICT the docker-compose.yml is not used for testing Pebble.

sheurich:
https://github.com/letsencrypt/pebble/pull/439 has follow-on work related to
> Fixes for golangci-lint

and also builds towards GitHub Actions wip in https://github.com/fastly/pebble/tree/4d797ae3ecf4d2a1825c9b19942002326890a2ae

A GHA PoC is available at https://github.com/fastly/pebble/actions/runs/8034580298

PR #436, opened 2024-02-22
ca: Pass OCSP Must-Staple from CSR into generated certificate

### LINKED ISSUE #433: Support must-staple extension
It'd be nice to move certbot off of using boulder for its integration tests (https://github.com/certbot/certbot/issues/9898), and we largely have, but the one remaining boulder-only test that remains is for issuing an OCSP Must-Staple enabled certificate. We'd love if pebble supported this so we can simplify our test environment! Happy to open a PR if you all think this would be a welcome addition.

wgreenberg:
This will allow for extensions such as OCSP stapling. Happy to add a test if desired, also I wasn't sure if it's expected to filter for specific extensions or to pass them wholesale onto the cert?

Fixes #433

### DIFF
--- a/ca/ca.go
+++ b/ca/ca.go
@@ -1,6 +1,7 @@
 package ca
 
 import (
+	"bytes"
 	"crypto"
 	"crypto/rand"
 	"crypto/rsa"
@@ -374,6 +375,10 @@ func New(log *log.Logger, db *db.MemoryStore, ocspResponderURL string, alternate
 	return ca
 }
 
+func isOCSPMustStapleExtension(ext pkix.Extension) bool {
+	return ext.Id.Equal(asn1.ObjectIdentifier{1, 3, 6, 1, 5, 5, 7, 1, 24}) && bytes.Equal(ext.Value, []byte{0x30, 0x03, 0x02, 0x01, 0x05})
+}
+
 func (ca *CAImpl) CompleteOrder(order *core.Order) {
 	// Lock the order for reading
 	order.RLock()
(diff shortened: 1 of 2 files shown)


jsha:
I would prefer to look for just the Must Staple extension and set that, which is what Boulder does. It's quite dangerous to pass through arbitrary extensions from the CSR, because some extensions change the semantics of certificates in dramatic ways. Of course Pebble is a testing CA but we should still avoid creating dangerous code for someone to possibly emulate. :D

ca/ca.go
```diff
@@ -396,9 +402,17 @@ func (ca *CAImpl) CompleteOrder(order *core.Order) {
 		authz.RUnlock()
 	}
 
+	// Build a list of approved extensions to include in the certificate
+	var extensions []pkix.Extension
+	for _, ext := range order.ParsedCSR.Extensions {
+		if isOCSPMustStapleExtension(ext) {
+			extensions = append(extensions, ext)
```
aarongable, line 409: Directly appending the extension from the CSR is still _slightly_ dangerous, as we are directly carrying over the critical bit. Boulder has a [pre-built OCSP must-staple extension](https://github.com/letsencrypt/boulder/blob/34b705f29348930ab6f6ac3c8b70f3dcdd061e24/issuance/cert.go#L202), and it uses that both to [detect must-staple in the CSR](https://github.com/letsencrypt/boulder/blob/34b705f29348930ab6f6ac3c8b70f3dcdd061e24/issuance/cert.go#L348) and [include must-staple in the resulting cert](https://github.com/letsencrypt/boulder/blob/34b705f29348930ab6f6ac3c8b70f3dcdd061e24/issuance/cert.go#L300). That way Boulder can guarantee that the extension is never critical.

What you have here is probably fine for Pebble, but if you want to be even safer that's a pattern you could follow.

ca/ca_test.go
```diff
@@ -0,0 +1,60 @@
...
+			Expires:     time.Now().AddDate(0, 0, 1).UTC().Format(time.RFC3339),
+			Identifiers: uniquenames,
+			NotBefore:   time.Now().UTC().Format(time.RFC3339),
+			NotAfter:    time.Now().AddDate(30, 0, 0).UTC().Format(time.RFC3339),
+		},
+		ExpiresDate: time.Now().AddDate(0, 0, 1).UTC(),
+	}
+
+	ca.CompleteOrder(order)
+	log.Printf("cert: %+v", order.CertificateObject.Cert)
```
aarongable, line 59: Seems like this log line should be replaced by parsing the resulting cert and inspecting its extensions to confirm the inclusion of must-staple.

ca/ca.go
wgreenberg: ah nice, I'll do that! I wasn't sure how risky the critical value was, but that makes sense.

ca/ca_test.go
aarongable: I think it's great to have a test for this!

wgreenberg:
Added some unit tests and modeled this after Boulder's approach. Ready for review again!

mcpherrinm:
Can you rebase or merge in main please?  We've merged some fixes in the last few day for pebble's CI

wgreenberg:
Should be all set

mcpherrinm:
thanks.  A few things from the linter, otherwise lgtm

mcpherrinm (approved):

PR #379, opened 2022-04-19
Add GitHub Actions support

jsha:

PR #439, opened 2024-02-23
GolangCI-Lint Fixes

sheurich:
This PR addresses https://github.com/letsencrypt/pebble/issues/438 by fixing all of the reported issues in Pebble code.

When run with the default lint configuration, golangci-lint reports:
```plain
wfe/wfe.go:2365:2: S1017: should replace this `if` statement with an unconditional `strings.TrimPrefix` (gosimple)
        if strings.HasPrefix(ident.Value, "*.") {
        ^
wfe/wfe.go:1443:32: S1039: unnecessary use of fmt.Sprintf (gosimple)
                return acme.MalformedProblem(fmt.Sprintf(
                                             ^
wfe/wfe.go:1405:2: SA5011: possible nil pointer dereference (staticcheck)
        order.RLock()
        ^
wfe/wfe.go:1409:5: SA5011(related information): this check suggests that the pointer can be nil (staticcheck)
        if order == nil {
           ^
wfe/wfe.go:1406:8: SA5011: possible nil pointer dereference (staticcheck)
        defer order.RUnlock()
              ^
wfe/wfe.go:181:2: SA1019: rand.Seed has been deprecated since Go 1.20 and an alternative has been available since Go 1.0: As of Go 1.20 there is no reason to call Seed with a random value. Programs that call Seed with a known value to get a specific sequence of results should use New(NewSource(seed)) to obtain a local random generator. (staticcheck)
        rand.Seed(time.Now().UnixNano())
        ^
va/va.go:321:10: S1039: unnecessary use of fmt.Sprintf (gosimple)
                msg := fmt.Sprintf("No TXT records found for DNS challenge")
                       ^
va/va.go:338:9: S1039: unnecessary use of fmt.Sprintf (gosimple)
        msg := fmt.Sprintf("Correct value not found for DNS challenge")
               ^

wfe/wfe.go:178: File is not `gofumpt`-ed (gofumpt)
        strict, requireEAB bool, retryAfterAuthz int, retryAfterOrder int) WebFrontEndImpl {
```

These errors are addressed here.

Other fixes include:
- A fix for a WFE panic: [Use built-in UTC for TZ to eliminate panics](https://github.com/letsencrypt/pebble/pull/439/commits/2ee3d6bde0450bb9edc7790f0d04d9fc3bae8577)
- Fixes for Appveyor and Travis related to Python
- Update golanci-lint to current in Travis

PR #442, opened 2024-02-24
Use GitHub Actions

### LINKED ISSUE #356: Switch from Travis to GitHub Actions
We've stopped using Travis for Boulder; working with two different systems between these two repos is needlessly confusing. Let's switch Pebble to use GHA too.

ldez:
First, I know there is a current open [PR](https://github.com/letsencrypt/pebble/pull/379) on the topic, but there has been no move for a long time, and it feels more like a WIP than a mergeable PR.

I just added the basics:
- build (cross-OS)
- linting
- test (Linux)

Please consider it as a first step (it's better than no CI :smile_cat:).

I didn't add a workflow on Windows because the current Appveyor setup file is not Linux user-friendly.
But I can spend time on it later and create a specific PR.

I don't add the release part, I want to discuss it because it's not a simple step.

You can take a look at the CI of my fork if needed: https://github.com/ldez/pebble/actions

Fixes #356
Closes #379

Note: the PR #441 should be merged first.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -1,0 +1,54 @@
+name: Checks
+
+on:
+  push:
+    branches:
+      - main
+      - ldez-ci
+  pull_request:
+
+permissions:
+  contents: read
+  pull-requests: read
+
+jobs:
+
+  checks:
+    name: Check Process
+    runs-on: ubuntu-latest
+    env:
+      GO_VERSION: oldstable
+      GOLANGCI_LINT_VERSION: v1.56.2
+      CGO_ENABLED: 0
+
+    steps:
+
+      - name: Check out code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: ${{ env.GO_VERSION }}
+
+      - name: Check and get dependencies
+        run: |
+          go mod tidy
+          git diff --exit-code go.mod
+          git diff --exit-code go.sum
+
+      - name: vendoring
+        run: go mod vendor
+
+      - name: vendoring diff
+        run: git diff --exit-code vendor/
+
+      # https://golangci-lint.run/usage/install#other-ci
+      - name: Install golangci-lint ${{ env.GOLANGCI_LINT_VERSION }}
+        run: |
+          curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin ${GOLANGCI_LINT_VERSION}
+          golangci-lint --version
+
+
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -48,6 +48,6 @@ jobs:
       - name: Install golangci-lint ${{ env.GOLANGCI_LINT_VERSION }}
         run: |
           curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin ${GOLANGCI_LINT_VERSION}
-          golangci-lint --version
+          golangci-lint run
 
 
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - main
-      - ldez-ci
   pull_request:
 
 permissions:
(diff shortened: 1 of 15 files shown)


.github/workflows/checks.yml
```diff
@@ -0,0 +1,53 @@
...
+
+      - name: vendoring diff
+        run: git diff --exit-code vendor/
+
+      # https://golangci-lint.run/usage/install#other-ci
+      - name: Install golangci-lint ${{ env.GOLANGCI_LINT_VERSION }}
+        run: |
+          curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin ${GOLANGCI_LINT_VERSION}
+          golangci-lint --version
+
```
aarongable, line 52: Looks like we're missing a step here to actually run golangci-lint

.github/workflows/go-cross.yml
```diff
@@ -0,0 +1,38 @@
...
+  pull_request:
+
+permissions:
+  contents: read
+  pull-requests: read
+
+jobs:
+
+  cross:
+    name: Go
```
aarongable, line 15:
```suggestion
    name: Build
```

.github/workflows/tests.yml
```diff
@@ -0,0 +1,92 @@
...
+        with:
+          repository: letsencrypt/boulder
+          path: boulder
+          ref: release-2024-02-20
+
+      - name: install load-generator
+        run: |
+          cd $GITHUB_WORKSPACE/boulder/test/load-generator
+          go install
+          cd -
```
aarongable, line 68: You can use [steps.working-directory](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsworking-directory) to set the pwd instead of using cd.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,92 @@
...
+
+      # Perform a test issuance with chisel2.py
+      - name: chisel
+        run: REQUESTS_CA_BUNDLE=./test/certs/pebble.minica.pem python ./test/chisel2.py example.letsencrypt.org elpmaxe.letsencrypt.org
+
+      # Run the load-generator briefly - note, because Pebble isn't using the
+      # load-generator's mock DNS server none of the issuances will succeed. This
+      # step is performed just to shake out data races with concurrent requests.
+      - name: load-generator
+        run: load-generator -config ./test/config/load-generator-config.json > /dev/null
```
aarongable, line 92: I'm fine with this, but either:
a) We should have a bug on file describing this failure mode so we don't forget about it; or
b) We should have a bug on file describing this failure mode and not include the load generator at all until that bug is fixed. This is the approach that https://github.com/letsencrypt/pebble/pull/439 was taking, and I think I slightly prefer it (and it would allow the load generator install steps to be removed).

.golangci.yaml
```diff
@@ -1,3 +1,6 @@
+run:
+  timeout: 9m
```
aarongable, line 2: This seems like a really long timeout for such a small repo. Were you actually seeing it take this long, or is this number just a "it-should-never-take-this-long" backstop?

aarongable (dismissed):
Thank you so much! This looks essentially good to me and I'd honestly be okay landing it as-is because it's just an improvement over the status quo, but I do have a few suggestions for further improvements as well. Thanks!

.golangci.yaml
ldez: I was just related to this PR https://github.com/letsencrypt/pebble/pull/379

.github/workflows/tests.yml
ldez: I just copy pasted the comment from the Travis file.

ldez:
I didn't dismiss your review :thinking:
Is it related to the review commits?

mcpherrinm:
Pushing a commit dismisses the review (this is configurable in GH)

ldez:
I think I fixed everything.
Appveyor is still very slow :smile:

.github/workflows/tests.yml
ldez: I will remove the load generator.

mcpherrinm (approved):

mcpherrinm:
Thanks!  This is a pretty significant improvement.  We can definitely merge this and continue to improve from here.

aarongable (approved):

Issue #440, opened 2024-02-24
Regression time limit exceeded / TimeoutError


ldez:
Hello,

since the merge of https://github.com/letsencrypt/pebble/pull/420, there a problem of timeout.

I observed that on lego but you have the same problem on your CI:

https://ci.appveyor.com/project/jsha/pebble/builds/49233994?fullLog=true

```
DEBUG:acme.client:Storing nonce: uheLwx93nDxWtXFPxcGvlA
Traceback (most recent call last):
  File ".\test\chisel2.py", line 209, in <module>
    auth_and_issue(domains)
  File ".\test\chisel2.py", line 122, in auth_and_issue
    order = client.poll_and_finalize(order)
  File "c:\projects\pebble\certbot\acme\acme\client.py", line 185, in poll_and_finalize
    return self.finalize_order(orderr, deadline)
  File "c:\projects\pebble\certbot\acme\acme\client.py", line 275, in finalize_order
    return self.poll_finalization(orderr, deadline, fetch_alternative_chains)
  File "c:\projects\pebble\certbot\acme\acme\client.py", line 259, in poll_finalization
    raise errors.TimeoutError()
acme.errors.TimeoutError
```

ldez:
For now, I haven't diagnosed the exact reason for this regression (except it's related to the PR).

Maybe @mcpherrinm, you will see the problem quicker than me.

mcpherrinm:
Interesting! I’ll take a look

ldez:
The order status seems to stay on `processing` :thinking:

ldez:
I found the problem:
```go
if len(domains) == 0 || len(ips) == 0 {
```

should be 
```
if len(domains) == 0 && len(ips) == 0 {
```

ldez:
I created a PR #441

PR #444, opened 2024-02-26
CI: Travis -> GitHub Actions; Create Release Binaries and Container Images

sheurich:
[Checks](https://github.com/fastly/pebble/actions/workflows/checks.yml)
[Tests](https://github.com/fastly/pebble/actions/workflows/tests.yml)

As a follow-on to https://github.com/letsencrypt/pebble/pull/439 this change request implements the first step of building a new CI system for Pebble. As this is a testbed environment for Boulder, some of these changes may prove to be useful in the latter context.

Per https://github.com/letsencrypt/pebble/issues/434#issuecomment-1958432503, GitHub Actions is a reasonable choice for a new CI implementation to supersede Travis. Much, but not all, of the existing functionality is present in this initial set of changes.

Items currently implemented include:
- Binary production of `pebble` and `challtestsrv` for platforms:
  - AMD64
    - darwin
    - linux
    - windows
  - ARM64
    - darwin
    - linux
    - windows
- Multiplatform docker images published on the GitHub Container Registry for platforms:
  - linux/amd64
  - linux/arm64
  - windows/amd64
- Golang linting.
- Golang coverage.
- Local test scripts for Go and Docker:
  - `./build.sh`
  - `./docker.sh`
  - `./test.sh`

This PR also adds a `-version` flag to Pebble, which is set during release build to the Git commit ref.

This change builds on earlier work done in https://github.com/letsencrypt/pebble/pull/379 and https://github.com/letsencrypt/pebble/pull/442.


sheurich:
Example workflow run can be seen in:
[CI](https://github.com/fastly/pebble/actions/workflows/ci.yml)

ldez:
I think we can merge our 2 PRs #442

sheurich:
> I think we can merge our 2 PRs #442

@ldez Yes let's do that. I can take a look tomorrow.

ldez:
I mean I think my PR can be merged and your PR will just need to be "cleaned" to remove the duplication with mine.

sheurich:
@ldez glad you found https://github.com/letsencrypt/pebble/issues/440! I spent some time troubleshooting at the CI level but hadn't checked Pebble itself yet.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,116 @@
...
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Run containerized test suite
+        run: ./t.sh
+      - name: Send code coverage
+        uses: shogo82148/actions-goveralls@v1
+        with:
+          path-to-profile: profile.cov
```
ldez, line 116: I think you can drop this part because it's already handled by the other workflow.

.github/workflows/build.yml
```diff
@@ -0,0 +1,93 @@
+name: Build
+
+# Run on push to main and any pull request.
+on:
+  push:
+    branches:
+      - main
+  pull_request:
```
ldez, line 8:
```suggestion
  push:
    tags:
      - v*
```

this workflow uses a secret and publish some elements. 

and I recommend to restore my previous file `go-cross.yml`

.github/workflows/checks.yml
```diff
@@ -1,53 +1,36 @@
...
-          curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin ${GOLANGCI_LINT_VERSION}
-          golangci-lint run
-
-
+          go mod vendor
+          git diff --exit-code vendor || (echo "::error::vendor directory needs to be updated" && exit 1)
+      - name: Run GolangCI-Lint
+        uses: golangci/golangci-lint-action@v4
+        with:
+          version: latest
```
ldez, line 36: I'm the main maintainer of golangci-lint, and based on this repo activity I recommend not using the latest.
The action can have some issues with the cache.

.github/workflows/checks.yml
```diff
@@ -1,53 +1,36 @@
...
 
 jobs:
-
   checks:
-    name: Check Process
     runs-on: ubuntu-latest
-    env:
-      GO_VERSION: oldstable
-      GOLANGCI_LINT_VERSION: v1.56.2
-      CGO_ENABLED: 0
```
ldez, line 21: It is better for the maintenance to have all the configurable elements in one place.

.github/workflows/build.yml
sheurich: Thanks for the quick feedback! What should be different in the PR build process versus release? Is the OS matrix in go-cross intended to verify build success for common platforms, whereas the release output can come from a single build platform?

.github/workflows/checks.yml
sheurich: Why use oldstable for these checks rather than stable?

.github/workflows/build.yml
ldez: The release will publish artifacts (archive, GitHub release, Docker images, etc.), a PR will not do that.
A PR doesn't need to build Docker images or create archives.
It's better to not create one workflow with the conditions but to have a dedicated workflow by topic.

In general, nobody should push on a repo without creating a PR before, so the workflow of the PR will check tests, linting, and binary builds.
So when you merge a PR, tests, and linting are useless because they have been already checked.

This is the same thing for builds: the build for a release is not just binary but a set of builds specific to a release (archive, GitHub release, Docker images, etc.).

.github/workflows/checks.yml
ldez: Because the previous CI was doing that.
You can use `stable`, it's not a problem for me, but I still recommend centralizing the variables, it helps for the maintenance and to diagnose CI problems.

.github/workflows/checks.yml
ldez: I know but I also know the problem behind the action.
This is why the previous workflow doesn't it.
If you to use it, you should disable the cache:
```yml
    skip-cache: true
    skip-build-cache: true
```

sheurich:
@ldez I think all of the changes have been made and would appreciate another review. The PR description hasn't yet been updated to match the actual changes.

.github/workflows/build.yml
ldez, line 1: maybe rename to `release.yml`

sheurich:
This will be ready for review soon, pending a bug fix.

sheurich:
I think this should be ready for review @mcpherrinm

mcpherrinm:
It doesn’t have to be in this PR, but we are generally standardizing on either distroless or debian-slim as the base for docker images instead of alpine.  We can make that change later, or it would be helpful if you want to make it here while we’re changing the docker files.

I’ll review more fully later today

sheurich:
> It doesn’t have to be in this PR, but we are generally standardizing on either distroless or debian-slim as the base for docker images instead of alpine.

I had assumed that Alpine was chosen over Debian as a derivative of the strategy outlined in the Pebble README:
> In places where the ACME specification allows customization/CA choice Pebble aims to make choices different from Boulder.

It's no problem to change to Debian for dev and Distroless for release.

sheurich:
Since `cgo` is not needed by pebble or pebble-challtestsrv, the release image can be `FROM scratch`. This WFM but are there other requirements which need glibc or other items from distroless?

mcpherrinm:
Scratch is fine by me

sheurich:
@mcpherrinm the `scratch` change is implemented and a few other bugs have been squashed.

.github/workflows/release.yml
```diff
@@ -0,0 +1,158 @@
+name: Build Release
+# Run on semver tags.
+on:
+  push:
+    tags:
+      - "v[0-9]+.[0-9]+.[0-9]+"
+permissions:
+  contents: read
+  packages: write # Allow docker/build-push-action to publish to GitHub Container Registry
+  pull-requests: read
```
aarongable, line 10: I don't think this one needs access to PRs, since it's only running on tag pushes

aarongable (commented):
The four shell scripts (docker, lint, mods, and test) seem like they'll be kinda difficult to keep in sync with the commands in the corresponding github actions. Can we either have the github actions call these scripts directly, or leave them out, or clearly document in both places how best to keep them in sync?

sheurich:
> The four shell scripts (docker, lint, mods, and test) seem like they'll be kinda difficult to keep in sync with the commands in the corresponding github actions.

I agree.

> Can we either have the github actions call these scripts directly, or leave them out, or clearly document in both places how best to keep them in sync?

I found them useful and like the idea of being able to run the same checks/tests as CI locally. I will update the workflows to call the scripts.

aarongable (approved):

mcpherrinm (approved):

mcpherrinm:
Merged and tagged as v2.5.0

PR #459, opened 2024-05-01
Require HS256, HS384, or HS512 for EAB

### LINKED ISSUE #455: EAB with pebble 2.5.x
Pebble before 2.5 used to support `HS256` with EAB. Now, only `RS256` and EC seem supported. Is there any description how to use an RSA key for EAB with pebble?

I'd prefer continue using HS256 in my test suite for Apache since that seems to be in use at several CAs still. I appreciated that you want to be on the bleeding edge here, but for testing purposes this seem to be a step back.

mcpherrinm:
During the go-jose/v4 upgrade, I accidentally required the same signature sets for EAB as for the account keys, which is incorrect. This allows the correct MAC-based algorithms. It drops the custom algorithm checks, which are now unreachable as go-jose will enforce the algorithms.

Fixes #455

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -59,6 +59,36 @@ jobs:
         run: |
           git clone https://github.com/eggsampler/acme.git /tmp/eggsampler-acme
           cd /tmp/eggsampler-acme && make test
+  lego-eab-linux:
+    name: Test lego with EAB
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v5
+        with:
+          go-version-file: go.mod
+      - name: Install lego cli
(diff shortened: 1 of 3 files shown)


mcpherrinm:
Running this same integration tests against the current main branch has lego return the expected error from #455:

```
acme: error: 400 :: POST :: https://localhost:14000/sign-me-up :: urn:ietf:params:acme:error:malformed :: failed to decode external account binding: go-jose/go-jose: unexpected signature algorithm "HS256"; expected ["RS256" "ES256" "ES384" "ES512"]
```

mcpherrinm:
I've cherry-picked the command and switched to the key from #428 to make the EAB integration test hit the "base64url" edge-case

aarongable (approved):

Issue #452, opened 2024-03-19
Cannot set DNS server in Docker image


tindzk:
After upgrading to `ghcr.io/letsencrypt/pebble:2.5.1`, I am seeing the following error in our test suite:

```
INFO event=SendRequest url=https://localhost:14000/authZ/WKrIeCnHQULq6mp3BD0eEYTlwq7M_MInYnISVm5RmXA
INFO event=ReceivedResponse location=None body="{\n   \"status\": \"invalid\",\n   \"identifier\": {\n      \"type\": \"dns\",\n      \"value\": \"<redacted>\"\n   },\n   \"challenges\": [\n      {\n         \"type\": \"tls-alpn-01\",\n         \"url\": \"https://localhost:14000/chalZ/qCiMNJz2zm1mLtYTUn9hXC4cizqg25g04UI1B2AWrjk\",\n         \"token\": \"TlK9my79nS0tSjtdmugDJHWUzE6N1MdK_nMjKQrd0zU\",\n         \"status\": \"invalid\",\n         \"validated\": \"2024-03-19T16:51:56Z\",\n         \"error\": {\n            \"type\": \"urn:ietf:params:acme:error:malformed\",\n            \"detail\": \"Error occurred while resolving URL \\\"<redacted>\\\": \\\"lookup <redacted> on 8.8.8.8:53: no such host\\\"\",\n            \"status\": 400\n         }\n      }\n   ],\n   \"expires\": \"2024-03-19T17:51:55Z\"\n}"
```

This suggests that the `-dnsserver` setting is not honoured. As a workaround, I was able to specify a custom DNS server for the entire Docker container, but this required changing the port to 53.

After further investigation, the issue seems to be that container parameters are being ignored altogether. The following invocation should fail, but the invalid parameter goes unnoticed:

```shell
docker run ghcr.io/letsencrypt/pebble:2.5.1 abc
```

The container listing shows that the command is translated into `/app abc`:

```
CONTAINER ID   IMAGE                              COMMAND                  CREATED         STATUS         PORTS                                         NAMES
e1b1b497e357   ghcr.io/letsencrypt/pebble:2.5.1   "/app abc"               2 minutes ago   Up 2 minutes                                                 recursing_kare
```

Another issue I found with the new Docker image is that it does not expose any ports, requiring users to override the `ExposedPorts` setting when using the Docker daemon directly. Previously, the `Dockerfile` contained these two lines:

```docker
EXPOSE 14000
EXPOSE 15000
[...]

PR #465, opened 2024-05-24
Document exposing API and management ports when not using docker-compose.yaml

pgporada:
https://github.com/letsencrypt/pebble/pull/444 removed exposing the API and management ports directly in the `Dockerfile` in favor of the `docker-compose.yml`. Not everyone will be using `docker compose` so let's document how it should be done.

Fixes https://github.com/letsencrypt/pebble/issues/452

### DIFF
--- a/Dockerfile.release
+++ b/Dockerfile.release
@@ -27,5 +27,8 @@ ENTRYPOINT [ "/app.exe" ]
 
 # Final stage: dynamically select between Linux and Windows stages based on TARGETOS argument
 FROM ${TARGETOS} AS final
+# Ports needed for testing when not using docker compose.
+EXPOSE 14000
+EXPOSE 15000
 # Add test configuration files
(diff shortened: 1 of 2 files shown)


mcpherrinm:
It’s not quite so simple, because we use the same docker file for both pebble and pebble-challtestsrv

pgporada:
We should split that apart then so they behave appropriately running as one-off `docker run`'s or `docker compose up`.

mcpherrinm:
`EXPOSE` is only really for the use `docker run --publish-all` instead of specifying the ports with `-p`, which is at most a minor inconvenience IMO, so we should consider how much extra complexity we want to add here to support that

pgporada:
How about switching all the documented `docker run` steps to add in those ports?

aarongable (approved):

PR #466, opened 2024-05-24
Merge challtestsrv package into Pebble

mcpherrinm:
This moves the challtestsrv package into the Pebble source tree.

All users of challtestsrv that I'm aware of also use Pebble, and having them in two repos means they're harder to maintain.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -12,7 +12,7 @@ permissions:
 env:
   CGO_ENABLED: 0
   GO_VERSION: stable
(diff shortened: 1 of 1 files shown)


aarongable:
When this is fully in shape, let's do a little bit of grooming to preserve history:
- Squash all commits _after_ the merge commit into a single commit, force-push to this PR
- Turn off "require linear history" in the repo settings
- Merge as-is (neither squashing nor merging) so that we get two commits on pebble/main: the merge commit itself, and the fixups commit

PR #468, opened 2024-06-14
Also support signing CSRs with Ed25519 keys

inteon:
**Background:**
We use pebble to test [cert-manager](https://github.com/cert-manager/cert-manager). We are still using an older pebble version that did not verify the CSR signature algorithm (before https://github.com/letsencrypt/pebble/pull/386). One of our tests makes sure that we can use ACME to obtain a Certificate for a Ed25519 CSR, we were surprised that this signature algorithm is no longer allowed by pebble.

**Proposed change:**
Add support for signing CSRs signed with Ed25519 keys.

### DIFF
--- a/wfe/wfe.go
+++ b/wfe/wfe.go
@@ -127,6 +127,7 @@ var goodSignatureAlgorithms = map[x509.SignatureAlgorithm]bool{
 	x509.ECDSAWithSHA256: true,
 	x509.ECDSAWithSHA384: true,
 	x509.ECDSAWithSHA512: true,
+	x509.PureEd25519:     true,
 }
 
 // newAccountRequest is the ACME account information submitted by the client


aarongable:
Pebble is intended as a test bed for ACME clients that will eventually integrate with Boulder or other WebPKI ACME servers. For better or for worse, Ed25519 is _not_ an [accepted public key algorithm](https://github.com/cabforum/servercert/blob/main/docs/BR.md#615-key-sizes) in the Public WebPKI. Adding support for Ed25519 here runs the risk of causing client authors or site operators to believe that Ed25519 issuance is also possible in production, so it is unlikely that we will accept this PR.

inteon:
> Pebble is intended as a test bed for ACME clients that will eventually integrate with Boulder or other WebPKI ACME servers. For better or for worse, Ed25519 is _not_ an [accepted public key algorithm](https://github.com/cabforum/servercert/blob/main/docs/BR.md#615-key-sizes) in the Public WebPKI. Adding support for Ed25519 here runs the risk of causing client authors or site operators to believe that Ed25519 issuance is also possible in production, so it is unlikely that we will accept this PR.

Thank you for the feedback!

PR #472, opened 2024-08-20
Simplify KU, EKU, and SKID fields of issued certs

aarongable:
To reflect current WebPKI best practices:
- Do not include the KeyEncipherment KU in end-entity certs (it is not used in TLS 1.2+)
- Do not include the TLS Client Auth EKU in any certs (root programs are moving towards single-purpose hierarchies)
- Do not include the Subject Key ID in end-entity certs (it is not useful for chain-building there)

### DIFF
--- a/ca/ca.go
+++ b/ca/ca.go
@@ -257,7 +257,6 @@ func (ca *CAImpl) newCertificate(domains []string, ips []net.IP, key crypto.Publ
 	if len(domains) == 0 && len(ips) == 0 {
 		return nil, errors.New("must specify at least one domain name or IP address")
 	}
-	var err error
 
 	defaultChain := ca.chains[0].intermediates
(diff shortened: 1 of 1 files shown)


ca/ca.go
```diff
@@ -257,18 +257,14 @@ func (ca *CAImpl) newCertificate(domains []string, ips []net.IP, key crypto.Publ
 	if len(domains) == 0 && len(ips) == 0 {
 		return nil, errors.New("must specify at least one domain name or IP address")
 	}
+	var err error
```
mcpherrinm, line 260: Can we move this down a bit? it's not needed until the `if notBefore` block

mcpherrinm: (and maybe even only inside that if statement)

mcpherrinm (approved):

PR #484, opened 2025-02-15
return logical and compliant ARI windows for expiring certs

jvanasco:
While testing ARI functionality with extremely short lived certs, Pebble returned illogical data - such as the windowEnd appearing in the future for already expired certs, or after the cert's notAfter.

This PR adjusts `RenewalInfoSimple` to do the following:

* if the cert has expired, just return `RenewalInfoImmediate`
* if the cert will expire before the `windowEnd`, set the `windowEnd` to the cert's `notAfter`
* cleanup: RFC requires the `windowStart` to be before `windowEnd`, so remove a second from the `windowStart` if needed.

I also made the `RenewalInfoImmediate` window change from `-60:-30` minutes to `-60:-59.59` minutes; the sole reason for this change is that it is visually easier to see the change (and not need to do mental math) and recognize this is a `RenewalInfoImmediate` from logs. 

There are likely better ways to accomplish this, and I may be overlooking some elements from the RFC.

With these changes, Pebble can be configured for certificate validity periods as low as 5-10seconds, enabling quick testing of real (unmocked) flows such as:

* issue cert
* query ari
* wait for expiry or query ari
* renew with `replaces`

Also: The linting errors existed on creating my fork; there may be some linting errors in this PR - but there is a library issue.

### DIFF
--- a/core/types.go
+++ b/core/types.go
@@ -276,6 +276,10 @@ func RenewalInfoSimple(issued time.Time, expires time.Time, now time.Time) *Rene
 	if windowEnd.After(expires){
 		windowEnd = expires
 	}
+	// Ensure correct start for future issueds
+	if windowStart.After(issued){
+		windowStart = issued
+	}
 	
 	// draft-ietf-acme-ari states:
 	// A RenewalInfo object in which the end timestamp
--- a/core/types.go
+++ b/core/types.go
@@ -260,29 +260,29 @@ type RenewalInfo struct {
 // using a very simple renewal calculation: calculate a point 2/3rds of the way
 // through the validity period, then give a 2-day window around that. Both the
 // `issued` and `expires` timestamps are expected to be UTC.
-func RenewalInfoSimple(issued time.Time, expires time.Time) *RenewalInfo {
+func RenewalInfoSimple(issued time.Time, expires time.Time, now time.Time) *RenewalInfo {
 	validity := expires.Add(time.Second).Sub(issued)
 	renewalOffset := validity / time.Duration(3)
 	idealRenewal := expires.Add(-renewalOffset)
 	windowStart := idealRenewal.Add(-24 * time.Hour)
 	windowEnd := idealRenewal.Add(24 * time.Hour)
-	now := time.Now().In(time.UTC)
 
-    if expires.Before(now) {
-        // If the cert has expired, return `RenewalInfoImmediate` instead.
-        return RenewalInfoImmediate(now)
-    }
-
-    if expires.Before(windowEnd) {
-        // If the windowEnd is after the cert's expiry, adjust it to the window end.
-        windowStart = expires.Add(-1 * time.Second)
-        windowEnd = expires
-    }
-    
-    if !windowEnd.After(windowStart) {
-        // Ensure windowStart is before windowEnd
-        windowEnd = expires.Add(-1 * time.Second)
-    }
+	if expires.Before(now) {
+		// If the cert has expired, return `RenewalInfoImmediate` instead.
+		return RenewalInfoImmediate(now)
+	}
+	
+	// Ensure RenewalWindow is not after the expiry
+	if windowEnd.After(expires){
+		windowEnd = expires
+	}
+	
+	// draft-ietf-acme-ari states:
+	// A RenewalInfo object in which the end timestamp
+	// equals or precedes the start timestamp is invalid.
+	if !windowStart.Before(windowEnd){
+		windowStart = windowStart.Add(-1 * time.Second)
+	}
 
 	return &RenewalInfo{
 		SuggestedWindow: SuggestedWindow{
--- a/core/types.go
+++ b/core/types.go
@@ -279,7 +279,7 @@ func RenewalInfoSimple(issued time.Time, expires time.Time) *RenewalInfo {
         windowEnd = expires
     }
     
-    if !windowEnd.after(windowStart) {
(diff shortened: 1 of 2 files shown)


core/types.go
```diff
@@ -264,10 +264,30 @@ func RenewalInfoSimple(issued time.Time, expires time.Time) *RenewalInfo {
 	validity := expires.Add(time.Second).Sub(issued)
 	renewalOffset := validity / time.Duration(3)
 	idealRenewal := expires.Add(-renewalOffset)
+	windowStart := idealRenewal.Add(-24 * time.Hour)
+	windowEnd := idealRenewal.Add(24 * time.Hour)
+	now := time.Now().In(time.UTC)
```
aarongable, line 269: Having this function compute `now`, while the other similar helpers take `now` as an argument, doesn't feel right. They should all behave the same way -- either this one should take `now` as an argument, or the others should have their single argument removed.

core/types.go
```diff
@@ -264,10 +264,30 @@ func RenewalInfoSimple(issued time.Time, expires time.Time) *RenewalInfo {
...
+    if expires.Before(windowEnd) {
+        // If the windowEnd is after the cert's expiry, adjust it to the window end.
+        windowStart = expires.Add(-1 * time.Second)
+        windowEnd = expires
+    }
+    
+    if !windowEnd.After(windowStart) {
+        // Ensure windowStart is before windowEnd
+        windowEnd = expires.Add(-1 * time.Second)
+    }
```
aarongable, line 285: There is a lint error here: these lines all have incorrect indentation.

More importantly, I think there's a much simpler version of this logic that accomplishes the same goal:
- If windowEnd is after expires, make it equal to expires
- If windowStart is before issued, make it equal to issued

True, this doesn't explicitly protect against a certificate which has a zero-second validity period, but I'm not worried about that case.

jvanasco: I did the first element  and changed initial spacing.  I don't know why my editor was auto-expanding tabs here.  I added the windowStart change to issued as well.

I left the last element, however, ensuring `windowStart` is BEFORE `windowEnd` - and included a reference to the RFC.  The RFC draft explicitly requires `windowStart` to precede `windowEnd`, and the payload is invalid otherwise.

Striving to simplify logic has caveats.  I prioritized ensuring an ARI compliant payload over edge cases of future notBefore values or zero second validity.

jvanasco (commented):
Some requested changes were made; some alternatives were presented.

core/types.go
```diff
@@ -260,14 +260,38 @@ type RenewalInfo struct {
...
+	windowStart := idealRenewal.Add(-24 * time.Hour)
+	windowEnd := idealRenewal.Add(24 * time.Hour)
+
+	if expires.Before(now) {
+		// If the cert has expired, return `RenewalInfoImmediate` instead.
+		return RenewalInfoImmediate(now)
+	}
+	
+	// Ensure RenewalWindow is not after the expiry
+	if windowEnd.After(expires){
```
aarongable, line 276: You still have formatting issues here, and below. You should be running `gofmt` automatically.
```suggestion
	if windowEnd.After(expires) {
```

core/types.go
```diff
@@ -260,14 +260,38 @@ type RenewalInfo struct {
...
 	validity := expires.Add(time.Second).Sub(issued)
 	renewalOffset := validity / time.Duration(3)
 	idealRenewal := expires.Add(-renewalOffset)
+	windowStart := idealRenewal.Add(-24 * time.Hour)
+	windowEnd := idealRenewal.Add(24 * time.Hour)
+
+	if expires.Before(now) {
+		// If the cert has expired, return `RenewalInfoImmediate` instead.
+		return RenewalInfoImmediate(now)
+	}
```
aarongable, line 273: This is redundant: if we're clamping `windowEnd` to `expires`, then an expired cert will always be given a window that is wholly in the past. No need to special case it here. This is why I suggested the removal of this stanza in my previous review.

If you want to keep it anyway, move it above the computation of the ideal window, since none of that computation is necessary for this stanza.

core/types.go
```diff
@@ -260,14 +260,38 @@ type RenewalInfo struct {
...
+	
+	// Ensure RenewalWindow is not after the expiry
+	if windowEnd.After(expires){
+		windowEnd = expires
+	}
+	// Ensure correct start for future issueds
+	if windowStart.After(issued){
```
aarongable, line 280: I didn't suggest this to handle issued timestamps that are in the future. I suggested this to handle the case where the certificate has a validity period of less than 48 hours, so that subtracting 24 hours from a point in the middle of the validity period results in a windowStart that precedes the issued timestamp.

Also, you've reversed the logic from what I suggested. This should be `if windowStart.Before(issued) {`.

aarongable (changes requested):

core/types.go
jvanasco: Ah! It's not performing on directories, just specific files.  Fixed.

core/types.go
jvanasco: Oh, that makes sense.  I just removed it.

core/types.go
jvanasco: Sorry, I was trying to do this in one window while debugging another upstream project in the cloud.

jvanasco:
This should have all the requested changes and formatting.  Apologies.  I've also tested it against with a client as I couldn't find any ARI tests here.

aarongable (approved):

PR #488, opened 2025-02-25
Update golangci-lint

mcpherrinm:
golangci-lint is failing on main due to a config problem.  Update configs, version of golangci-lint, and remove a variable that's unused and causing a lint failure

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -12,7 +12,7 @@ permissions:
 env:
   CGO_ENABLED: 0
   GO_VERSION: stable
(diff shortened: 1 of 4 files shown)


mcpherrinm:
```
Error: core/types.go:263:61: unused-parameter: parameter 'now' seems to be unused, consider removing or renaming it as _ (revive)
 func RenewalInfoSimple(issued time.Time, expires time.Time, now time.Time) *RenewalInfo {
```

Seem like a legitimate problem

aarongable (approved):

PR #489, opened 2025-02-28
fix(dns-account-01): update validation domain to remove scope

sheurich:
sync with https://datatracker.ietf.org/doc/html/draft-ietf-acme-dns-account-label-00


sheurich:
👋🏽 Hi @mcpherrinm would you be able to take a look at this change? It is fairly minor but needed to reflect the current spec.

mcpherrinm:
Sorry, I totally missed this PR. I just merged a fix from another PR, so the issue is solved.

PR #492, opened 2025-04-01
currectly triggers BadSignatureAlgorithmProblem at JWS parse time

orangepizza:
Pebble didn't have badSignatureAlgorithm error, while RFC8555 section 6.2 mandates it. 

jws parse time check for bad algorithm is needed because same as jose.parsesigned refuese to parse without algolist(which is good thing) so error already thrown at that time and we need to catch this type of error there.

see related boulder PR
https://github.com/letsencrypt/boulder/pull/8091


wfe/wfe.go
```diff
@@ -858,6 +858,9 @@ func (wfe *WebFrontEndImpl) verifyPOST(
 	body := string(bodyBytes)
 	parsedJWS, err := wfe.parseJWS(body)
 	if err != nil {
+		if strings.Contains(err.Error(), "unexpected signature algorithm") {
```
aarongable, line 861: Rather than checking the contents of the error string, I think it would be better for `wfe.parseJWS` to behave the same as `wfe.verifyJWS` and simply return a ProblemDetails directly instead of returning an error. Then parseJWS can decide which acme error is most appropriate in each case.

(Note, this is the opposite of how Boulder does things. But pebble doesn't have boulder's automatic err-to-problem conversion.)

orangepizza: is it safe to update requried go version to 1.24 to use new go-jose? othersiwe string match on error is required as that is only way to catch signiture algorithm error from current version. this error is sent from go-jose, not parseJWS

orangepizza: updated go version to used with newer go-jose: if one revert to 6b7368b this can be cancel

wfe/wfe.go
```diff
@@ -675,34 +675,41 @@ func (wfe *WebFrontEndImpl) parseJWS(body string) (*jose.JSONWebSignature, error
...
 
 	parsedJWS, err := jose.ParseSigned(body, goodJWSSignatureAlgorithms)
 	if err != nil {
-		return nil, fmt.Errorf("Parse error reading JWS: %w", err)
+		var unexpectedSignAlgo *jose.ErrUnexpectedSignatureAlgorithm
+		if errors.As(err, &unexpectedSignAlgo) {
+			errmsg := fmt.Sprintf(
+				"JWS header contained unsupported algorithm, supported algorithms are: %s", goodJWSSignatureAlgorithms,
+			)
+			return nil, acme.BadSignatureAlgorithmProblem(errmsg)
```
aarongable, line 702:
```suggestion
			return nil, acme.BadSignatureAlgorithmProblem(fmt.Sprintf(
				"JWS header contained unsupported algorithm, supported algorithms are: %s", goodJWSSignatureAlgorithms))
```

aarongable (approved):

aarongable (approved):
LGTM; waiting on second review+approval.

mcpherrinm (approved):

PR #498, opened 2025-05-07
use newer validation subdomain for dns-account-01 (fix CI eggsampler/acme error)

orangepizza:
at 2024 November draft writers of draft-ietf-acme-dns-account-label decide to drop concept in dns-account-01 challenge. This changed validation subdomain schema and breaks test about it as they are inserting record in newer subdomain and pebble was checking old subdomain.

this patch makes pebble use new subdoman in newer draft, finally solve CI error from eggsampler/acme lingered in every PR here for few months.


mcpherrinm (approved):
Thanks!

PR #501, opened 2025-06-04
add overriding of ARI response

### LINKED ISSUE #486: Add API to adjust ARI window of certificates
in current implementation renew by ari info won't triggered, because due to normally ARI window is 2/3 of certs lifetime (5 years by default, so 3 years after sign) and only why do change is revoke it, but then it will trigger revoked cert renew now path first, ignoring ARI. 
and currently pebble will never populate explanationURL, so new api to adjust ARI to set time with explanation can be helpful.

jsha:
Fixes #486

This moves the GetCertificateBySerial call earlier, which means that call needs to succeed even for revoked certificates. So this also follows up on #252 by keeping revoked certs in the primary certificatesByID map (while still adding them to the revokedCertificatesByID map).

### DIFF
--- a/wfe/wfe.go
+++ b/wfe/wfe.go
@@ -57,7 +57,7 @@ const (
 	// Draft or likely-to-change paths
 	renewalInfoPath = "/draft-ietf-acme-ari-03/renewalInfo/"
 
-	// Theses entrypoints are not a part of the standard ACME endpoints,
+	// These entrypoints are not a part of the standard ACME endpoints,
 	// and are exposed by Pebble as an integration test tool. We export
 	// RootCertPath so that the pebble binary can reference it.
 	RootCertPath         = "/roots/"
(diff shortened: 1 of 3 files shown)


wfe/wfe.go
```diff
@@ -1917,7 +1935,47 @@ func (wfe *WebFrontEndImpl) RenewalInfo(_ context.Context, response http.Respons
...
-func (wfe *WebFrontEndImpl) determineARIWindow(id *core.CertID) (*core.RenewalInfo, error) {
+// SetRenewalInfo overrides the default ARI response for a certificate.
+func (wfe *WebFrontEndImpl) SetRenewalInfo(_ context.Context, response http.ResponseWriter, request *http.Request) {
+	body, err := io.ReadAll(request.Body)
+	if err != nil {
+		wfe.sendError(acme.InternalErrorProblem("Error reading body"), response)
+	}
+
+	var reqJSON struct {
+		Certificate string // in PEM form
```
aarongable, line 1946: Would it make sense for this to be the cert ID (like you would use in the path of the ARI request) rather than the whole certificate?

aarongable (approved):

PR #503, opened 2025-06-06
Set Content-Type for custom ARI response

### LINKED ISSUE #502: text/plain content-type on custom ARI response
If you set a custom ARI response as per #501, the http response you get has a content-type of text/plain instead of application/json. Correct me if I'm wrong but I'm pretty sure that is not according to expectations from https://datatracker.ietf.org/doc/draft-ietf-acme-ari/.

For completeness, these are the HTTP response headers from a request where the custom ARI response was not set:
```
content-type: application/json; charset=utf-8
link: <https://localhost:14000/dir>;rel="index"
retry-after: 21600
content-length: 107
date: Fri, 06 Jun 2025 11:41:43 GMT
```
And these are the HTTP response headers from a request where it was set:
```
link: <https://localhost:14000/dir>;rel="index"
content-type: text/plain; charset=utf-8
content-length: 110
date: Fri, 06 Jun 2025 11:42:44 GMT
```

I noticed the `retry-after` is also missing but I don't know if that is a problem or not.

jsha:
Fixes #502

### DIFF
--- a/wfe/wfe.go
+++ b/wfe/wfe.go
@@ -3012,13 +3012,13 @@ func (wfe *WebFrontEndImpl) processRevocation(
 		return acme.MalformedProblem("Error decoding Base64url-encoded DER: " + err.Error())
 	}
 
+	revokedCert := wfe.db.GetRevokedCertificateByDER(derBytes)
+	if revokedCert != nil {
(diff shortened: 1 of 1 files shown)


aarongable (commented):
We should probably set the Retry-After header, too

Issue #508, opened 2025-07-28
Consider (randomly) dropping Location header from order finalization response


benburkert:
On the ACME mailing list I asked about the `Location` header for finalize order responses being required or optional, and @aarongable weighed in that it's not a requirement but it is nice to provide. I agree and think it's great for boulder to provide the header, but I also think that it would be better for pebble to drop the header aka revert #85.

Pebble is pretty commonly used as the reference server for new ACME clients, and it seems like including this header has created a de facto assumption that all servers will include the header. It's great for clients since they can rebuild an order object from scratch using just the response from finalize if the `Location` header is present. But there are ACME servers (like Buypass.com) that do not include this header, so any client that depends on the presence of the header may not be compatible. Here's an example: https://github.com/unixcharles/acme-client/pull/250

Would you consider a PR to remove the `Location` header from finalize order responses, either conditionally (via an env var or cli flag) or unconditionally? Or alternatively, randomly include it only a percentage of the time in a similar way to the `PEBBLE_WFE_NONCEREJECT` env var?

PR #509, opened 2025-07-28
Revert "Add Location header to finalize response (#85)"

aarongable:
This reverts commit 9c3d91bf5b6b798a677d50232e6149ab99620a52, which was merged as a result of https://github.com/letsencrypt/pebble/pull/85. That PR based on the idea that the Location header is required to be included with the Finalize response in RFC 8555, but this is not the case: it appears in the example response, but its inclusion there is not mandated by the text.

Although Boulder does provide the Location header in Finalize responses, there exist other ACME implementations which do not. Therefore it is useful for Pebble to omit it, so that clients developed against Pebble do not come to rely on its inclusion.

Fixes https://github.com/letsencrypt/pebble/issues/508


jsha (approved):

aarongable:
Tests are failing because eggsampler/acme assumes the Finalize response has a Location header :D
https://github.com/eggsampler/acme/blob/cee0a72e65be5f65e4aa783e5c96d010c4e51b18/order.go#L162

aarongable:
https://github.com/eggsampler/acme/pull/28 to update eggsampler; if/when that merges I'll update the version of eggsampler used by the pebble tests in this PR.

PR #515, opened 2025-09-14
permit `Transfer-Encoding: chunked` HTTP request bodies

benburkert:
Drop the requirement for POST requests to contain a Content-Length header, which breaks support for chunked transfer encoded requests. See letsencrypt/boulder#8403 for more context.


jsha (approved):
Thanks!

PR #517, opened 2025-10-02
Allow ML-DSA roots and intermediates

aarongable:
Add support for using ML-DSA-65 to generate Pebble's root and intermediate keys and certificates.

This uses Cloudflare's "circl" library to  handle generating ML-DSA keys, serializing pubkeys, and generating signatures. Specifically, it uses the `mldsa44`, `mldsa65`, and `mldsa87` packages, as well as the abstractions provided by the `sign` and `pki` packages:
- https://pkg.go.dev/github.com/cloudflare/circl@v1.6.1/sign/mldsa/mldsa44
- https://pkg.go.dev/github.com/cloudflare/circl@v1.6.1/sign/mldsa/mldsa65
- https://pkg.go.dev/github.com/cloudflare/circl@v1.6.1/sign/mldsa/mldsa87
- https://pkg.go.dev/github.com/cloudflare/circl@v1.6.1/sign
- https://pkg.go.dev/github.com/cloudflare/circl@v1.6.1/pki#CertificateScheme

However, the Go stdlib crypto/x509 package does not support using these keys. Although they implement the crypto.Signer interface, the x509 package also needs to be able to map keys to PublicKeyAlgorithm and SignatureAlgorithm OIDs, and does not provide any hooks for crypto.Signers to provide that information. As such, this PR also forks the relevant portions of crypto/x509 (MarshalPKIXPublicKey, CreateCertificate, and ParseCertificate), strips them down to their bare bones, and then adds support for ML-DSA.


Muzosh:
> We may need to fork crypto/x509

Forking individual package from Go's standard library is almost impossible, I tried. See the reason [here](https://github.com/ISRI-PQC/agiligo/blob/feature-crypto-agility/README_JOURNEY.md#why-is-this-repository-a-complete-fork-of-go-couldnt-this-be-just-a-modified-gocrypto-package).

As I already mentioned in https://github.com/golang/go/issues/64537, check out [AgiliGo](https://github.com/ISRI-PQC/agiligo/tree/feature-crypto-agility).

pgporada:
We should be able to issue `mldsa65` roots and intermediates once go1.27 is released. https://github.com/golang/go/issues/78888

PR #544, opened 2026-04-01
Add accounturi to dns-persist-01 challenges

felixfontein:
This has been added to draft 01: https://www.ietf.org/archive/id/draft-ietf-acme-dns-persist-01.html#section-3.1 vs. https://www.ietf.org/archive/id/draft-ietf-acme-dns-persist-00.html#section-3.1

Diff: https://author-tools.ietf.org/iddiff?url1=draft-ietf-acme-dns-persist-00&url2=draft-ietf-acme-dns-persist-01

### DIFF
--- a/wfe/wfe.go
+++ b/wfe/wfe.go
@@ -1625,6 +1625,7 @@ func (wfe *WebFrontEndImpl) makeChallenge(
 	}
 	if chalType == acme.ChallengeDNSPersist01 {
 		chal.IssuerDomainNames = append([]string(nil), wfe.caaIdentities...)
+		chal.AccountURI = wfe.relativeEndpoint(request, fmt.Sprintf("%s%s", acctPath, authz.Order.AccountID))
 	}
 
 	// Add it to the in-memory database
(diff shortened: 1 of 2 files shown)


beautifulentropy (dismissed):
Thanks for the contribution, this looks good to me.

wfe/wfe.go
```diff
@@ -1625,6 +1625,7 @@ func (wfe *WebFrontEndImpl) makeChallenge(
 	}
 	if chalType == acme.ChallengeDNSPersist01 {
 		chal.IssuerDomainNames = append([]string(nil), wfe.caaIdentities...)
+		chal.AccountURI = wfe.relativeEndpoint(request, fmt.Sprintf("%s%s", acctPath, authz.Order.AccountID))
```
aarongable, line 1631:
```suggestion
		// Note: By using web.relativeEndpoint here, Pebble will reflect the Host header
		// into the accountURI here. This would not be acceptable in a security-conscious
		// context, but is okay for Pebble.
		chal.AccountURI = wfe.relativeEndpoint(request, fmt.Sprintf("%s%s", acctPath, authz.Order.AccountID))
```

aarongable (commented):
Code change LGTM, but we really need to make sure that no one thinks merely reflecting the Host: header into the challenge's canonical accountURI is _actually_ acceptable, so I'd like a comment something like the suggestion below to be included.

aarongable (approved):

beautifulentropy (approved):

felixfontein:
@beautifulentropy @aarongable thanks for reviewing and merging!

PR #547, opened 2026-05-26
chore: update workflows

ldez:
I updated the versions of the actions to the latest possible and applied some security good practices.

I also updated golangci-lint to the latest version and lego to v5.

---

Not in this PR, but I recommend several things to improve security and maintainability:
- use goreleaser to simplify the maintenance of the release process and review the permissions related to this process
- use immutable releases (I think this is not possible with the current release process)
- pin the versions of the actions and use dependabot or renovatebot to keep them up to date
- add [zizmor](https://github.com/zizmorcore/zizmor-action) as a check

The fact to use goreleaser was rejected in 2019, but it's still think that it's the best way to automate the release process.
This is more maintainable than the current process with upload and download of artifacts.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -4,9 +4,9 @@ on:
     branches:
       - main
   pull_request:
-permissions:
-  contents: read
-  pull-requests: read
+
+permissions: {}
+
 jobs:
   certbot-linux:
     name: Run Certbot tests on Linux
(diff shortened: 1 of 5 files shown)


.github/workflows/tests.yml
```diff
@@ -63,12 +67,14 @@ jobs:
...
     steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-go@v5
+      - uses: actions/checkout@v6
         with:
-          go-version-file: go.mod
+          persist-credentials: false
+      - uses: actions/setup-go@v6
+        with:
+          go-version: stable
```
aarongable, line 75: Why change this?
```suggestion
          go-version-file: go.mod
```

ldez: Because lego requires min go1.25 but pebble uses [go1.24](https://github.com/letsencrypt/pebble/blob/b1e1ca4f3c30abb64111adaca4544bc5374cc306/go.mod#L3).

The Go version used to compile lego doesn't have to match the Go version used to compile pebble.

I can use `oldstable` if you want, as lego always uses Go latest-1 (== `oldstable`).

ldez:
Because of PR #548 I was forced to rewrite the PR from scratch, this was not fun.
I don't understand why you didn't ask me to split my PR or to add zizmor.

aarongable (approved):
Apologies for the mess on this PR. The zizmor change that caused you trouble was part of a set of changes we made across _all_ our repos, not specific to pebble at all.

PR #551, opened 2026-08-04
Use RFC 7093 SHA-512 truncated SKIs

pzduniak:
Always generate Subject Key Identifiers according to RFC 7093 Section 2 method 3: the leftmost 160 bits of the SHA-512 hash of the `subjectPublicKey` BIT STRING.

This removes Pebble's use of SHA-1, so it can run with `GOFIPS140=v1.0.0` / `GODEBUG=fips140=only`. It also stays distinct from Boulder, which uses RFC 7093 Section 2 method 1 (truncated SHA-256).

No configuration or library options: Pebble is a test harness and does not need SHA-1 legacy compatibility.

Tests:
- `go test ./ca ./cmd/pebble`
- `GOFIPS140=v1.0.0 go test ./ca`
- `GODEBUG=fips140=only GOFIPS140=v1.0.0 go test ./ca`


aarongable (changes requested):
No need for legacy, optionality, or backwards compatibility. Pebble is not intended as a long-lived system for issuing trusted certificates, it is intended as a test harness.

I think the correct action here should be simply changing Pebble to always produce Subject Key Identifiers which are produced according to RFC 7093, Section 2, Paragraph (3). In other words, they should be the leftmost 160 bits of the SHA512 hash of the subjectPublicKey bit string. This should allow Pebble to do away with all use of SHA1, which still behaving differently from Boulder to ensure flexibility in the ecosystem.

aarongable (approved):

ca/ca_test.go
```diff
@@ -30,6 +31,33 @@ func makeCa() *CAImpl {
...
+func TestMakeSubjectKeyID(t *testing.T) {
+	// RFC 7093 Section 3 example P-256 public key.
+	publicKeyBytes, err := hex.DecodeString(
+		"047f7f35a79794c950060b8029fc8f363a28f11159692d9d34e6ac948190434735" +
+			"f833b1a66652dc514337aff7f5c9c75d670c019d95a5d639b72744c64a9128bb",
+	)
+	if err != nil {
+		t.Fatal(err)
+	}
+	x, y := elliptic.Unmarshal(elliptic.P256(), publicKeyBytes)
```
aarongable, line 43: See lint finding: use crypto/ecdh instead of elliptic.Unmarshal

aarongable (changes requested):

aarongable (approved):

PR #553, opened 2026-09-17
Fix locking in CompleteOrder error path

notroj:
I've been trying to diagnose some flaky CI in `mod_md` over in `apache/httpd` which uses pebble. Along the way Claude identified this problem, which I'm not sure is definitely the root cause but the bug looks genuine.


jfclere (approved):
Look good to me.

aarongable (approved):
LGTM. I wish Go had "defer until this loop iteration is exited", to avoid situations like this.

